PR #19, opened 2018-06-05
pt-br metadata description translated

ncoelho:


Overtorment:
@ncoelho can you please take a look at other files in fastlane/metadata/pt-BR folder?
at least `name` and `keywords`

ncoelho:
Sure. Name should stay the same.

Overtorment:
Its now `BlueWallet - Bitcoin wallet` - maybe should be in PT as well..?

ncoelho:
"bitcoin" and "wallet" are like two expressions that don't get translated and are used from the english, it's common in latin languages to use this "anglo-saxon"/english terms. Same as a "segwit" doesn't get translated...

PR #44, opened 2018-09-14
Updated Expon and depracated calls.

marcosrdz:


loc/index.js
```diff
@@ -15,8 +15,8 @@ let strings;
   }
 
   // TODO: rewrite this when Expo version is upped
```
Overtorment: comment is obsolte

package.json
```diff
@@ -47,7 +47,7 @@
     "crypto-js": "^3.1.9-1",
     "eslint-config-prettier": "^2.9.0",
     "eslint-plugin-prettier": "^2.6.0",
-    "expo": "^23.0.6",
+    "expo": "^30.0.0",
     "expo-analytics": "^1.0.7",
     "frisbee": "^1.6.4",
```
Overtorment: jest-expo should be updated to to "^30.0.0"
(we use it)

Overtorment:
Also, jest tests are broken completely.
Looks like some magic with dependencies/configuration needed.

at least tests HDWallet.test.js and  App.test.js should pass
(LightningCustodianWallet.test.js is broken because backend is broken atm)

Overtorment:
Very well done

PR #69, opened 2018-09-18
Fixed undefined constants and layout issues.

marcosrdz:


Overtorment:
damn, why didnt circleCI launch, I should have fixed the integration...

PR #75, opened 2018-09-26
Upgrade to Expo 30.

marcosrdz:


screen/lnd/scanLndInvoice.js
```diff
@@ -95,7 +96,7 @@ export default class ScanLndInvoice extends React.Component {
       onCameraReady: function() {
         alert('onCameraReady');
```
Overtorment: Never seen this alert. This is  copypasted from example. Probably better remove?

PR #81, opened 2018-09-29
 FIX: Added share button to Receive screen. User is now able to tap-to-copy

marcosrdz:


Overtorment:
@marcosrdz circleci magically started working and caught broken tests

PR #93, opened 2018-10-06
Review translations pt

ncoelho:
Reviewed translations pt BR
Reviewed translations pt PT
Created new language and menus pt PT


loc/index.js
```diff
@@ -19,7 +19,7 @@ let strings;
       locale = locale.split('-');
       locale = locale[0];
       console.log('current locale:', locale);
-      if (locale === 'en' || locale === 'ru' || locale === 'ua' || locale === 'es' || locale === 'pt') {
+      if (locale === 'en' || locale === 'ru' || locale === 'ua' || locale === 'es' || locale === 'pt_br' || locale === 'pt_pt') {
```
Overtorment: you sure this is how locales are named in iOS environment..?

ncoelho: https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html

more like pt-pt instead

PR #96, opened 2018-10-07
Development

Overtorment:


Overtorment:
Issues:

* [ ] When you try to delete wallet it brings you back to the wallet view.
Should bring you to the main view.
* [ ] Empty tx list message missing on new wallet view

screen/settings/encryptStorage.js
```diff
@@ -68,15 +68,14 @@ export default class EncryptStorage extends Component {
                       this.setState({ isLoading: true });
                       let p1 = await prompt(loc.settings.password, loc.settings.password_explain).catch(() => {
                         this.setState({ isLoading: false });
-                        this.props.navigation.goBack();
+                        p1 = undefined;
```
Overtorment: I dont understand this particular change @marcosrdz

marcosrdz: Since Encryption is it's own component, the navigation should not pop back. 
The "p1 = undened" is for the user to cancel out of encrypting the storage.
The prompt, once the storage is encrypted, is not cancellable.

Overtorment: thx

PR #102, opened 2018-10-08
Better call to actions

ncoelho:
Improved call to actions matching the design


Overtorment:
photo5330481527814990073
photo5330481527814990072

send &receive buttons look to big. is that ok?
manage funds button has different rounding on the left and on the right

ncoelho:
Sizes are good. Seems your fonts are weird, too small. 
Did a blind fix to manage founds button, on this branch appears the send instead of manage funds.

Overtorment:
4a993d10-934f-405e-b4bc-452fd1f1b099
db870005-fedb-4cd2-baec-7d7fe47a1401

Incorrect rounding radius for single buttons

PR #105, opened 2018-10-21
ADD: better source of entropy for wallet creation

Overtorment:


Overtorment:
Will have to do for now, until https://expo.canny.io/feature-requests/p/crypto-api is ready

Issue #66, opened 2018-09-17
Improve broadcasting experience


ncoelho:
Tester feedback

> As a somewhat knowledgeable user, I understand the concept of creating a signed transaction and then broadcasting it to the network, however not all users will understand those details.  You can make it less confusing by including more explanation or by using simpler terms such as "Continue" and "Send" or "Confirm".

ncoelho:
This is what we have at the moment.

Modal with signed Hex TX, user confirm or copy for later.
Confirmation screen

https://app.zeplin.io/project/5b252c1057a0c54542fe1f02/screen/5bcb5750393bb37a2c278dde
https://app.zeplin.io/project/5b252c1057a0c54542fe1f02/screen/5bcb574f33a8342f1cff6e21
https://app.zeplin.io/project/5b252c1057a0c54542fe1f02/screen/5bcb574f7543ed09a001bc71

marcosrdz:
Fixed in https://github.com/BlueWallet/BlueWallet/pull/107

PR #107, opened 2018-10-26
FIX: Improve broadcasting experience #66

marcosrdz:


screen/send/success.js
```diff
@@ -0,0 +1,111 @@
...
+            width: 120,
+            height: 120,
+            borderRadius: 60,
+            alignSelf: 'center',
+            justifyContent: 'center',
+            marginTop: 43,
+            marginBottom: 53,
+          }}
+        >
+          <Image style={{ alignSelf: 'center' }} source={require('../../img/baseline-check-24-px.png')} />
```
Overtorment: Id prefer to use font-awesome or some other icon already bundled

screen/send/details.js
```diff
@@ -538,7 +544,8 @@ SendDetails.propTypes = {
       params: PropTypes.shape({
         address: PropTypes.string,
         fromAddress: PropTypes.string,
-        fromSecret: PropTypes.string,
+        satoshiPerByte: PropTypes.string,
+        fromSecret: PropTypes.fromSecret,
```
Overtorment: broadcast() is now duplicated in 2 screens. lets completely remove broadcast functionality from "Transaction details" and leave it only on `send/confirm` screen. will be easier to maintain and look more natural

screen/send/confirm.js
```diff
@@ -0,0 +1,176 @@
...
+    headerTintColor: '#0c2550',
+  };
+
+  constructor(props) {
+    super(props);
+    console.log('send/create constructor');
+
+    this.state = {
+      isLoading: false,
+      amount: props.navigation.state.params.amount,
```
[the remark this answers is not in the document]
Overtorment: nvm if this all was copypasted from send/create

screen/send/confirm.js
```diff
@@ -0,0 +1,176 @@
...
+            style={{
+              color: '#37c0a1',
+              fontSize: 14,
+              marginHorizontal: 4,
+              paddingBottom: 6,
+              fontWeight: '500',
+              alignSelf: 'center',
+            }}
+          >
+            {Number(this.state.satoshiPerByte).toFixed(0)} {BitcoinUnit.SATS}
```
Overtorment: Is that as per  design..?
Better use actual fee, so user knows how much he pays to move funds. satoshiPerByte is a technical metric which shows how profitable it is for miner to include tx in block.
Also, it doesnt day anywhere that this is Fee, its not obvious. Can we write "Fee: xxx"  ?

Overtorment: same for send/success

screen/send/details.js
marcosrdz: Done.

screen/send/confirm.js
```diff
@@ -0,0 +1,176 @@
...
+let loc = require('../../loc');
+let EV = require('../../events');
+
+export default class Confirm extends Component {
+  static navigationOptions = {
+    headerStyle: {
+      backgroundColor: '#FFFFFF',
+      borderBottomWidth: 0,
+    },
+    headerTintColor: '#0c2550',
```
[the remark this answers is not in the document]
marcosrdz: Yup. This is definitely something that needs to be done.

PR #121, opened 2018-11-03
Upgrade to Expo 31 & Android related Fixes

marcosrdz:


entropy.js
```diff
@@ -1,5 +1,5 @@
 import { AppStorage } from './class';
-import Expo from 'expo';
+import { Accelerometer } from 'expo';
```
Overtorment: I have a feeling that there are way more places where we need to change that import line

Issue #27, opened 2018-08-15
Unable to re-name wallets.


Astro92:

Overtorment:
Thanks! Probably name on the card not refreshing till you scroll it left or right a bit.
Will look into.

Overtorment:
Fixed in https://github.com/BlueWallet/BlueWallet/pull/150

PR #150, opened 2018-12-10
Refresh wallet info on page focus

pajasevi:
Wallet name stays the same when changed in options and user has to manually refresh to get new name (as described in #27 ).

Since `react-navigation` keeps staked pages mounted, `componentDidMount` is called only after first render. We have to use `NavigationEvents` component which listens on `onWillFocus` event and calls the `refreshFunction` to get new data.

I'm not sure if this has any large drawbacks, but it is currently the only way to do it properly, the other being using Redux or other state container which would be the single source of truth.

I also removed `async` keywords which do nothing.

### DIFF
--- a/screen/wallets/list.js
+++ b/screen/wallets/list.js
@@ -14,6 +14,7 @@ import {
   BlueHeaderDefaultMain,
 } from '../../BlueComponents';
 import { Icon } from 'react-native-elements';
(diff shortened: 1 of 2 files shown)


Overtorment:
Wow cool. I was wanderring whats the correct way to achieve that.
Thanks!

PR #160, opened 2018-12-18
WIP: Handle BIP21 deeplinks

pajasevi:
This is a work in progress to handle BIP21 (`bitcoin:`) deeplinks.
Current state is: clicked link opens the BlueWallet app, redirects to `SendDetails` page which opens the first wallet and parses amount+label and inserts into send form.

Desired state: Open app, open modal window with available wallets (only those able to send and only those with sufficient balance) and let user choose a wallet.

### DIFF
--- a/App.js
+++ b/App.js
@@ -1,0 +1,45 @@
+import React from 'react';
+import { Linking } from 'react-native';
+import { NavigationActions } from 'react-navigation';
+
+import MainBottomTabs from './MainBottomTabs';
+
+export default class App extends React.Component {
+  navigator = null;
+
+  componentDidMount() {
+    Linking.getInitialURL()
+      .then(url => this.handleOpenURL({ url }))
+      .catch(console.error);
+
+    Linking.addEventListener('url', this.handleOpenURL);
+  }
+
+  componentWillUnmount() {
+    Linking.removeEventListener('url', this.handleOpenURL);
+  }
+
+  handleOpenURL = event => {
+    if (event.url && event.url.indexOf('bitcoin:') === 0) {
+      this.navigator &&
+        this.navigator.dispatch(
+          NavigationActions.navigate({
+            routeName: 'SendDetails',
+            params: {
+              uri: event.url,
+            },
+          }),
+        );
+    }
+  };
+
+  render() {
+    return (
+      <MainBottomTabs
+        ref={nav => {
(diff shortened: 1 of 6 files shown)


ncoelho:
Cool stuff. Have you seen the wallet selection screen @marcosrdz has done recently. When you force tap on a wallet card you have a vertical list view with all wallets.

Maybe this solution can be reuse to wallet selection...

screen/send/details.js
```diff
@@ -50,10 +50,12 @@ export default class SendDetails extends Component {
...
     let fromSecret;
     if (props.navigation.state.params) fromSecret = props.navigation.state.params.fromSecret;
-    let fromWallet = {};
+    let fromWallet = null;
+
+    const wallets = BlueApp.getWallets();
 
     let startTime2 = Date.now();
-    for (let w of BlueApp.getWallets()) {
+    for (let w of wallets) {
```
Overtorment: why is it better to use const variable instead of inplace `BlueApp.getWallets()` ? curious

App.js
```diff
@@ -0,0 +1,45 @@
...
+
+    Linking.addEventListener('url', this.handleOpenURL);
+  }
+
+  componentWillUnmount() {
+    Linking.removeEventListener('url', this.handleOpenURL);
+  }
+
+  handleOpenURL = event => {
+    if (event.url && event.url.indexOf('bitcoin:') === 0) {
```
Overtorment: should be lowercased before `indexOf`. In real life capitalized URIs (like, "BITCOIN:1.....") sometimes happen

Overtorment: or not. bip21 lib wont handle `BITCOIN:` in caps...

marcosrdz:
Im currently working on such screen. No need to worry about it. Just send the address and amount like you're currently doing.

screen/send/details.js
pajasevi: Because `BlueApp.getWallets()` returns an array, so it can be stored in a variable. Calling it multiple times would hurt performance and can be avoided in the same function scope.

App.js
pajasevi: But will `BITCOIN:` open the app at all? I can add `toLowerCase` for the sake of sanitation.

pajasevi:
@marcosrdz Amazing! Let me know when the screen is ready.

PR #161, opened 2018-12-19
WIP: Spend from BIP44 wallets

pajasevi:
This is a work in progress and should allow spending from imported BIP44 wallets.
Please check if this code is valid. I am not able to test it without sending funds to the mainnet wallet.

Is there anything I missed?
Closes #149

PR #167, opened 2018-12-22
Pajasevi spend bip44

Overtorment:
re https://github.com/BlueWallet/BlueWallet/pull/161


Overtorment:
@pajasevi I dont know whether you made unit tests work (it uses secret from environment, and real mainnet wallet to create real transactions), but I rewrote tests for CI secret. added some fixes here and there. tested on mainnet - sends ok!

pajasevi:
Awesome! Will try on imported wallet when new iOS release is out. Thanks for your support!

PR #177, opened 2018-12-27
Fix: Display wallet balance consistently on all screens

pajasevi:
This is something I stumbled upon: on "reorder wallets" screen and "choose wallet" screen, the Lightning wallet displays BTC as unit and not sats - as on the home screen.

This fixes it for now. I will create a separate reusable component for wallet item so these inconsistencies won't happen anymore.


Overtorment:
Great!

PR #180, opened 2018-12-28
Refactor: Use static properties for wallet types

pajasevi:
It is very costly (performance-wise) to create new instance of wallet everytime we need to read a static property. Let's just read static field from class and access this field using native getter when reading from class instance.

This is supposed to be a non-breakable change and I tested it locally on various types of wallets with success.


Overtorment:
looks like mechanism to serialize/save/load wallets doesnt like new changes

pajasevi:
Will fix that.

pajasevi:
Awesome. Thanks for you feedback :)

PR #189, opened 2019-01-01
Update en.js

klauslovgreen:
Is that the intention?


Overtorment:
oh it should be " Is it even valid?"
without `t`

Issue #172, opened 2018-12-24
[BIP84] Add HD Segwit (bech32) wallet support


Overtorment:
Also, I'm thinking about making it a default wallet, and build all advanced features (cpfp, rbf, batching, signing) only for this wallet.

https://iancoleman.io/bip39/
https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki

Overtorment:
We have a somewhat complete test suite for HD Segwit p2sh wallets. Better start with tests for new wallet

Overtorment:
TODO:

* [ ] Create class
* [ ] TDD: test vector: add tests for internal/externall address gen
* [ ] TDD: add test for fetch balance
* [ ] implement fetch balance (Electrum)
* [ ] TDD: add test for fetch transactions
* [ ] implement fetch transactions (Electrum)
* [ ] TDD: add tests for fetch UTXO
* [ ] imlement fetch UTXO
* [ ] implement sign tx (with batching)

Overtorment:
WIP PR https://github.com/BlueWallet/BlueWallet/pull/352

Issue #170, opened 2018-12-23
Optimize Segwit HD wallet


Overtorment:
Right now it works very slow. It does lots of remote fetches, and lots of crypto operations just to operate
(for example, generating multiple addresses in hierarchy takes precious milliseconds, and those are then used in remote fetch to check balance or transactions). 
Some caching is present, but I bet theres a way to do it way more efficient.  Maybe pre-generate (warm up) some caches while user is a bit idle in the app, or pre-fetch and re-use some data from remote (when creating TX mostly, Each TX create fetches UTXO from server - how can we _reliably_ reuse UTXO and invalidate cache if smth changed..?).
Also, maybe batch some remote fetches.

PS. we are back to 'invalidating cache' classic CS problem lol

PR #192, opened 2019-01-02
Refactor: HD wallets address derivation

pajasevi:
I refactored the HD wallets, mainly Segwit P2SH, to derive addresses from XPUB every time.
There was a lot of repeated code which is now deleted so it is more readable and prone to errors.
Next step will be using Typescript.

### DIFF
--- a/class/hd-legacy-breadwallet-wallet.js
+++ b/class/hd-legacy-breadwallet-wallet.js
@@ -1,7 +1,7 @@
 import { AbstractHDWallet } from './abstract-hd-wallet';
 import Frisbee from 'frisbee';
-const bitcoin = require('bitcoinjs-lib');
-const bip39 = require('bip39');
+import bitcoin from 'bitcoinjs-lib';
+import bip39 from 'bip39';
 
 /**
  * HD Wallet (BIP39).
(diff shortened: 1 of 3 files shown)


Overtorment:
Great! Will review later. 
What I'm really interested in is performance gains. Is deriving from xpub faster..? If yes, how much?
Also, maybe you can try typescript on new type of wallet? Bech32 HD. Start with tests (generating internal and external addresses in hierarchy, generate zpub; https://iancoleman.io/bip39/  & https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki ). 


related: https://github.com/BlueWallet/BlueWallet/issues/170  https://github.com/BlueWallet/BlueWallet/issues/172

pajasevi:
@Overtorment So I created some benchmark tests: one generates 100 addresses in `for` cycle from mnemonic, the other from XPUB.
The one using mnemonic took over 14 seconds to finish, the one using XPUB took only 6.7 seconds.
I can send you the files to try it yourself if you're curious.

Yeah, it is a good idea to start with Typescript on new wallet type 👍

Overtorment:
good job!

PR #212, opened 2019-01-05
Dk danish language support

klauslovgreen:
Here is my first stab at danish language support (as well as my first pull request) - comments and suggestions are appreciated.


klauslovgreen:
Something went wrong with the en.js - that was a previous comment which I thin was already addresses.

PR #230, opened 2019-01-10
FIX: limit number of userinvoices feetched per polling request

Overtorment:


screen/lnd/lndViewInvoice.js
```diff
@@ -38,8 +38,11 @@ export default class LNDViewInvoice extends Component {
     this.fetchInvoiceInterval = setInterval(async () => {
       if (this.state.isFetchingInvoices) {
         try {
-          const userInvoices = JSON.stringify(await this.state.fromWallet.getUserInvoices());
```
Overtorment: @marcosrdz was there a specific reason why it is first serialized and then parsed..?

PR #273, opened 2019-01-25
REF: Start transition to Electrum server

Overtorment:


Overtorment:
closed in favor https://github.com/BlueWallet/BlueWallet/pull/278

PR #290, opened 2019-02-01
ADD: release notes in about screen

Overtorment:


Overtorment:
wow that was approved fast :-)
@marcosrdz you sure I didnt screw up royally? :-)

marcosrdz:
It all looked good. Standard screen addition.

PR #293, opened 2019-02-03
REF: better electrum support

Overtorment:


marcosrdz:
Yeah, this makes a lot of sense.

PR #297, opened 2019-02-04
FIX: Show error message on alert.

marcosrdz:


screen/lnd/scanLndInvoice.js
```diff
@@ -148,8 +148,7 @@ export default class ScanLndInvoice extends React.Component {
         } catch (Err) {
           console.log(Err.message);
           this.setState({ isLoading: false });
-          this.props.navigation.goBack();
```
Overtorment: u sure goBack is not needed here..?

marcosrdz: This dismisses the modal. It should display the error without taking away what the user had on screen

PR #301, opened 2019-02-06
Adds NZD as a currency

matt-auckland:
If I'm missing anything let me know.


matt-auckland:
Looks like the build broke on some encryption tests, is the build currently broken in master? Doesn't seems like something my changes would touch

PR #312, opened 2019-02-10
FIX: Impossible to scan bech32 QR invoice with a specific amount

Overtorment:
@marcosrdz if you have time test this a bit


marcosrdz:
Tested on my iPhone X. Works as intended

PR #318, opened 2019-02-14
Acinq only

marcosrdz:


Overtorment:
@marcosrdz can you do that trick with exporting your own patch and applying it to a fresh branch and commiting it with meaningfull messages..? when you overwrite your own remote branch with `--force` it is kinda a poor-mans rebase.

Issue #302, opened 2019-02-08
Android not redering QR code


ncoelho:
Some reports that android phones are not rendering/drawing the QR code

massi93:
I also have noticed this issue on a couple of phones, but not on every android one, I can give you the exact models, if this can be useful.  Maybe is related to screen resolution?

Overtorment:
Please do, cant reproduce it on Pixel on emulator

massi93:
Huawei P8 Lite and Essential Phone not rendering. Works on OnePlus 6

marcosrdz:
c615ebbc07d1c52fe5f0e2cdfcc14cb2da6096e8

PR #320, opened 2019-02-14
FIX: Android not redering QR code #302

marcosrdz:


package.json
```diff
@@ -64,7 +64,6 @@
     "react-localization": "^1.0.10",
     "react-native": "^0.58.1",
     "react-native-camera": "^1.10.0",
-    "react-native-custom-qr-codes": "^2.0.0",
```
Overtorment: that one with cool logo actualy the one who worked.
the fast ugly one is the one who failed (on display ln invoice screen)

screen/wallets/export.js
```diff
@@ -83,12 +83,13 @@ export default class WalletExport extends Component {
             if (this.state.showQr) {
```
Overtorment: if this new qr library is fast - this IF statement makes no sense

Issue #271, opened 2019-01-24
Android back button


plengkeek:
Found that the Android back button does not work when you try to go back from transaction details. Additionally, when you use the back button in the lightning market place browsing a site instead of bringing you back to the list of all the applications it closes the marketplace. Hope this is clear enough.

Blue Wallet 3.6.2 
Android 9.0.3

Overtorment:
Acknowledged.

marcosrdz:
b763f77

PR #331, opened 2019-02-16
FIX: Android back button #271 - Navigation changes to comply with har…

marcosrdz:
…dware back button.


screen/lnd/browser.js
```diff
@@ -222,16 +220,33 @@ export default class Browser extends Component {
...
     };
+    BackHandler.addEventListener('hardwareBackPress', this.handleBackButton.bind(this));
   }
 
+  componentWillUnmount = () => {
+    BackHandler.removeEventListener('hardwareBackPress', this.handleBackButton.bind(this));
+  };
+
+  handleBackButton() {
+    this.state.canGoBack ? this.webview.goBack() : this.props.navigation.goBack(null);
```
Overtorment: why is it so important to handle `canGoBack`?
what happens if browser has canGoBack:false and we try to call `.goBack` ..?

lets drop completely `this.props.navigation.goBack(null);` here

screen/lnd/browser.js
```diff
@@ -417,75 +433,71 @@ export default class Browser extends Component {
...
-              this.setState({ url: 'https://bluewallet.io/marketplace/' });
-            }}
-          >
-            <Ionicons
-              name={'ios-home'}
-              size={36}
-              style={{
-                color: this.state.weblnEnabled ? 'green' : 'red',
-                backgroundColor: 'transparent',
+          <View style={{ flex: 1, marginHorizontal: 8, alignItems: 'center', justifyContent: 'center' }}>
```
Overtorment: So, instead of an input this is now just a text..? Bad idea.
I was thinking to actually make this field editable, so users can navigate their own Lapps if they want. I think thats the way mobile Dapp browsers work (like TrustWallet)

screen/lnd/lndViewInvoice.js
```diff
@@ -96,7 +96,7 @@ export default class LNDViewInvoice extends Component {
   }
 
   handleBackButton() {
-    this.props.navigation.popToTop();
+    this.props.navigation.goBack(null);
```
Overtorment: will user land on intermediate state screen with  `goBack` ..?

screen/lnd/browser.js
marcosrdz: When its false, it means ure at the marketplace, and on android pressing back dismisses the modal.

screen/lnd/lndViewInvoice.js
marcosrdz: they will land on their  tx list

screen/lnd/browser.js
marcosrdz: Fixed.

Issue #335, opened 2019-02-17
BlueWallet should register to handle the BOLT11 lightning: URI


drmoore718:
Clicking a link that points to a lightning: URI should open BlueWallet to pay the invoice.

marcosrdz:
8398881b35e5d7ade59afefe4c06fe41ad0db714

PR #337, opened 2019-02-18
FIX: BlueWallet should register to handle the BOLT11 lightning: URI #335

marcosrdz:


screen/send/details.js
```diff
@@ -544,10 +550,10 @@ export default class SendDetails extends Component {
   };
 
   render() {
-    if (!this.state.fromWallet.getAddress) {
+    if (this.state.isLoading || typeof this.state.fromWallet === 'undefined') {
       return (
         <View style={{ flex: 1, paddingTop: 20 }}>
-          <Text>System error: Source wallet not found (this should never happen)</Text>
+          <BlueLoading />
```
Overtorment: no readable message for user..?

Overtorment:
Also, I thought we handle all schemes already

screen/send/details.js
marcosrdz: The alert displaying the lack of a wallet covers the view. Which gets dismissed immediately and the alert remains

PR #340, opened 2019-02-18
REF: main screen only 15 txs. wallet screen txs lazy-load

Overtorment:


Overtorment:
@marcosrdz pls review

PR #342, opened 2019-02-19
ADD: CHF Currency

marcosrdz:


vuongvoviet:
Can Bluewallet add VND ?. it help Vietnamese user  so much. https://api.bitcoinvn.io/#get--api-prices

PR #343, opened 2019-02-19
ADD: Chinese locale

Overtorment:


Overtorment:
Chinese added. Thanks @oday0311

PR #345, opened 2019-02-21
ADD: Tap, and hold, on balance to display a Copy balance option

marcosrdz:


package.json
```diff
@@ -85,6 +85,7 @@
     "react-native-sortable-list": "0.0.22",
     "react-native-svg": "9.1.1",
     "react-native-tcp": "3.3.0",
+    "react-native-tooltip": "^5.2.0",
```
Overtorment: not pinned

screen/wallets/transactions.js
```diff
@@ -222,7 +229,17 @@ export default class WalletTransactions extends Component {
         >
           {this.state.wallet.getLabel()}
         </Text>
-        <TouchableOpacity onPress={() => this.changeWalletBalanceUnit()}>
+        {Platform.OS === 'ios' && (
```
Overtorment: doesn't work on android..?

marcosrdz: Just added

package.json
marcosrdz: Fixed

PR #361, opened 2019-02-26
ADD: Switch Lightning wallets on send screen.

marcosrdz:
ADD: If providing an LND address in BTC send screen, throw an error.
ADD: Chain type in order to determine what type of wallet has been selected.


models/bitcoinUnits.js
```diff
@@ -3,3 +3,8 @@ export const BitcoinUnit = Object.freeze({
   SATS: 'sats',
   LOCAL_CURRENCY: 'local_currency',
 });
+
+export const Chain = Object.freeze({
```
Overtorment: nice

PR #378, opened 2019-03-05
FIX: UI fixes for  transactions list

marcosrdz:
ADD: Haptic feedbacks on error alerts

### DIFF
--- a/screen/wallets/export.js
+++ b/screen/wallets/export.js
@@ -40,13 +40,7 @@ export default class WalletExport extends Component {
     Privacy.enableBlur();
     this.setState({
       isLoading: false,
-      showQr: false,
     });
-
-    let that = this;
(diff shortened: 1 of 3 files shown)


screen/wallets/export.js
```diff
@@ -40,13 +40,7 @@ export default class WalletExport extends Component {
     Privacy.enableBlur();
     this.setState({
       isLoading: false,
-      showQr: false,
     });
-
-    let that = this;
-    setTimeout(function() {
-      that.setState({ showQr: true });
```
Overtorment: this trick actually helped the screen NOT hang, when old android device was taking its time to render the QR. this way it felt more responsive for the user to wait for the QR.
with new QR lib hope it wont be slow anymore

PR #430, opened 2019-03-22
Make patch files keep their line endings. Fix #429

### LINKED ISSUE #429: Error when applying git patches on Windows
The "prepare" doesn't work on Windows and it seems to be a line ending problem. The files in node_modules will only have LF but the patches will have CRLF (because git changes them on checkout on Windows).

davh:
This fixes a problem with applying the patches
on Windows where the patch and the files to be
patched have different format for line endings.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 *.pbxproj -text
+*.patch -text


Overtorment:
Thanks!

PR #439, opened 2019-03-29
Color cleanup

davh:
There are a lot of color codes spread out all over the app so I started by going through BlueComponent.js and cleaned it up. This should give a better overview of what colors are actually used in the app and which ones to use in new components.

It's not complete, there are a lot of colors in other files too, but it's a start.

Note: I haven't actually changed any colors in the app, it should look exactly as before.


Overtorment:
Thanks! Tests are broken. Can you look into that?

davh:
Sorry about that, should be fixed now.

I believe that the timeout error in HDWallet.test.js is unrelated to my changes but haven't looked into it in detail - have you seen this error before?

Overtorment:
Thanks!

Some Electrum server misbehaves. I'm improving work with electrum in master, not an issue

PR #464, opened 2019-04-21
FIX: buy bitcoin now happens in builtin webview instead of external b…

Overtorment:
…rowser


Overtorment:
cc @ncoelho 
safello webview - Peek 2019-04-21 19-48

PR #480, opened 2019-05-09
Support Safello redirect URLs

t4t5:
When the user buys bitcoin through Safello, they will be redirected to the Yoti app to verify their identity.

Previously, when the process was done, they would be redirected to a browser window to continue their purchase, but since bluewallet has a `bluewallet://` redirect URI associated to its
app ID, we can redirect them back to the app with a `safello-state-token` query parameter appended to the URI instead.

Bluewallet then simply needs to check if that query parameter is present, and if so, open a new WebView with the same Safello URL + `stateToken=xxx` in order to resume the user's session. This way, the whole buying process can take place inside Bluewallet's app, which is a clear improvement in terms of user experience.

What it looks like after the changes:

yoti

Note: You'll probably still want to make some changes, like passing the state token to your bluewallet.io URL instead of using my solution which opens app.safello.com directly. This is mostly to show how it should be handled. 😊

### DIFF
--- a/App.js
+++ b/App.js
@@ -95,6 +95,12 @@ export default class App extends React.Component {
     return isValidLightningInvoice;
   }
 
+  isSafelloRedirect(event) {
+    let urlObject = url.parse(event.url, true) // eslint-disable-line
+
+    return !!urlObject.query["safello-state-token"]
+  }
+
   handleOpenURL = event => {
     if (event.url === null) {
       return;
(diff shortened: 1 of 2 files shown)


Overtorment:
LGTM. Final url needs some specific parameters:

```
appId=4904c769-8a1d-43ca-8bd0-896520c94fdf
theme=bluewallet
layout=app
```

Im thinking whether its ok to link directly to safello with those params, or add another redirect via `https://bluewallet.io/buy-bitcoin-redirect.html`

Overtorment:
@marcosrdz thoughts?

t4t5:
> LGTM. Final url needs some specific parameters:
> 
> ```
> appId=4904c769-8a1d-43ca-8bd0-896520c94fdf
> theme=bluewallet
> layout=app
> ```
> 
> Im thinking whether its ok to link directly to safello with those params, or add another redirect via `https://bluewallet.io/buy-bitcoin-redirect.html`

I agree that it's better to go through https://bluewallet.io/buy-bitcoin-redirect.html (so that you don't need to set the `appId` in 2 different places). 👍

I added a commit and sent another related PR here: https://github.com/BlueWallet/Website/pull/101

Overtorment:
ok lets merge and test on real build from CI

PR #484, opened 2019-05-14
REF: reworded error msg

Overtorment:


models/signer.js
```diff
@@ -33,7 +33,7 @@ exports.createHDTransaction = function(utxos, toAddress, amount, fixedFee, chang
   }
 
   if (unspentAmountSatoshi < amountToOutputSatoshi + feeInSatoshis) {
-    throw new Error('Not enough confirmed inputs');
+    throw new Error('Not enough balance. Try sending smaller amount');
```
ncoelho: Text can be "Not enough balance. Try to send a smaller amount"

models/signer.js
```diff
@@ -88,7 +88,7 @@ exports.createHDSegwitTransaction = function(utxos, toAddress, amount, fixedFee,
...
```
ncoelho: Code is replicated in two places?

Overtorment: yep. its signer model for 2 different wallet types

models/signer.js
Overtorment: can be or should be?

Overtorment: @ncoelho

ncoelho: should be

PR #488, opened 2019-05-19
Addwalletui

Overtorment:


Overtorment:
cc @ncoelho 

image


image



image

ncoelho:
This was already done. It’s on the acinq branch I think :/

ncoelho:
It is looking a bit different from the implementation we worked on...

ezgif com-video-to-gif (4)

Overtorment:
I had to resolve conflicts and lost some styling along the way. Not an issue, important that single-address wallets are hidden from normies

Overtorment:
Acinq branch was split into ui improvements and acinq integration, the way it should be

ncoelho:
Yes, the acinq branch was my confusion.

But we lost all the improvements done overtime on this PR... it come back
to the first iteration?

- We ended up to not put the advanced settings on settings. Because users
don’t have anyway to know what the advanced settings are. So we kept them
locally.
- The styling is important, it matches the wallets cards that we changed
the colors on a previous PR.
- The goal was to remove the segwit and hd references, they are there again.

We can do the changes again later, but at least the styling should be ok,
so it doesn’t look bad

Overtorment:
Address reuse should be hidden as far away as possible from regular users, otherwise it wont pass b org review

ncoelho:
Yes, I’m totally ok with that, can be fixed later.

PR #547, opened 2019-07-18
Experimental

Overtorment:
that fixes android build, but It breaks ios build lol


Overtorment:
@marcosrdz can u pls fix ios build before merging

marcosrdz:
> @marcosrdz can u pls fix ios build before merging

Done.

PR #555, opened 2019-07-28
Add support for LNURL withdraws

fiatjaf:
Spec: https://github.com/btcontract/lnurl-rfc/blob/master/spec.md#3-withdrawing-funds-from-a-service.

It's specially useful in these days post 0-sat invoice exploit.

How it works on lightning-wallet.com, for reference:

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "appcenter": "2.1.0",
     "appcenter-analytics": "2.1.0",
     "appcenter-crashes": "2.1.0",
+    "bech32": "^1.1.3",
     "bignumber.js": "9.0.0",
     "bip21": "2.0.2",
     "bip32": "2.0.3",
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "appcenter": "2.1.0",
     "appcenter-analytics": "2.1.0",
     "appcenter-crashes": "2.1.0",
-    "bech32": "^1.1.3",
+    "bech32": "1.1.3",
     "bignumber.js": "9.0.0",
     "bip21": "2.0.2",
     "bip32": "2.0.3",
--- a/screen/lnd/lndCreateInvoice.js
+++ b/screen/lnd/lndCreateInvoice.js
@@ -97,6 +97,10 @@ export default class LNDCreateInvoice extends Component {
           throw new Error('Reply from server: ' + reply.reason);
         }
 
+        if (reply.tag !== 'withdrawRequest') {
+          throw new Error('Unsupported lnurl');
+        }
+
         // setting the invoice creating screen with the parameters
         this.setState({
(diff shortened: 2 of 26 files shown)


package.json
```diff
@@ -52,6 +52,7 @@
     "appcenter": "2.1.0",
     "appcenter-analytics": "2.1.0",
     "appcenter-crashes": "2.1.0",
+    "bech32": "^1.1.3",
```
Overtorment: pin to specific version plz

screen/lnd/lndCreateInvoice.js
```diff
@@ -32,15 +36,83 @@ export default class LNDCreateInvoice extends Component {
         const invoiceRequest = await this.state.fromWallet.addInvoice(this.state.amount, this.state.description);
         EV(EV.enum.TRANSACTIONS_COUNT_CHANGED);
         ReactNativeHapticFeedback.trigger('notificationSuccess', { ignoreAndroidSystemSettings: false });
+
+        // send to lnurl-withdraw callback url if that exists
+        if (this.state.lnurlParams) {
+          let {callback, k1} = this.state.lnurlParams;
+          let callbackUrl = callback + (callback.indexOf('?') !== -1 ? '&' : '?') + 'k1=' + k1 + '&pr=' + invoiceRequest;
+          let resp = await fetch(callbackUrl, {method: 'GET'});
```
Overtorment: we have Frisbee for http rest but fuck it, let it be. yolo

Overtorment:
LNURL button should be more subtle, users will be confused

image


make it look like an ahref, not button, and stick it to bottom border

Overtorment:
after initial fetch it makes sense to hide this button completely:

image

Overtorment:
@ncoelho  @marcosrdz  thoughts..?

screen/lnd/lndCreateInvoice.js
fiatjaf: I saw it, but I thought it was more for when you have a fixed Host and calls paths over it. Looked like an overkill here. And Frisbee imports some kind of `fetch` too. This, however, is using the default `fetch` from the standard library I believe.

fiatjaf:
photo5055338810367387789
photo5055338810367387790

ncoelho:
Replace scan lnurl with “scan to receive”. Users  shouldn’t be aware of
protocol terms. And remove the icon, that area shouldn’t have icons.

On the second screen, is the value editable by the user?

fiatjaf:
No, because the protocol response specified max=10, min=10. If it was, for example, min=5 and max=15, then it would be editable, but any edit > 10 would go back to 10 and any edit < 5 would go back to 5.

fiatjaf:
photo5055338810367387794

PR #561, opened 2019-08-03
ADD: Allow user to hide their balance.

marcosrdz:
Screen Shot 2019-08-03 at 5 30 09 PM
Screen Shot 2019-08-03 at 5 30 12 PM
Screen Shot 2019-08-03 at 5 30 19 PM


Overtorment:
```
/home/circleci/repo/screen/wallets/transactions.js
  396:7  error  'styles' is assigned a value but never used  no-unused-vars
```

always look at circleci to check why tests fail.
also `@react-native-community/blur` is not pinned.
but fuck it, lets merge, Ill fix it in master later

whiteyhat:
This looks cool! What about making the hiding balances with a blurry effect. It might be more UX-friendly. 

There are also mock-ups in #534

Issue #136, opened 2018-12-06
Add an option to empty a wallet


haarts:
I find it useful to empty a wallet now and then. Just going to the 'send' screen and add an 'empty/max' button which fills the amount so the wallet is left empty after the transaction (taking into account the fee).

ncoelho:
Proposal here:
https://app.zeplin.io/project/5b252c1057a0c54542fe1f02/dashboard?seid=5b252d0d192a7ce62415da75

<img width="815" alt="screenshot 2018-12-06 at 17 50 55" src="https://user-images.githubusercontent.com/1238429/49599219-ba358580-f980-11e8-8c24-7a0429474ff5.png">

Specifications:

- Space between all elements changed to allow more space when keyboard goes up
- New element on the price right to allow change between local currency and bitcoin (thoughts?)
- Scan element color change to follow styleguide
- Two new elements on the keyboard top
- Total balance with a soft gray
- Use total button to empty wallet

kristapsk:
This is IMO essential feature from privacy perspective. Today I decided to test BlueWallet, sent some amount to Bitcoin wallet and then wanted to transfer it all to Lightning wallet, but the only way to do it currently is to memorize the amount you have, know the number of UTXO's in a wallet, then manually calculate tx vsize / fees and specify amount. Leaving dust amounts from such transactions reduces user's privacy on the blockchain.

Overtorment:
Acknowledged

allthegoodnamesweret4ken:
I opened an issue that is not quite the same, but definitely touches the same subject:
#489
In short it proposes to send small change to your lightning wallet. No idea if this is feasible, would love to hear your thoughts.

PR #563, opened 2019-08-04
Send max

Overtorment:


Overtorment:
image

image

kristapsk:
If I understood this correctly, it resolves #136, right?

Overtorment:
Correct

PR #567, opened 2019-08-06
add proper 3 dots icon to missing elements

ncoelho:


Overtorment:
pls fix conflicts

Overtorment:
I think thats the icon that was looking strange on android, so thats why I changed it

PR #570, opened 2019-08-06
fix address position

ncoelho:
Brings address position to near the QR code instead the bottom of the page

Screenshot 2019-08-06 at 18 19 22


marcosrdz:
Does the address jump from top to center when QR finishes loading? If so, place a specific height for the loading indicator.

Overtorment:
pls fix conflicts

PR #571, opened 2019-08-06
fix wallet selector on send view

ncoelho:
Fix margins between elements, and makes both elements tapable


Overtorment:
pls fix conflict

PR #588, opened 2019-08-14
FIX: Center Cancel button

marcosrdz:


Overtorment:
this pr is called strange

PR #596, opened 2019-08-19
Keystore mvp

Overtorment:


marcosrdz:
https://github.com/pradeep1991singh/react-native-secure-key-store/blob/master/example/RNSecureKeyStoreMock

Overtorment:
stripped everything I could. other stuff can be added later with separate PRs, unless it leaks privacy.

ps. had no luck with RNSecureKeyStore as well

PR #599, opened 2019-08-21
FIX: Fix overlooked values during data migration.

marcosrdz:


class/app-storage.js
```diff
@@ -96,7 +96,7 @@ export class AppStorage {
       return false;
     }
 
-    return !!data;
+    return data === true || data === '1';
```
Overtorment: unnecessary

screen/wallets/walletMigrate.js
```diff
@@ -21,12 +21,13 @@ export default class WalletMigrate extends Component {
     try {
       const data = await AsyncStorage.getItem('data');
       if (data) {
-        const isEncrypted = (await AsyncStorage.getItem('data_encrypted')) || '';
+        const isEncrypted = (await AsyncStorage.getItem('data_encrypted')) || '0';
```
Overtorment: all other code works with that value being either '1' or ''

screen/wallets/add.js
```diff
@@ -256,6 +256,7 @@ export default class WalletsAdd extends Component {
                             }
                             A(A.ENUM.CREATED_LIGHTNING_WALLET);
                             await w.generate();
+                            await w.createID();
```
Overtorment: I dont see the need in UUID. thats extra dependency and extra code.  if you need uniq identifier of the wallet use `someHash(getSecret())`, it is unique across wallets in currently unlocked bucket

screen/wallets/walletMigrate.js
```diff
@@ -21,12 +21,13 @@ export default class WalletMigrate extends Component {
...
         await RNSecureKeyStore.set('data', data, { accessible: ACCESSIBLE.WHEN_UNLOCKED });
         await RNSecureKeyStore.set('data_encrypted', isEncrypted, {
           accessible: ACCESSIBLE.WHEN_UNLOCKED,
         });
         await AsyncStorage.removeItem('data');
+        await AsyncStorage.removeItem('data_encrypted');
```
Overtorment: thats the only legit line. Ill push it to master manually

Issue #573, opened 2019-08-07
If there is a lightning or Bitcoin address on the clipboard, you should be able to clear your clipboard


mikethetike:
[Android] Currently if I have an address copied and I switch to the app, it asks me if I'd like to create a transaction with the address. It can be a bit annoying every time I switch to the app and the only way to get rid of it is to go to another app and copy something. 

I suggest that the options in the "Would you like to use it for a transaction" dialog should be "Cancel", "OK" and "Clear it from clipboard"

ncoelho:
Thanks for reporting.

Just to make it clear, I'm without any android device at the moment. When you tap cancel, and after you enter the app again, it will ask you again if you want to create a transaction?

marcosrdz:
https://github.com/BlueWallet/BlueWallet/pull/610/commits/59c9905ae8b2106c9a4eb17f335cd63a5dcffaca

PR #610, opened 2019-08-25
FIX: Don't display send modal if address belongs to self #573

marcosrdz:


screen/receive/details.js
```diff
@@ -56,14 +56,15 @@ export default class ReceiveDetails extends Component {
       if (wallet.getAddressAsync) {
         try {
           address = await Promise.race([wallet.getAddressAsync(), BlueApp.sleep(1000)]);
+          wallet._address = address;
```
Overtorment: remove

screen/receive/details.js
```diff
@@ -56,14 +56,15 @@ export default class ReceiveDetails extends Component {
...
         } catch (_) {}
         if (!address) {
           // either sleep expired or getAddressAsync threw an exception
           console.warn('either sleep expired or getAddressAsync threw an exception');
           address = wallet._getExternalAddressByIndex(wallet.next_free_address_index);
-        } else {
-          BlueApp.saveToDisk(); // caching whatever getAddressAsync() generated internally
+          wallet._address = address;
```
Overtorment: remove

screen/receive/details.js
```diff
@@ -56,14 +56,15 @@ export default class ReceiveDetails extends Component {
...
         }
+        BlueApp.saveToDisk(); // caching whatever getAddressAsync() generated internally
```
Overtorment: why is this line marked as changed?  it should not be if you didnt change it

App.js
```diff
@@ -47,7 +48,14 @@ export default class App extends React.Component {
     if (BlueApp.getWallets().length > 0) {
       if (this.state.appState.match(/inactive|background/) && nextAppState === 'active') {
         const clipboard = await Clipboard.getString();
-        if (this.state.clipboardContent !== clipboard && (this.isBitcoinAddress(clipboard) || this.isLightningInvoice(clipboard))) {
+        const isAddressFromStoredWallet = BlueApp.getWallets().some(wallet =>
+          wallet.chain === Chain.ONCHAIN ? wallet._address === clipboard : wallet.isInvoiceGeneratedByWallet(clipboard),
```
Overtorment: instead of `wallet._address` there is a more reliable method `weOwnAddress` for HD wallets. add this method for single-address wallets as well and use it.
should be ez, just add a single method to `AbstractWallet`, smth like

```
  weOwnAddress(address) {
    return this._address === address;
  }
```

ps. this is unreliable because we might change the logic of generating receive address someday, and this wont work anymore

screen/receive/details.js
marcosrdz: Most likely GitHub ui bug

App.js
marcosrdz: Done.

Overtorment:
@marcosrdz why is it closed?
image

marcosrdz:
Merge conflitcs. Gotta fix it

PR #612, opened 2019-08-25
ADD: Auto adjust screen brightness on screens with QRCode.

marcosrdz:


package.json
```diff
@@ -108,6 +108,7 @@
     "react-native-snap-carousel": "3.8.0",
     "react-native-sortable-list": "0.0.23",
     "react-native-svg": "9.5.1",
+    "react-native-system-setting": "1.7.2",
```
Overtorment: dragging in deps from random nobodies from interwebz is a security flaw.
we need to minimize deps. or at least, fork it uder bluewallet project and pull from there.
I understand we are over capacity, but its better if you look through all the code to spot if it has any backdoors. and every time you update dep's fork to keep uk with source - diff should also be reviewed to spot backdoors or anything malicious

marcosrdz: Forked. Looked at dep's JS, iOS and Android code. No malicious code.

PR #623, opened 2019-08-30
FIX: Add warning to or restrict delete wallet option when it has balance

marcosrdz:
Screen Shot 2019-08-30 at 12 42 43 AM


marcosrdz:
Sort of like how GitHub forces you to type a repo's full name before deleting it.

PR #634, opened 2019-09-03
WIP: Batch UI

marcosrdz:
Ready to review.

### DIFF
--- a/BlueComponents.js
+++ b/BlueComponents.js
@@ -382,6 +382,36 @@ export const BlueNavigationStyle = (navigation, withNavigationCloseButton = fals
   headerBackTitle: null,
 });
 
+export const BlueCreateTxNavigationStyle = (navigation, withAdvancedOptionsMenuButton = false, advancedOptionsMenuButtonAction) => ({
+  headerStyle: {
+    backgroundColor: BlueApp.settings.brandingColor,
+    borderBottomWidth: 0,
+    elevation: 0,
+  },
+  headerTitleStyle: {
+    fontWeight: '600',
+    color: BlueApp.settings.foregroundColor,
+  },
+  headerTintColor: BlueApp.settings.foregroundColor,
+  headerLeft: (
+    <TouchableOpacity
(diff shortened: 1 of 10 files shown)


marcosrdz:
Using PR to create TestFlight build

BlueComponents.js
```diff
@@ -382,6 +382,36 @@ export const BlueNavigationStyle = (navigation, withNavigationCloseButton = fals
   headerBackTitle: null,
 });
 
+export const BlueCreateTxNavigationStyle = (navigation, withAdvancedOptionsMenuButton = false, advancedOptionsMenuButtonAction) => ({
```
Overtorment: Does it make sense starting splitting components into different files soon..?

models/bitcoinTransactionInfo.js
```diff
@@ -0,0 +1,6 @@
+export class BitcoinTransaction {
```
Overtorment: Very strange class. Can we live without it?

currency.js
```diff
@@ -111,7 +111,7 @@ function BTCToLocalCurrency(bitcoin) {
 function satoshiToBTC(satoshi) {
   let b = new BigNumber(satoshi);
   b = b.dividedBy(100000000);
-  return b.toString(10) + ' BTC';
+  return b.toString(10);
```
Overtorment: are you sure? maybe you need `loc.formatBalanceWithoutSuffix()` ?

BlueComponents.js
marcosrdz: Yup. 100%

models/bitcoinTransactionInfo.js
marcosrdz: Its used through multiple screens for Batch UI purposes. Unless there's a better approach?

currency.js
marcosrdz: loc.formatBalanceWithoutSuffix needs a little revisit. Didn't want to break it so I used satoshiToBTC since it wasn't being used anywhere else.

PR #636, opened 2019-09-03
ADD: Debug options

marcosrdz:


screen/settings/about.js
```diff
@@ -150,6 +164,24 @@ export default class About extends Component {
...
+                />
+                <BlueSpacing20 />
+                <BlueButton
+                  onPress={() => {
+                    BlueApp.setItem(AppStorage.FLAG_ENCRYPTED, '1');
+                  }}
+                  title="Storage Encrypted Flag ( ON )"
+                />
+              </>
+            )}
```
Overtorment: whats the point of all of this? you set the flag when bucket is not actually encrypted and your next launch is screwed - youre stuck on decryption state. same if bucket is encrypted and you clear the flag. i dont think next launch will work.

also, new dependency is bad, for such minor thing. and if you need to display some debug stuff there are simpler ways, like having to press `w, h = {width}, {height}` 10 times to show them. this is a common backdoor

PR #641, opened 2019-09-13
Update to BitcoinJS 5.1.6

junderw:
TODO:

- Should we also migrate non-segwit to PSBT? This will require fetching the full raw tx being referenced for each input being signed.


junderw:
Also, after running tests 2-3 times I started getting rejected by the API.

Rate limiting is pretty prohibitive for devs doing testing. Perhaps consider creating a docker image that runs with regtest or something.

ncoelho:
All the latest developments are being done for bip49 and 84. It would be ok
to start there for now

junderw:
TransactionBuilder has been replaced with Psbt.

Some tests were creating transactions with total output value greater than total input value.

PSBT checks for this, so I needed to fix the tests.

Let me know if you want to use Psbt for multisig features etc. The multisig flow using PSBT is insanely easy (the only hard part would be managing all the xpubs between participants.)

It also makes offline signing more secure for non-segwit...

junderw:
(However, right now non-segwit is not using PSBT)

Overtorment:
Terrific work, thanks!
I had to maintain 2 bitcoinjs versions (v3 and latest, v5) because I was not sure how to migrate old code properly, and decided to roll with backwards compatibility for some time. I'm relieved to know that this is now sorted.

Some integration tests are failing when API limits reached - thats okay. APIs are used in older wallets only, and we will eventually rewrite them to work through electrum as well (we will soon make BIP84 a default and flagship wallet for everyone, and it already works through electrum. All new features will be made for BIP84 exclusively).

If would be awesome if you could prepare multisig in bip84 wallet class @junderw. Some methods we could jump in and use right away. @marcosrdz @ncoelho and I will do all the work with UI and joining the party ceremony.

>> Should we also migrate non-segwit to PSBT?

Not sure, probably not, those wallets are in maintenance mode.

Oh, I'm quite excited to have PSBT as I plan integration with completely airgapped HW wallet. I thinks its a most optimal way to roll - mobile app has access to xpub and creates transactions, airgapped hw wallet signs it and gives back to mobie app, then mobile app broadcasts. Perfect balance between convenience and healthy paranoia. Probably multisig involved.
Any guidance on that, @junderw ? Maybe some helper methods can be introduced in bip4 class?

Overtorment:
Note: need to test it a bit before merging to master

junderw:
well, there are two ways to do multisig:

- coordinated, which requires management from the server to delegate addresses and exchange psbts between peers, lock utxos currently in the signing process, etc.
- uncoordinated, which requires no management, but the number of addresses that need to be checked increases and it is much easier for multiple PSBTs using the same utxo causing problems and re-signing for users.

Airgap is very straightforward, the only things that might need fixing are adding BIP32 path info to the inputs to help with signing.

With path info, signing is extremely simple.

PR #647, opened 2019-09-17
Ln send

ncoelho:
Some visual enhancements to the send screen on ln


ncoelho:
@Overtorment unused var removed

PR #667, opened 2019-09-29
ADD: Refill with external wallet

marcosrdz:


screen/receive/details.js
```diff
@@ -70,15 +70,12 @@ export default class ReceiveDetails extends Component {
           address: address,
           addressText: address,
         });
-      } else if (wallet.getAddress) {
-        address = wallet.getAddress();
+      } else if (wallet.getAddress && address.trim().length === 0) {
```
Overtorment: cant understand this

marcosrdz: getAddress() returns empty string on lightning wallet. However, we already have the address from previous code. You are more familiar with this area though

PR #771, opened 2019-12-21
Decode Lightning invoices locally.

fiatjaf:
This is a simple computation involving a small library that will save a roundtrip to the server every time a Lightning invoice is read.

If this is merged Lndhub's API can also deprecate the decode endpoint.

PR #772, opened 2019-12-22
lnurl-pay support

fiatjaf:
Implementation of the [lnurl-pay spec](https://github.com/btcontract/lnurl-rfc/blob/master/spec.md#4-lnurl-pay).

Changes here:

* Includes the [js-lnurl](https://www.npmjs.com/package/js-lnurl) library and uses it to improve the lnurl-withdraw flow a little.
* Adds lnurl type detection and parameter fetching to the `App.js` url router.
* Adds support for scanning lnurl-pay addresses from the `Send` screen.
  * That shows a form with plaintext and optionally image metadata about the lnurl;
  * Amounts can be either fixed or editable within a range (there's a debounced function that brings them back into the range -- also added that on the lnurl-withdraw flow which had a broken text input for most cases);
  * Once the user presses "Pay" the invoice is fetched from the lnurl-provider, checked against amount and metadata hash and paid if all is ok;
  * The `LnurlPaySuccessAction` screen is shown afterwards. Each of the the 4 types of action are supported (nothing, a simple message, an URL or a secret message encrypted with the payment preimage).
* When browsing a paid invoice from the transactions list, a "metadata" button is shown that takes the user to the `LnurlPaySuccessAction` screen again.
  * From that screen, when viewed from the transactions list, the user can call the lnurl-pay again so some payment flows that make sense to be repeated can be (in the future this behavior could be used to build an automatic and personalized in-app marketplace based on repeating custom user lnurl-pay calls -- like refilling your phone, ordering the same pizza, recharging your subway card, paying your monthly bill of something etc.).
* To enable the historical view of lnurl payments, `AsyncStorage` is used to keep payment data (indexed by the payment_hash) and metadata (indexed by the description_hash).
  * To not fill up the entire user phone with this, lnurl-pay entries (prefix `lp:`) older than 30 days in `AsyncStorage` are deleted.

Below it's a very bad screencast of the browsing part and repeating a payment (couldn't make a screencast of the actual act of scanning an lnurl QR because my phone is old and my screencast recorder crashes when the scanner is open for some reason):

output

This branch builds on top of https://github.com/BlueWallet/BlueWallet/pull/771, but that is not necessary. If that PR gets rejected I can rebase this.


Overtorment:
Wow that’s a big diff, thanks!
I don’t have time to review immediately, but I definitely will when I have
a chance

ncoelho:
Wow, awesome. Gonna need some UX love as well

fiatjaf:
Thank you for looking at it.

While looking at UI stuff maybe you could have a better decision as to where show the metadata for the paid invoice, if in the same invoice screen or a different place; if (and how without affecting performance -- although it should be a small hit nonetheless) to show lnurl-pay stuff at the invoices list; if we should ignore the image parameter or not. Anyway, you'll know better.

Also I think once better real-world examples are available to try this against we'll have better insight on the best UI.

marcosrdz:
@fiatjaf  conflict.

fiatjaf:
@marcosrdz resolved.

Overtorment:
oh hey, can you pls resolve conflicts? we are going to work on this

ncoelho:
I tested it here https://github.com/BlueWallet/BlueWallet/pull/1136

Solved some conflicts, but pretty sure I messed up somewhere. It would be nice for @fiatjaf to check them from scratch. And the current status from lnurl-withdraw in production. I got some weird messages lately.

The solution is working pretty well, we don't really need to do major changes to have this ready for production. Just needs a proper review from @Overtorment after the conflicts are solved.

fiatjaf:
Ok, I try to fix everything and so on.

This PR I think makes some improvements on lnurl-withdraw too, I think, I'll pay attention and hopefully things will improve.

fiatjaf:
Ok, there are way too many conflicts, this will take a while.

Overtorment:
probably easier to do it from scratch? and manually port chunks of code in new branch

Issue #802, opened 2019-12-31
README.md building on console error


nickycutesc:
I'm following the directions to build BlueWallet on the console:

git clone https://github.com/BlueWallet/BlueWallet.git
cd BlueWallet
npm install

However, I'm running into the following error:

node-pre-gyp WARN Using request for node-pre-gyp https download 
node-pre-gyp WARN Tried to download(404): https://fsevents-binaries.s3-us-west-2.amazonaws.com/v1.2.9/fse-v1.2.9-node-v79-darwin-x64.tar.gz 
node-pre-gyp WARN Pre-built binaries not found for fsevents@1.2.9 and node@13.5.0 (node-v79 ABI, unknown) (falling back to source compile with node-gyp) 
Traceback (most recent call last):
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/gyp_main.py", line 50, in <module>
    sys.exit(gyp.script_main())
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 554, in script_main
    return main(sys.argv[1:])
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 547, in main
    return gyp_main(args)
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.py", line 532, in gyp_main
    generator.GenerateOutput(flat_list, targets, data, params)
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/generator/make.py", line 2215, in GenerateOutput
    part_of_all=qualified_target in needed_targets)
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/generator/make.py", line 802, in Write
    self.WriteCopies(spec['copies'], extra_outputs, part_of_all)
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/generator/make.py", line 1145, in WriteCopies
    env = self.GetSortedXcodeEnv()
  File "/usr/local/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/generator/make.py", line 1885, in GetSortedXcodeEnv
    additional_settings)
[...]

PR #803, opened 2020-01-02
updating instructions for console compilation

nickycutesc:
refer here for more information:

https://github.com/BlueWallet/BlueWallet/issues/802

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,22 @@ cd BlueWallet
 npm install
 ``` 
 
+Please make sure that your console is running the most stable versions of npm and node (even-numbered versions).
+
+```
+npm --version && node --version
+```
+
+If not, please change your node version in the shell, assuming you have installed node and npm via the Homebrew Package Manager.
+
+```
+brew install node@12
+brew unlink node
+brew link node@12
+```
+
+For more information and background information, please refer to this link: https://apple.stackexchange.com/questions/171530/how-do-i-downgrade-node-or-install-a-specific-previous-version-using-homebrew
+
 * To run on Android:
 
 You will now need to either connect an Android device to your computer or run an emulated Android device using AVD Manager which comes shipped with Android Studio. To run an emulator using AVD Manager:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,6 +38,22 @@ cd BlueWallet
 npm install
 ``` 
 
+Please make sure that your console is running the most stable versions of npm and node (even-numbered versions).
```
bolatovumar: "your console" here is not appropriate as you are not running a specific version of console software but rather specific version of Node and npm.

I would adjust this to say something like:

> Please refer to the `engines` field in `package.json` file for the minimum required versions of Node and npm. It is preferred that you use an even-numbered version of Node as these are LTS versions.

I would also place this line above the line that tells the user to run `npm install` as the version of Node and npm matter for the `npm install` command.

README.md
```diff
@@ -38,6 +38,22 @@ cd BlueWallet
...
+If not, please change your node version in the shell, assuming you have installed node and npm via the Homebrew Package Manager.
+
+```
+brew install node@12
+brew unlink node
+brew link node@12
+```
+
+For more information and background information, please refer to this link: https://apple.stackexchange.com/questions/171530/how-do-i-downgrade-node-or-install-a-specific-previous-version-using-homebrew
+
```
bolatovumar: I would remove the above lines (42 to 56) as they are specific to MacOS and require the user to have `brew` installed which not everyone may use. It's best to leave it up to the user how they want to install the required dependencies.

nickycutesc: Noted.

README.md
nickycutesc: Noted.

nickycutesc:
I'm not 100% sure why the ci/circleci test is failing, especially since I haven't touched any code @bolatovumar . Should I have made the updates to a new PR?

marcosrdz:
The circle ci failure is related to one of the unit tests we haven’t updated.

nickycutesc:
Ok, so I shouldn't be worried about this error, correct?

marcosrdz:
> Ok, so I shouldn't be worried about this error, correct?

Correct.

Overtorment:
Thanks!

Issue #687, opened 2019-10-10
useHooks


marcosrdz:
Migrate existing codebase to useHooks as we go along fixing defects/implementing features.

PR #906, opened 2020-03-23
rewrite currency screen using functional components

skv-headless:
related #687


screen/settings/currency.js
```diff
@@ -7,84 +7,73 @@ import { FiatUnit } from '../../models/fiatUnit';
 let loc = require('../../loc');
 let currency = require('../../currency');
 
-export default class Currency extends Component {
-  static navigationOptions = () => ({
-    ...BlueNavigationStyle(),
-    title: loc.settings.currency,
-  });
+const data = Object.values(FiatUnit);
```
skv-headless: removed data from state since it is not updating

PR #910, opened 2020-03-26
[WIP] begin with typescript

### LINKED ISSUE #141: Use Flow
use flow. Flow will allow us to continue writing pure JS and add type checking. TypeScript requires us to learn special syntaxes and have it compile back to pure JS.

skv-headless:
closes #141


class/abstract-wallet.js
```diff
@@ -92,7 +92,7 @@ export class AbstractWallet {
     return true;
   }
 
-  allowSendMax(): boolean {
```
skv-headless: What types here for?

Overtorment: they help itelij webstorm with autocomplete

class/deeplinkSchemaMatch.js
```diff
@@ -5,7 +5,7 @@ import RNFS from 'react-native-fs';
 import url from 'url';
 import { Chain } from '../models/bitcoinUnits';
 const bitcoin = require('bitcoinjs-lib');
-const BlueApp: AppStorage = require('../BlueApp');
```
Overtorment: whats wrong with typehinting here? its specifically for intelij

skv-headless: typescript throws an error that types are not allowed in javascript. I can try to shut down this error if needed

Overtorment: please do, that thing helps with code navigation and autocomplete in IDE

Overtorment:
this PR needs tests to be fixed and conflict resolved

screen/settings/about.tsx
```diff
@@ -135,7 +135,7 @@ const About = () => {
           <BlueTextCentered>
             {getApplicationName()} ver {getVersion()} (build {getBuildNumber()})
           </BlueTextCentered>
-          <BlueTextCentered>{new Date(getBuildNumber() * 1000).toGMTString()}</BlueTextCentered>
+          <BlueTextCentered>{new Date(parseInt(getBuildNumber()) * 1000).toUTCString()}</BlueTextCentered>
```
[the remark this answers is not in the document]
skv-headless: getBuildNumber returns string. We can't multiple string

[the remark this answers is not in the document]
Overtorment: image
looks like we can. build number is provided by CI/CD (appcenter.ms) and is always a timestamp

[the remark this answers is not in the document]
skv-headless: I mean in typescript we can't do it https://www.typescriptlang.org/play/index.html#code/EQCwpgNhD2wAQCo4EYAMqg

here https://github.com/react-native-community/react-native-device-info/blob/master/src/index.ts#L323
return type is string 
`let buildNumber: string;`

class/deeplinkSchemaMatch.js
skv-headless: I get this error. Typescript doesn't allow types in js files.
```
screen/wallets/add.js:37:14 - error TS8010: Type annotations can only be used in TypeScript files.
37 let BlueApp: AppStorage = require('../../BlueApp');
```
I didn't find how to suppress specific error in ts and I can't use `//@ts-ignore` in js files.

So the best solution that I found is to rename files with types to ts and ignore all errors there.

Overtorment:
ts support was added in https://github.com/BlueWallet/BlueWallet/pull/3403

PR #958, opened 2020-04-07
Feature/broadcast page

dr3am3r:


ncoelho:
Try to use the same header style we are using. See the view where the feature sits as an example

Overtorment:
`/` in branch name breaks Glados.
here's the appetizer link: https://appetize.io/app/37c893hphpb9eb0ez9u67agz54

dr3am3r:
> Try to use the same header style as the other views. See the view where the feature sits as an example

done

Overtorment:
closed in favor of https://github.com/BlueWallet/BlueWallet/pull/1000

PR #996, opened 2020-04-16
ADD: support for Electrum Seed format, legacy (closes #954)

### LINKED ISSUE #954: Bluewallet fails to import Electrum wallet (Legacy Addresses) from wallet's seed and key
I recently got Bluewallet and it fails to import my Electrum wallet from the wallet's seed or key. This would be an extremely useful feature to have as most of my wallets are on Electrum.

The wallet type is BIP32 with multiple addresses (legacy addresses).

I have not tried importing a SegWit address yet, but I do not use SegWit addresses. I prefer to use legacy addresses as they are more compatible.

I am not sure if this is a bug or a compatibility issue but it would be good to have working compatibility with Electrum wallets.

Overtorment:


GladosBlueWallet:

 [android in browser] https://appetize.io/app/vkr48w5dcpag3b3g53rgnxzwmw

[download apk](http://filestorage.bluewallet.io:1488/1587048731.apk)

PR #1040, opened 2020-05-01
Wrap RNRandomBytes in module

tanx:
@Overtorment thanks for your great work on BlueWallet. This PR wraps the native `RNRandomBytes` in a separate module for easier mocking. This also allows removal of the testing case from your production build and ensures the correct RNG is always used. Hope this is useful :)


tanx:
Also removed the `global.crypto` assignment from test in a9714c6

Overtorment:
lets run all tests an builds properly here https://github.com/BlueWallet/BlueWallet/pull/1053

Overtorment:
noice, all tests passed

random.js
```diff
@@ -0,0 +1,24 @@
+/**
```
Overtorment, line 1: lets move this file under `blue_modules/` as par of the refactoring I had in mind for a while

tests/setup.js
```diff
@@ -1,3 +1,12 @@
+import crypto from 'crypto';
+import { promisify } from 'util';
+
+jest.mock('../random', () => {
```
Overtorment, line 4: i see what you did here, and its a legit approach, but i would like to keep it isomorphic, where all classes work out of the box at least in RN and nodejs. this mock will make `randomBytes` work _only_ under tests.

lets put in `random.js` a platform check and call either `randomBytes` for react-native or for nodejs/cli.
i belive recommended RN check is  `typeof navigator !== 'undefined' && navigator.product === 'ReactNative'`

Overtorment:
noice, all tests passed. solid contribution, thanks! some changes required tho

tanx:
@Overtorment thanks for your feedback. I changed the PR to use the node `crypto` api directly so things stay isomorphic. I also moved the new module to `blue_modules/crypto.js` and wired that up in the package.json to be used for the react-native case. Let me know if that works for you. It's probably cleaner this way than a runtime check. I've seen too many security issues in JS libs caused by accidentally invoking insecure apis.

tanx:
Another option might be to use https://github.com/tradle/react-native-crypto instead of the new `blue_modules/crypto.js` module. But there are less dependencies this way.

Overtorment:
travis build failed with

```
Error: Unable to resolve module `crypto` from `class/legacy-wallet.js`: 
```

tanx:
> travis build failed with

Oops. Looks like local modules are resolved using the package.json `name`. Should be fixed in https://github.com/BlueWallet/BlueWallet/pull/1040/commits/816bab6c97797731d360402756eef9f33e7d749d

Overtorment:
https://github.com/BlueWallet/BlueWallet/pull/1056

Overtorment:
merged in https://github.com/BlueWallet/BlueWallet/pull/1056

PR #1056, opened 2020-05-08
Tanx randombytes

Overtorment:


junderw:
why use promisify?

why not just write your helper function to use a promise?

Overtorment:
@tanx why?

GladosBlueWallet:

 [android in browser] https://appetize.io/app/t9t2p84nxmm8cfemknfnjxby4c

[download apk](http://filestorage.bluewallet.io:1488/1588934829.apk)

blue_modules/crypto.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+exports.randomBytes = (size, callback) => {
+  RNRandomBytes.randomBytes(size, (err, bytes) => {
+    if (err) {
+      callback(err);
+    } else {
+      callback(null, Buffer.from(bytes, 'base64'));
+    }
+  });
+};
```
junderw, line 23:
```suggestion
exports.randomBytes = async size =>
  new Promise((resolve, reject) => {
    RNRandomBytes.randomBytes(size, (err, bytes) => {
      if (err) {
        return reject(err);
      } else {
        return resolve(Buffer.from(bytes, 'base64'));
      }
    });
  });
```

class/abstract-hd-electrum-wallet.js
```diff
@@ -1,7 +1,8 @@
-import { NativeModules } from 'react-native';
 import bip39 from 'bip39';
 import BigNumber from 'bignumber.js';
 import b58 from 'bs58check';
+import crypto from 'crypto';
+import { promisify } from 'util';
```
junderw, line 5:
```suggestion
```

class/abstract-hd-electrum-wallet.js
```diff
@@ -72,26 +71,8 @@ export class AbstractHDElectrumWallet extends AbstractHDWallet {
...
-
-      // RN environment
-      RNRandomBytes.randomBytes(32, (err, bytes) => {
-        if (err) throw new Error(err);
-        let b = Buffer.from(bytes, 'base64').toString('hex');
-        that.secret = bip39.entropyToMnemonic(b);
-        resolve();
-      });
-    });
+    const buf = await promisify(crypto.randomBytes)(32);
```
junderw, line 74:
```suggestion
    const buf = await crypto.randomBytes(32);
```

class/legacy-wallet.js
```diff
@@ -45,34 +44,8 @@ export class LegacyWallet extends AbstractWallet {
...
-        that.secret = bitcoin.ECPair.makeRandom({
-          rng: function(length) {
-            let b = Buffer.from(bytes, 'base64');
-            return b;
-          },
-        }).toWIF();
-        resolve();
-      });
-    });
+    const buf = await promisify(crypto.randomBytes)(32);
```
junderw, line 47:
```suggestion
    const buf = await crypto.randomBytes(32);
```

class/legacy-wallet.js
```diff
@@ -1,9 +1,8 @@
+import crypto from 'crypto';
+import { promisify } from 'util';
```
junderw, line 2:
```suggestion
```

tanx:
> why use promisify? why not just write your helper function to use a promise?

This is was originally proposal in https://github.com/BlueWallet/BlueWallet/pull/1040/commits/52a8f530c669de1fcf968ddfb4b21bf822865cd0

But if we want to stay isomorphic as requested in https://github.com/BlueWallet/BlueWallet/pull/1040#discussion_r420074454 without doing platform specific code branches. Mocking the node `crypto` api and leveraging the metro packager via the `“react-native”` attribute in package.json seems like the cleanest solution.

I would propose going this route with `net` and `tls` in the electrum client package as well. That way we can get rid of the globals, which prevent using es6 import instead of require throughout.

blue_modules/crypto.js
tanx: Unfortunately incompatible to https://nodejs.org/api/crypto.html#crypto_crypto_randombytes_size_callback

junderw:
wait... how are Promises not isomorphic?

Node, React Native, all modern browsers support `new Promise()` out of the box? No?

I'm not too versed in React and React Native.

tanx:
Promises are of course supported everywhere. It’s just that we use the metro packager to import our `crypto` implementation in https://github.com/BlueWallet/BlueWallet/pull/1056/commits/816bab6c97797731d360402756eef9f33e7d749d .

Have you tested your suggestion? You’ll see if you run it that way it will not run under node with `npm test` since the native apis expect a callback. See also
https://nodejs.org/api/crypto.html#crypto_crypto_randombytes_size_callback

junderw:
I ran all the tests with those suggestions, and they all passed (though a few were skipped due to some ENV vars)

In node `await crypto.randomBytes(32)` will return 32 random bytes due to 2 facts.

1. Node crypto randomBytes will perform Synchronusly if a callback is not passed as the second arg.
2. A synchronus function with an `await` in front of it is still valid and will return immediately (await is ignored) without freeing the event loop.

tanx:
> I ran all the tests with those suggestions, and they all passed (though a few were skipped due to some ENV vars)
> 
> In node `await crypto.randomBytes(32)` will return 32 random bytes due to 2 facts.
> 
> 1. Node crypto randomBytes will perform Synchronusly if a callback is not passed as the second arg.
> 2. A synchronus function with an `await` in front of it is still valid and will return immediately (await is ignored) without freeing the event loop.

It's true that the node api also works with synchronous style in this particular case. The problem is react native modules cannot use synchronous calls since the native modules apis go over the react native bridge. That means we would rely on different semantics synchronous/blocking for node and async for react native. Which seems like a potential bugs waiting to happen. Why not just use the same semantics for both platforms and prevent the headache?

junderw:
> react native modules cannot use synchronous calls

In my suggestion, they aren't.

That's why I suggested `const buf = await crypto.randomBytes(32)`

If React-Native: it uses the module in blue_modules/crypto.js
If Node: it uses node crypto

Both of the above modules return 32 random bytes when `const buf = await crypto.randomBytes(32)` is called.

I don't see the problem. React Native is using the asynchronous call.

The whole reason why `await` doesn't throw on synchronous code was specifically for isomorphic usage between two environments that use sync and async but with the same call signature.

I mean. It's not really a big deal to me. At this point I'm arguing for the sake of pointing out why I'm not wrong here, do whatever you want.

Also, there's a case to be made about readability. `await promisify(crypto.randomBytes)(32)` makes sense to the person who wrote it and people who use promisify a lot, but newer people reading through the code to make sense of what is going on might get confused.

class/abstract-hd-electrum-wallet.js
junderw, line 5: Maybe something like:

```suggestion
import { promisify } from 'util';
const randomBytes = promisify(crypto.randomBytes);
```

Near the top, and instead of `await crypto.randomBytes(32)` just do `await randomBytes(32)`

tanx:
> I mean. It's not really a big deal to me. At this point I'm arguing for the sake of pointing out why I'm not wrong here, do whatever you want.

Well I appreciate your input. But I agree this discussion isn’t really very productive either.

My intention was to clean up the code by providing a shim that is compatible with the node api. If you guys prefer to return a promise instead of the node semantics then I can change it.

tanx:
Ok. Updated the original PR in https://github.com/BlueWallet/BlueWallet/pull/1040/commits/4a041d061b3c8775be2a33805b40ec12f5b85f4c

tanx:
Just a heads up. If the wallet generate code is used with a different `crypto` implementation like [tradle/react-native-crypto](https://github.com/tradle/react-native-crypto) it will fallback to using sjcl's PRNG since no callback is provided. See: 

https://github.com/mvayngrib/react-native-randombytes/blob/master/index.js#L39-L46

GladosBlueWallet:

 [android in browser] https://appetize.io/app/b8yx6enc16rwnpyreva943d3ym

[download apk](http://filestorage.bluewallet.io:1488/1589009430.apk)

junderw:
ok, I'm not used to React, but here's an idea I have pushed to this branch and my reasoning behind it.

1. 'crypto' imports should be in one place on the app side. So I made an rng.js file. This seems a lot better than potentially having multiple files calling into a library and someone adding a `'./'` and a `crypto.js` file and suddenly what we thought was a secure rng is suddenly insecure.
2. I moved the promisify step to the rng.js file.

Hearing about how all these standard modules are getting shimmed from all these different places I think keeping the calls to those crucial libraries in one file where we won't have to go hunting them down if some shim breaks or something happens that makes our usage insecure. Also, shimming utils might seem harmless... so someone might shim in a malicious utils that will promisify our crypto into a weak RNG or something... so I opted against using it.

class/rng.js
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+import crypto from 'crypto';
+
+/**
+ * Generate cryptographically secure random bytes using native api.
+ * @param  {number}   size      The number of bytes of randomness
+ * @return {Promise.<Buffer>}   The random bytes
+ */
+exports.randomBytes = async size =>
```
tanx, line 13:
```suggestion
export async function randomBytes(size) {
```

class/rng.js
```diff
@@ -0,0 +1,19 @@
...
+  new Promise((resolve, reject) => {
```
tanx, line 14:
```suggestion
  return new Promise((resolve, reject) => {
```

class/rng.js
```diff
@@ -0,0 +1,19 @@
...
+    crypto.randomBytes(size, (err, data) => {
+      if (err) return reject(err);
+      return resolve(data);
+    });
+  });
```
tanx, line 19:
```suggestion
  });
}
```

class/rng.js
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Generate cryptographically secure random bytes using native api.
+ * @param  {number}   size      The number of bytes of randomness
+ * @return {Promise.<Buffer>}   The random bytes
+ */
+exports.randomBytes = async size =>
+  new Promise((resolve, reject) => {
+    crypto.randomBytes(size, (err, data) => {
+      if (err) return reject(err);
```
tanx, line 16:
```suggestion
      if (err) reject(err);
```

class/rng.js
```diff
@@ -0,0 +1,19 @@
...
+      return resolve(data);
```
tanx, line 17:
```suggestion
      else resolve(data);
```

class/rng.js
tanx: To clarify ... I used `exports.randomBytes` in my `crypto.js` shim since the node api used commonjs. But no reason not to use esm syntax for `rng.js`.

junderw:
FYI you can make multiline suggestions by clicking the plus sign and dragging downward to the end, then clicking the suggestion link.

tanx:
Ah gotcha thanks. First time I used that feature.

GladosBlueWallet:

 [android in browser] https://appetize.io/app/c8n7a14zb08ctbv6xb6a7zdppm

[download apk](http://filestorage.bluewallet.io:1488/1589013534.apk)

GladosBlueWallet:

 [android in browser] https://appetize.io/app/566pheu0jfg06brq54xpcbppac

[download apk](http://filestorage.bluewallet.io:1488/1589194306.apk)

Overtorment:
thanks!

PR #1065, opened 2020-05-10
Fix npm install

lukechilds:
Fixes:

```
npm ERR! code 128
npm ERR! Command failed: git checkout 38332f629f577cdd57c69fc8cc971b3cbad193c9
npm ERR! fatal: reference is not a tree: 38332f629f577cdd57c69fc8cc971b3cbad193c9
npm ERR!
```

Looks like the specified commit for `marcosrdz/react-native-secure-key-store` has been removed and now it can't be resolved.

I've done a quick check and the updated tip of master commit appears to still work ok. Not super familiar with how it's used though so would be good for someone more familiar to confirm.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11826,8 +11826,8 @@
       }
     },
     "react-native-secure-key-store": {
-      "version": "git+https://github.com/marcosrdz/react-native-secure-key-store.git#38332f629f577cdd57c69fc8cc971b3cbad193c9",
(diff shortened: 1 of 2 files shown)


marcosrdz:
I was debugging a bug that relies on that repo. Made the mistake of doing push force after updating it, resulting in commit hash being removed.

Overtorment:
@marcosrdz move this repo under BW organization plz

PR #1075, opened 2020-05-14
FIX: Line break on tx list

ncoelho:


GladosBlueWallet:

 [android in browser] https://appetize.io/app/b2cy3p7x66v2qpt6w2q8rndm50

[download apk](http://filestorage.bluewallet.io:1488/1589482666.apk)

ncoelho:
> Check if u can avoid being super explicit with fontSize. Some users may have dynamic type and this would affect their experience

Doesn't dynamic type override font sizes? How do you suggest to go about it, no font size defined or is there a dynamic way to define the font size...

PR #1101, opened 2020-05-19
FIX: Reduced batchsize to prevent "undefined" error when retrieving transactions

mgurzenski:
This batchsize change allowed me to retrieve transactions for the following 2 wallet addresses without receiving an undefined exception.
34xp4vRoCGJym3xR7yCVPFHoCNxv4Twseo
18rnfoQgGo1HqvVQaAN4QnxjYE7Sez9eca

### DIFF
--- a/BlueElectrum.js
+++ b/BlueElectrum.js
@@ -339,7 +339,7 @@ module.exports.multiGetHistoryByAddress = async function(addresses, batchsize) {
 };
 
 module.exports.multiGetTransactionByTxid = async function(txids, batchsize, verbose) {
-  batchsize = batchsize || 40;
+  batchsize = batchsize || 49;
   // this value is fine-tuned so althrough wallets in test suite will occasionally
   // throw 'response too large (over 1,000,000 bytes', test suite will pass
   verbose = verbose !== false;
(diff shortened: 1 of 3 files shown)


Overtorment:
What exactly was throwing undefined?

Overtorment:
And have you tried values between 40 and 60?

GladosBlueWallet:

 [android in browser] https://appetize.io/app/abudc4dyutfnpf6pd2zzdmpquw

[download apk](http://filestorage.bluewallet.io:1488/1589927770.apk)

mgurzenski:
> What exactly was throwing undefined?

When legacy-wallets.fetchTransactions() is looping through the values returned by BlueElectrum.multiGetTransactionByTxid() into the "txdatas" variable, some values are undefined when the batchsize is 61. 

image

mgurzenski:
> And have you tried values between 40 and 60?

49 was the largest batchsize that worked for one of my test wallets. My thought was to set it to 40 to give a decent buffer.

Overtorment:
Im guessing the issue is the same as I had once, batch reply from electrum is too big.
this is very rare case, so lets keep batchsize as big as possible, until next time we will have to revise it. 

ps. huge transactions are usually batched payouts from exchanges, and my precious value of 61 already accounted for some of the biggest txs Ive seen.

mgurzenski:
> Im guessing the issue is the same as I had once, batch reply from electrum is too big.
> this is very rare case, so lets keep batchsize as big as possible, until next time we will have to revise it.
> 
> ps. huge transactions are usually batched payouts from exchanges, and my precious value of 61 already accounted for some of the biggest txs Ive seen.

That makes sense. Thanks for the background!

Overtorment:
thanks for the pr!

PR #1113, opened 2020-05-22
FIX: Status bar color on main view

ncoelho:


package.json
```diff
@@ -142,7 +142,7 @@
     "secure-random": "1.1.2",
     "stream-browserify": "2.0.2",
     "url": "0.11.0",
-    "util": "0.12.3",
+    "util": "0.12.1",
```
Overtorment, line 145: you branched out from wrong commit and trying to revert latest changes in master. redo from scratch

Issue #905, opened 2020-03-23
Get rid of Frisbee


Overtorment:
rewrite on react native fetch and remove dependency

Overtorment:
actually, now I think its best to just update frisbee.
frisbee is isomorphic, works nicely in rn & nodejs cli, and potentially in browser (if we ever need it)

limpbrains:
I want to work on this
Should be pretty straightforward
According to https://github.com/niftylettuce/frisbee/releases frisbee 3 doesn't have any breaking changes
plan:
- upgrade package
- make sure tests works
- test some functionality which depends on it. Hodl Hodl integration, lightning

Correct?

PR #1116, opened 2020-05-22
FIX: bump frisbee package to 3.1.2

limpbrains:
Fixing #905 
upgraded frisbee version from v2.0.9 to v3.1.2
- tests work
- checked lightning network screen
- checked hodlhodl screen
Everything works fine


Overtorment:
very noice! i was afraid we would have to rewrite all calls because frisbee API changed

Issue #139, opened 2018-12-06
Separate styles from inline JSX styles


ncoelho:
After releasing our next iOS update, code clean up and refactoring should be a priority. This will reduce bugs, and allow us to grow the app without having to write “cryptic” code to get around bugs. Here is the list of things I propose:

**Separate styles from inline JSX styles**

limpbrains:
I'll work on this

PR #1123, opened 2020-05-24
REF: move all inline styles to StyleSheet.create 💅

limpbrains:
Related to #139

At first, I wanted to move only styles with 3 and more rules to static. But it still doesn't feel right.
So then I've discovered this nice [codemod](https://github.com/ignacioola/react-native-fix-inline-styles) and converted all the styles.

I understand this probably too much and you will want to exclude some screens from this patch, no problem.

Personally I prefer this style to be forced everywhere. It helps with readability of huge render functions. 

I haven't changed BlueComponents. They should be treated differently, each component probably should has it's own set of styles, not the global value.

Also, I've run `eslint --fix` on every file I've changed.

PS
I've found out that almost all usage of `SafeBlueArea` comes with `<SafeBlueArea style={{ flex: 1 }}>`. Probably I can make it default ?


Overtorment:
is there a eslint rule to forbid inline styles?

limpbrains:
> is there a eslint rule to forbid inline styles?

[eslint-plugin-react-native](https://github.com/Intellicode/eslint-plugin-react-native/blob/master/docs/rules/no-inline-styles.md) has what we need. Do you want me to add it to this PR ?

Also looks like I've missed a few files in root like App and UnlockWith

Overtorment:
ok lets _not_ add extra dep

Overtorment:
@marcosrdz can I get your blessing?

PR #1126, opened 2020-05-24
REF: Make statusbar on android the wallet color

ncoelho:
ezgif com-video-to-gif (2)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/2bwrrajmpdhaa2cqa10w4tjjh8

[download apk](http://filestorage.bluewallet.io:1488/1590327631.apk)

marcosrdz:
Would the color change look smoother if u use DidFocus instead of willfocus?

ncoelho:
Haven’t tried it.

The right away is to use the “translucent” attribute. 

But it gets propagated to views that don’t need it. So I went just with the backgroundColor

PR #1159, opened 2020-05-30
REF: run eslint before any other tests

limpbrains:
This will make tests fail much faster if there are any problems with linter


Overtorment:
yeah, makes sense, thanks!
we also have `npm run lint:fix`  and `npm run lint:quickfix` to run locally

Issue #1160, opened 2020-05-30
Status bar colors logic lost after rn5


ncoelho:
If you wanna take a look on this. There should be two exceptions when the statusBar is light.
- tranjactions.js screen for the wallet view
- scanQRCode.js screen for camera

in android the bar background should just take the gradients color on the wallet view
`WalletGradient.headerColorFor(route.params.wallet.type)`

Overtorment:
attach a screenshot..?

ncoelho:
![image](https://user-images.githubusercontent.com/1238429/83326666-2ba07c80-a276-11ea-933c-6f02d38a5f84.png)

![image](https://user-images.githubusercontent.com/1238429/83326670-365b1180-a276-11ea-9448-c0990fded341.png)

limpbrains:
I've fixed transactions.js screen. But I don't see any logic to change status bar in ScanQRCode screen. Was it there or it is not implemented yet ?

limpbrains:
Like this?

![android](https://user-images.githubusercontent.com/155891/83349621-f2800f00-a33e-11ea-9f9b-c2734ff51a52.gif)
![ios](https://user-images.githubusercontent.com/155891/83349623-f9a71d00-a33e-11ea-8aae-79abc8320c0a.gif)

ncoelho:
yes, looks great that away.

limpbrains:
> yes, looks great that away. 
Added this to #1164

> Did you manage to reproduce the one on start screen, where it starts light?
No. But I think It could be fixed with my patch.
Probably, `StatusBar.setBarStyle` was fired too soon when you have already left transactions.js screen.
Can you describe steps how to reproduce it?

ncoelho:
On the emulator doesn't happen, but on a real device does. I think it is because the splash screen, where it starts light. So before it reaches the transactions.js, it has the list.js and the splash screen.

ncoelho:
Figured out how to reproduce this behaviour. On the emulator:
Settings > developer > dark appearance > Open the app

<img width="407" alt="Screenshot 2020-06-02 at 15 19 43" src="https://user-images.githubusercontent.com/1238429/83524865-7f180200-a4e4-11ea-9b3a-ae22b8edb9c5.png">

marcosrdz:
2a4c821d02ccf763a61c34604f1c19afb6c79628

PR #1164, opened 2020-05-31
FIX: listen focus events correctly on wallet transactions screen

limpbrains:
Fixing #1160
because componentDidMount is an async function, addListener was fired too soon.
I've fixed it by moving before awat

Also, I've added style change of StatusBar in ScanQRCode screen as described in
https://github.com/BlueWallet/BlueWallet/issues/1160#issuecomment-636449188


Overtorment:
can you pls fix conflicts?

Overtorment:
Thanks!

PR #1175, opened 2020-06-02
ADD: README iOS instructions updated

arizala13:
Added additional instruction on how to run iOS example of BlueWallet


Overtorment:
thanks!

PR #1181, opened 2020-06-03
FIX: round fiat currencies to 2 significant digits

limpbrains:
In my lightning wallet I have transactions with very small amounts, like 1 Satoshi.
When I'm switching wallet to show values in USD it shows me transaction amount of *$0.00*
I've changed `Currency.satoshiToLocalCurrency` function so: 

if value is more than 1, it renders 2  digits after decimal point.

1.123 -> 1.12

If value less than 1, it renders 2 significant digits after decimal point

0.123 -> 0.12

0.00000567 -> 0.0000057

More examples at [tests](https://github.com/BlueWallet/BlueWallet/commit/f120bc95fbcbef3985334c35f684e3da8c9e5e67#diff-6d0d8e8ba20fdcb5559fec2003df241a) 

Because current version uses `Intl` polyfill which doesn't supports modern `Intl.NumberFormat` API, I've removed it and used `Intl` wich comes with react-native. Starting 0.60 it comes with it

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -108,7 +108,7 @@ def enableProguardInReleaseBuilds = false
  * give correct results when using with locales other than en-US.  Note that
  * this variant is about 6MiB larger per architecture than default.
  */
-def jscFlavor = 'org.webkit:android-jsc:+'
+def jscFlavor = 'org.webkit:android-jsc-intl:+'
 
 /**
  * Whether to enable the Hermes VM.
--- a/tests/integration/Currency.test.js
+++ b/tests/integration/Currency.test.js
@@ -30,32 +30,4 @@ describe('currency', () => {
     cur = JSON.parse(await AsyncStorage.getItem(AppStorage.EXCHANGE_RATES));
     assert.ok(cur['BTC_EUR'] > 0);
   });
-
-  it('formats everything correctly', async () => {
-    const currency = require('../../currency');
-    await currency.setPrefferedCurrency(FiatUnit.USD);
-    await currency.startUpdater();
-    currency.exchangeRates.BTC_USD = 10000;
-    assert.strictEqual(currency.satoshiToLocalCurrency(1), '$0.0001');
-    assert.strictEqual(currency.satoshiToLocalCurrency(-1), '-$0.0001');
-    assert.strictEqual(currency.satoshiToLocalCurrency(123), '$0.012');
-    assert.strictEqual(currency.satoshiToLocalCurrency(146), '$0.015');
-    assert.strictEqual(currency.satoshiToLocalCurrency(123456789), '$12,345.68');
-
-    assert.strictEqual(currency.BTCToLocalCurrency(1), '$10,000.00');
-    assert.strictEqual(currency.BTCToLocalCurrency(-1), '-$10,000.00');
-    assert.strictEqual(currency.BTCToLocalCurrency(1.00000001), '$10,000.00');
-    assert.strictEqual(currency.BTCToLocalCurrency(1.0000123), '$10,000.12');
-    assert.strictEqual(currency.BTCToLocalCurrency(1.0000146), '$10,000.15');
-
-    assert.strictEqual(currency.satoshiToBTC(1), '0.00000001');
-    assert.strictEqual(currency.satoshiToBTC(-1), '-0.00000001');
-    assert.strictEqual(currency.satoshiToBTC(100000000), '1');
-    assert.strictEqual(currency.satoshiToBTC(123456789123456789), '1234567891.2345678');
-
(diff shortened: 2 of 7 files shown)


limpbrains:
Before


изображение

After

изображение

GladosBlueWallet:

 [android in browser] https://appetize.io/app/0gegp54tppdxhuqnhpzwvrzf68

[download apk](http://filestorage.bluewallet.io:1488/1591188700.apk)

Overtorment:
that doesnt feel right:

image
(thats -573 sat) 

should be rounded to`-0.06` imo.
you probably need a rule to round everything to 2 digits unless rounding would bring you exactly `0.00`

android/app/build.gradle
```diff
@@ -108,7 +108,7 @@ def enableProguardInReleaseBuilds = false
  * give correct results when using with locales other than en-US.  Note that
  * this variant is about 6MiB larger per architecture than default.
  */
-def jscFlavor = 'org.webkit:android-jsc:+'
+def jscFlavor = 'org.webkit:android-jsc-intl:+'
```
Overtorment, line 111: is that what allows removing `intl` dep? cool! but will it work in ios?

tests/integration/Currency.test.js
```diff
@@ -30,4 +30,32 @@ describe('currency', () => {
     cur = JSON.parse(await AsyncStorage.getItem(AppStorage.EXCHANGE_RATES));
     assert.ok(cur['BTC_EUR'] > 0);
   });
+
+  it('formats everything correctly', async () => {
+    const currency = require('../../currency');
+    await currency.setPrefferedCurrency(FiatUnit.USD);
+    await currency.startUpdater();
```
Overtorment, line 37: maybe there's a way to skip `startUpdater()` and move this test to `unit/`

Overtorment:
@ncoelho ^^^

android/app/build.gradle
limpbrains: Yes, it works. 
RN on iOS uses JavaScriptCore from the OS
https://github.com/react-native-community/jsc-android-buildscripts

ncoelho:
Not sure if showing $0.00 compared to $0.00058 is an improvement. I would rather show $0 as an improvement to $0.00

Thoughts?

tests/integration/Currency.test.js
limpbrains: > that doesnt feel right:
> 
> image
> (thats -573 sat)
> 
> should be rounded to`-0.06` imo.
> you probably need a rule to round everything to 2 digits unless rounding would bring you exactly `0.00`


I'm using 2 digits because there is significant difference between 0.01 and 0.014

limpbrains:
> you probably need a rule to round everything to 2 digits unless rounding would bring you exactly `0.00`

I'm using 2 digits because there is significant difference between 0.01 and 0.014

tests/integration/Currency.test.js
limpbrains: I will export currency.preferredFiatCurrency to mock in a test without AppStorage or Update process

limpbrains:
Showing $0 or $0.00 where it is actual transfer occurs feels wrong to me.
I would rather prefer @Overtorment variant 

0.014 -> 0.01

0.005123 -> 0.01

0.004123 -> 0.0041

_0.014 and 0.005123 represented as 0.01, but have they x2 actual difference_ 


Also, currently it is always 2 decimal digits is shown. $1 represented as $1.00. I can remove it, if you want

GladosBlueWallet:

 [android in browser] https://appetize.io/app/mz5fzzwrnacz8vd4xwjw7d6ajr

[download apk](http://filestorage.bluewallet.io:1488/1591198730.apk)

Overtorment:
ok, for transactions less than $1 it makes sense to me

limpbrains:
> ok, for transactions less than $1 it makes sense to me

Just to clarify, [second option](https://github.com/BlueWallet/BlueWallet/pull/1181#issuecomment-638276816), right ?

ncoelho:
> > ok, for transactions less than $1 it makes sense to me
> 
> Just to clarify, [second option](https://github.com/BlueWallet/BlueWallet/pull/1181#issuecomment-638276816), right ?

yep, sounds good to me.

Overtorment:
yep

GladosBlueWallet:

 [android in browser] https://appetize.io/app/7kndxfx5ha36fv73hhqyjq4u6w

[download apk](http://filestorage.bluewallet.io:1488/1591258403.apk)

limpbrains:
Ready to for merge

Overtorment:
thanks!

PR #1168, opened 2020-06-01
REF: upgrade eslint, prettier, fix errors, reformat everything 💄

limpbrains:
- prettier upgraded to v2. `"arrowParens": "avoid"` in config to reduce the diff 
- package.json simplified, added missing files to eslint
- eslint plugins and config files upgraded to the latest versions
- No more formatting conflicts between eslint and prettier
- `react-native/no-inline-styles` enabled everywhere, except BlueComponents
- `obj.hasOwnProperty(prop)` replaced with `prop in obj` to fix eslint error
- a few more minor eslint errors fixed 
- reviewed all `eslint-disable` rules, removed unused. 
- eslint --fix and prettier reformat everything

_PS
I've also tried to replace `require` and `module.exports` with `import` and `export` es6, but failed, too much to fix :)
Maybe next time_

PR #1185, opened 2020-06-03
REF: upgrade eslint, prettier, fix errors, reformat everything 💄

limpbrains:
Local rebased version of  #1168


GladosBlueWallet:

 [android in browser] https://appetize.io/app/ucrwkz99h1u9wauw7v62we7mmg

[download apk](http://filestorage.bluewallet.io:1488/1591217033.apk)

Overtorment:
bueno.

screen/lnd/scanLndInvoice.js
```diff
@@ -208,7 +208,7 @@ export default class ScanLndInvoice extends React.Component {
   };
 
   async pay() {
-    if (!this.state.hasOwnProperty('decoded')) {
+    if ('decoded' in this.state) {
```
Overtorment, line 211: found that `!` is lost here only after merge. will try to find more such cases and fix it in master, don't really want to unmerge

limpbrains: OMG sorry about that
I've walked through the patch and couldn't find more errors like this

PR #1191, opened 2020-06-05
FIX: following onchain+offchain invoice

limpbrains:
Originally Success screen was 2 levels deep and when `navigation.dangerouslyGetParent().pop()` was called in it, it didn't close the SelectWallet screen.

I've fixed it by dismissing SelectWallet after wallet is selected.
Also I've discovered that all screen that are needed for this workflow to works are already exists inside RootStack. So I've removed HandleOffchainAndOnChain completely to simplify the code.

Less code - better!

### DIFF
--- a/class/deeplink-schema-match.js
+++ b/class/deeplink-schema-match.js
@@ -68,12 +68,11 @@ class DeeplinkSchemaMatch {
     }
     if (isBothBitcoinAndLightning) {
       completionHandler([
-        'HandleOffchainAndOnChain',
+        'SelectWallet',
         {
-          screen: 'SelectWallet',
-          params: {
-            onWalletSelect: wallet =>
-              completionHandler(DeeplinkSchemaMatch.isBothBitcoinAndLightningOnWalletSelect(wallet, isBothBitcoinAndLightning)),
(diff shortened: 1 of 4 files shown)


class/deeplink-schema-match.js
```diff
@@ -68,12 +68,11 @@ class DeeplinkSchemaMatch {
...
     if (isBothBitcoinAndLightning) {
       completionHandler([
-        'HandleOffchainAndOnChain',
+        'SelectWallet',
         {
-          screen: 'SelectWallet',
-          params: {
-            onWalletSelect: wallet =>
-              completionHandler(DeeplinkSchemaMatch.isBothBitcoinAndLightningOnWalletSelect(wallet, isBothBitcoinAndLightning)),
+          onWalletSelect: (wallet, { navigation }) => {
```
limpbrains, line 73: I think it is better to inject navigation here this way

GladosBlueWallet:

 [android in browser] https://appetize.io/app/kvw9856mfqdeg5mcdh2bjz2p7c

[download apk](http://filestorage.bluewallet.io:1488/1591364165.apk)

PR #1192, opened 2020-06-05
ADD: extend pull to close gesture area to full modal height

limpbrains:
On the iOS you can close modal by pull down gesture.
It already works in BlueWallet on scanQRCore, import and other modals, but only if you grab them by Header. 
I'm extending grab area to the full height of the modals.
Unfortunately it doesn't work everywhere correctly. That's why I'm disabling gestures for reorderWallets and selectWallet screens.

On android gestures are disabled by default, so it is not affected. 
I'm only changing gestureResponseDistance.vertical, horizontal remains default - 50

It feels very cool on import or scanqrcode screens

### DIFF
--- a/Navigation.js
+++ b/Navigation.js
@@ -1,6 +1,7 @@
 // import { createAppContainer } from '@react-navigation/native';
 import React from 'react';
 import { createStackNavigator, TransitionPresets } from '@react-navigation/stack';
+import { Platform, Dimensions } from 'react-native';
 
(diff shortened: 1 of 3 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/40k1a05b4crx70h7jhmfrczqeg

[download apk](http://filestorage.bluewallet.io:1488/1591366677.apk)

Overtorment:
thanks!

PR #1276, opened 2020-06-22
REF: disable some warnings

Overtorment:


marcosrdz:
We shouldn’t ignore this warning. This means we gotta update the parameters we pass with nav 5.

Overtorment:
which one of 3?

GladosBlueWallet:

 [android in browser] https://appetize.io/app/vx46nx7qnzd7n1uu8hq9nr1dxg

[download apk](http://filestorage.bluewallet.io:1488/1592851161.apk)

Overtorment:
also, this means those warnings are not displayed in yellowbox, but still visible in console output

GladosBlueWallet:

 [android in browser] https://appetize.io/app/fhfpkh0qz8j29u9y6uvn9u1p04

[download apk](http://filestorage.bluewallet.io:1488/1593455625.apk)

Overtorment:
@limpbrains do you think its okay to mute those warnings? as far as I understood, they are hidden from yellowbox only, should be still present in console output

limpbrains:
I'm ok with disabling `Non-serializable values were` and `Can't perform a React state update`.

I don't know about `Require cycle`, if it is safe to turn it of, I'm fine

`{"code":404` looks like it can be fixed

PR #1294, opened 2020-06-25
FIX: local trader: when user is not logged in, clicking 'accept offer…

### LINKED ISSUE #1280: HodlHodl improvements
* [x] when user is not logged in, clicking 'accept offer' should not just throw alert, but open login screen
* [x] before accepting offer, refetch it and use fresh id and version, so there are no 'outdated' errors
* [x] when 'I am selling', sort order should be reversed

Overtorment:
…' should not just throw alert, but open login screen

FIX: local trader: before accepting offer, refetch it and use fresh id and version so there are no 'outdated' errors
FIX: local trader: when 'I am selling', sort order should be reversed

(closes #1280)

### DIFF
--- a/screen/wallets/hodlHodl.js
+++ b/screen/wallets/hodlHodl.js
@@ -109,7 +109,10 @@ export default class HodlHodl extends Component {
 
     const sort = {
       [HodlHodlApi.SORT_BY]: HodlHodlApi.SORT_BY_VALUE_PRICE,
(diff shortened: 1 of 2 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/1khzmpe4r64br65w1wdqqdje7r

[download apk](http://filestorage.bluewallet.io:1488/1593103729.apk)

Issue #1217, opened 2020-06-10
speed up e2e tests


Overtorment:
travis tests are crazy slow. typically 40 min, of which 30 min is setting up the environment with java and compiling production build. also, some tests randomly fail for an unknown reason, and it looks like app just crashes on CI.
it was hard to setup travis, and integration feels hacky as travis was not made to handle android emus.

we need to try out alternatives:

* github actions https://medium.com/@reime005/the-best-ci-cd-for-react-native-with-e2e-support-4860b4aaab29
* bitrise https://github.com/wix/Detox/blob/master/docs/Guide.RunningOnCI.md
* buddy https://buddy.works/docs/quickstart/react-native

Overtorment:
target: 15min

limpbrains:
I will try to port e2e tests on github actions

PR #1298, opened 2020-06-26
OPS: github-actions for e2e and jest tests

limpbrains:
After endless amount of attempts this is the best option to run e2e tests on github-actions. **30** min. Not much of a win. But config much simpler than travis. It caches `~/.npm` and `~/.gradle/caches`

If test is failing, recorded video is saved as artifacts.

I've also moved circleci tests to github actions because why not?

All you need to do is to create Secrets for this repository.

Also, I've renamed scripts in package.json because their names were confusing.

What I've tried and it didn't work:
- running tests against android debug build. It just freezes for unknown reason
- running tests against iOS sim build. It is possible in theory, but right now `npx detox build -c ios.sim.release` is broken and I can't fix it
- caching node_modues - breaks detox

You can see some examples here:
https://github.com/limpbrains/BlueWallet/actions

For #1217

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,103 @@
+name: Tests
+
+# https://dev.to/edvinasbartkus/running-react-native-detox-tests-for-ios-and-android-on-github-actions-2ekn
+# https://medium.com/@reime005/the-best-ci-cd-for-react-native-with-e2e-support-4860b4aaab29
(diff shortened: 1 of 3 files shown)


Overtorment:
image

Overtorment:
merged in https://github.com/BlueWallet/BlueWallet/pull/1301

PR #1318, opened 2020-07-02
ADD: proxy fields in electrum settings (DO NOT MERGE YET)

chrisc93:
Needs https://github.com/BlueWallet/rn-electrum-client/pull/4 before merging

Works well for Android devices running Orbot. I didn't get around to writing anything related to failed connections. Will need to alert the user if a socks server isn't detected.

**Note** 
I wasn't too happy that I had to go back to the older react-native-tcp framework, but the new react-native-tcp-socket isn't mature enough for some of the API calls in the socks library used in the rn-electrum-client merge request.

Also, for some reason `npm install` never worked for me. It wouldn't pull down the code for the `socks` dependency. I got around this by using `yarn install`


package.json
```diff
@@ -139,7 +139,7 @@
     "react-native-snap-carousel": "3.9.1",
     "react-native-sortable-list": "0.0.24",
     "react-native-svg": "12.1.0",
-    "react-native-tcp-socket": "3.7.1",
+    "react-native-tcp": "git+https://github.com/BlueWallet/react-native-tcp.git",
```
Overtorment, line 142: this library is abandoned by maintainer, and it crashes on android a lot, that's why `react-native-tcp-socket` is a way to go. its maintainer is active and friendly and Im sure will accept PRs

chrisc93: Interesting, I haven't noticed any crashes on Android, but still good to know. I can try to rework this with the new library.

Overtorment:
ok I feel we will sooner add native TOR support, which is better than ugly messign with proxies

Overtorment:
hi @chrisc93! this is a perfect time to pick this up. long story short, https://github.com/BlueWallet/BlueWallet/pull/2295 adds complete TOR daemon into bluewallet. HTTP calls work perfectly, but there is no socket support. BUT, upon startup daemon returns socks proxy port it listens on, so if we could teach BW to work with socks proxy we would solve TOR electrum connection. do you want to work on this? there is also some  bounty on  (https://github.com/BlueWallet/BlueWallet/pull/2295)

chrisc93:
Hey @Overtorment I'm swamped at the moment. But it looks like some good progress has been made #2295 so this can probably be closed now

Cheers!

PR #1326, opened 2020-07-04
Add support for zpub with fingerprint

stepansnigirev:
Bitcoin Core uses an HD key format with maser fingerprint and derivation path in square brackets:
`[fingerprint/derivation]xpub`

In [Specter](https://github.com/cryptoadvance/specter-diy) we use a similar way to display xpubs. This PR adds support for this format of xpub and fills the master fingerprint.

### DIFF
--- a/tests/unit/watch-only-wallet.test.js
+++ b/tests/unit/watch-only-wallet.test.js
@@ -120,7 +120,8 @@ describe('Watch only wallet', () => {
   });
 
   it('can import zpub with master fingerprint', async () => {
-    const zpub = "[8cce63f8/84h/0h/0h]zpub6s2RJ9qAEBW8Abhojs6LyDzF7gttcDr6EsR3Umu2aptZBb45e734rGtt4KqsCMmNyR1EEzUU2ugdVYez2VywQvAbBjUSKn8ho4Zk2c5otkk";
+    const zpub =
+      '[8cce63f8/84h/0h/0h]zpub6s2RJ9qAEBW8Abhojs6LyDzF7gttcDr6EsR3Umu2aptZBb45e734rGtt4KqsCMmNyR1EEzUU2ugdVYez2VywQvAbBjUSKn8ho4Zk2c5otkk';
     const w = new WatchOnlyWallet();
     w.setSecret(zpub);
     w.init();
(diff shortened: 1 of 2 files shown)


stepansnigirev:
Done.

tests/unit/watch-only-wallet.test.js
```diff
@@ -119,6 +119,21 @@ describe('Watch only wallet', () => {
     assert.strictEqual(w.getLabel(), 'Cobo Vault 5271c071');
   });
 
+  it('can import zpub with master fingerprint', async () => {
+    const zpub =
+      '[8cce63f8/84h/0h/0h]zpub6s2RJ9qAEBW8Abhojs6LyDzF7gttcDr6EsR3Umu2aptZBb45e734rGtt4KqsCMmNyR1EEzUU2ugdVYez2VywQvAbBjUSKn8ho4Zk2c5otkk';
```
Overtorment, line 124: looking at https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md I dont see this specific example..?

Overtorment:
thanks! will be out in next release

tests/unit/watch-only-wallet.test.js
stepansnigirev: This PR adds only partial support of the Core's key expressions - parsing master fingerprint from square brackets.

Bitcoin Core doesn't use zpubs, so for full support of Core's `KEY` expressions we would need to:
- parse derivation path, check that it starts with `8-byte-fingerprint/84h`
- set derivation path to what was provided in brackets
- parse xpub, convert it to zpub and set as wallet secret

I am not sure I will be capable of doing all that, so I stopped at square bracket import.

Overtorment: do you have test vectors for those? I could do that.

btw we already have code that converts xpub to zpub.

stepansnigirev: These two are basically the same xpubs just with different prefixes:
- `[8cce63f8/84h/0h/0h]xpub6DMtgpVKvpRAU1Ka59X6Z3oEmkbziys6QeNbuz7Fpp8o5PRd8nhwc9ac1uvhCYTYA8mck3HM7ayXiyRrb79upSoPT45b9xVjFcSTFQ2n9ji`
- `[8cce63f8/84h/0h/0h]zpub6s2RJ9qAEBW8Abhojs6LyDzF7gttcDr6EsR3Umu2aptZBb45e734rGtt4KqsCMmNyR1EEzUU2ugdVYez2VywQvAbBjUSKn8ho4Zk2c5otkk`

PR #1329, opened 2020-07-06
manual rebase of changes from PR #772

fiatjaf:
I took the changes from https://github.com/BlueWallet/BlueWallet/pull/772 and applied them to current `master`.

PR #1338, opened 2020-07-08
Lnurlpay2

limpbrains:
Local version of #1329 by @fiatjaf


Overtorment:
lnurl-pay support merged in https://github.com/BlueWallet/BlueWallet/pull/1374

PR #1345, opened 2020-07-15
FIX: When doing max amount, some scenarios would return invalid amoun…

marcosrdz:
…t due to missing value change.


GladosBlueWallet:

 [android in browser] https://appetize.io/app/1w24fyjejjqwk016yu3b87x6ac

[download apk](http://filestorage.bluewallet.io:1488/1594779496.apk)

PR #1392, opened 2020-07-28
TST: fix some tests failing due to: 1) timers 2) linux locale differe…

Overtorment:
…nces

### DIFF
--- a/tests/unit/loc.test.js
+++ b/tests/unit/loc.test.js
@@ -1,9 +1,10 @@
-/* global it, describe */
+/* global it, describe, jest */
 import { BitcoinUnit } from '../../models/bitcoinUnits';
 import { FiatUnit } from '../../models/fiatUnit';
 import { _leaveNumbersAndDots, formatBalanceWithoutSuffix, formatBalancePlain, formatBalance } from '../../loc';
 const assert = require('assert');
 const currency = require('../../blue_modules/currency');
+jest.useFakeTimers();
 
 describe('Localization', () => {
   it('internal formatter', () => {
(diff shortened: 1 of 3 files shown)


tests/unit/loc.test.js
```diff
@@ -16,17 +17,17 @@ describe('Localization', () => {
     currency._setExchangeRate('BTC_RUB', 660180.143);
     currency._setPreferredFiatCurrency(FiatUnit.RUB);
     let newInputValue = formatBalanceWithoutSuffix(152, BitcoinUnit.LOCAL_CURRENCY, false);
-    assert.strictEqual(newInputValue, 'RUB 1.00');
+    assert.ok(newInputValue === 'RUB 1.00' || newInputValue === '1,00 ₽', 'Unexpected: ' + newInputValue);
```
Overtorment, line 20: @limpbrains I updated nodejs to 13 on my localhost, and noticed that this fails

GladosBlueWallet:

 [android in browser] https://appetize.io/app/45t8xfzjrrd3yxnm7fkdh5a2h4

[download apk](http://filestorage.bluewallet.io:1488/1595952350.apk)

tests/unit/loc.test.js
limpbrains: `[].includes()` is better when number of items 3 and more. When 2 `||` is readable.
I would say lets skip node 13, it is already outdated https://nodejs.org/en/about/releases/
Node 14 will be next LTS, so if other tests and build is working on it, lest switch and remove OR in tests 

I think it's not because of locale settings, but because of `Intl` upgrade

PR #1518, opened 2020-08-25
FIX: recieve button for watch-only wallets

limpbrains:
When wallet is watch-only we need to show one button - Receive and it should take all the space in floatButtons container
Before:

image

After

image

### DIFF
--- a/BlueComponents.js
+++ b/BlueComponents.js
@@ -1374,7 +1374,7 @@ const sendReceiveScanButtonFontSize =
 export class BlueReceiveButtonIcon extends Component {
   render() {
     return (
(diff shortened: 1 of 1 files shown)


Overtorment (approved):

ncoelho:
There was a case for the "send" to activate hardware wallet in watch-only mode

GladosBlueWallet:

 [android in browser] https://appetize.io/app/ywczaw97pffua7aemdx3rfeqng

[download apk](http://filestorage.bluewallet.io:1488/1598357024.apk)

limpbrains:
> There was a case for the "send" to activate hardware wallet in watch-only mode

Yes, when it's zpub. And I've checked - working fine

PR #1545, opened 2020-08-27
ref-storage-in-realm-db

Overtorment:
I researched on all available databases for RN and chose realm for now, because its not sql-based (its document oriented) and built in encryption

so I rewrote storing data, and heavy data is now offloaded to realm, so we are not encrypting megabytes of data in JS process, it is offloaded to native realm process where it is encrypted and then flushed to disk
so, wallet with 1k txs taking 10MB now works very well, our saveToDisk() procedure takes 1sec max
we call saveToDisk every time some data changed, like, if we fetched new txs from electrum
bad news: realm here is used in a goofy way. because realm requires strict schema definition, and we have very complex internal representation of transactions (tree-like) and it differs from wallet type to wallet type
so heavy transaction data is serialized and placed in realm, and then read back during cold boot, like some key-value nosql. so all transactions are stored in RAM during runtime (as before). cool realm features like lazy load are not used. OTOT, there is no faster storage than RAM in runtime unless you need querying (which we dont)
so Id say we are good untill we enounter some beast who has 10k+ transactions in a wallet. then we will have to re-evaluate, and who knows, maybe it wil still perform decent
so Im gona test this build for a while to see if there are issues
note: basic wallet data & secrets are still stored in keystore (which will survive if ios app deleted/reinstalled), but realm keeps its stuff in files in app documents dir. so if anything happens - wallets will survive, they will just have empty transactions, which is solved with plain pull-to-refresh

### DIFF
--- a/blue_modules/BlueElectrum.js
+++ b/blue_modules/BlueElectrum.js
@@ -222,7 +222,7 @@ module.exports.getTransactionsFullByAddress = async function (address) {
  * @returns {Promise<{balance: number, unconfirmed_balance: number, addresses: object}>}
  */
 module.exports.multiGetBalanceByAddress = async function (addresses, batchsize) {
-  batchsize = batchsize || 100;
+  batchsize = batchsize || 200;
   if (!mainClient) throw new Error('Electrum client is not connected');
   const ret = { balance: 0, unconfirmed_balance: 0, addresses: {} };
 
@@ -251,6 +251,7 @@ module.exports.multiGetBalanceByAddress = async function (addresses, batchsize) 
     }
 
     for (const bal of balances) {
+      if (bal.error) console.warn('multiGetBalanceByAddress():', bal.error);
       ret.balance += +bal.result.confirmed;
       ret.unconfirmed_balance += +bal.result.unconfirmed;
       ret.addresses[scripthash2addr[bal.param]] = bal.result;
--- a/class/app-storage.js
+++ b/class/app-storage.js
@@ -17,7 +17,11 @@ import {
 } from './';
 import WatchConnectivity from '../WatchConnectivity';
 import DeviceQuickActions from './quick-actions';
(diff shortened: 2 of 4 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/j7p6qh60yy19m2344g53zjg89g

[download apk](http://filestorage.bluewallet.io:1488/1598523078.apk)

GladosBlueWallet:

 [android in browser] https://appetize.io/app/a6421pr4bm544zg90hxaw7w094

[download apk](http://filestorage.bluewallet.io:1488/1598538326.apk)

GladosBlueWallet:

 [android in browser] https://appetize.io/app/pp0dendt1er8ttmnxr4r9t0kk0

[download apk](http://filestorage.bluewallet.io:1488/1598615842.apk)

limpbrains (approved):
Look like you are open/close realm db each time you call `saveToDisk` 
Would it be faster to open realm db on app start and close on app close ?

Overtorment:
> Look like you are open/close realm db each time you call `saveToDisk`
> Would it be faster to open realm db on app start and close on app close ?

docs say `Remember to use the close() method on opened realms when you are done with them.` and `Remember to close the realm when finished`.

how do you even catch app close if app can be killed by iOS when its out of memory, or user can kill the app manually

marcosrdz:
> > Look like you are open/close realm db each time you call `saveToDisk`
> 
> > Would it be faster to open realm db on app start and close on app close ?
> 
> 
> 
> docs say `Remember to use the close() method on opened realms when you are done with them.` and `Remember to close the realm when finished`.
> 
> 
> 
> how do you even catch app close if app can be killed by iOS when its out of memory, or user can kill the app manually

This would require Native code

limpbrains:
Ok!

PR #1609, opened 2020-09-01
Fix unconfirmed balance

tanx:
This should include the unconfirmed balance get calling getBalance()

### DIFF
--- a/class/wallets/abstract-hd-electrum-wallet.js
+++ b/class/wallets/abstract-hd-electrum-wallet.js
@@ -41,7 +41,7 @@ export class AbstractHDElectrumWallet extends AbstractHDWallet {
     for (const bal of Object.values(this._balances_by_internal_index)) {
       ret += bal.c;
     }
-    return ret + (this.getUnconfirmedBalance() < 0 ? this.getUnconfirmedBalance() : 0);
+    return ret + (this.getUnconfirmedBalance() > 0 ? this.getUnconfirmedBalance() : 0);
   }
 
   /**


class/wallets/abstract-hd-electrum-wallet.js
```diff
@@ -41,7 +41,7 @@ export class AbstractHDElectrumWallet extends AbstractHDWallet {
     for (const bal of Object.values(this._balances_by_internal_index)) {
       ret += bal.c;
     }
-    return ret + (this.getUnconfirmedBalance() < 0 ? this.getUnconfirmedBalance() : 0);
+    return ret + (this.getUnconfirmedBalance() > 0 ? this.getUnconfirmedBalance() : 0);
```
tanx, line 44: Ah wait. I see what you're were doing here. Am I correct that this is meant to include just the unconfirmed balance of outgoing transactions (negative values), but not those of unconfirmed incoming transactions (positive values)?

tanx: Why not just call the following?
```suggestion
    return ret + this.getUnconfirmedBalance();
```

Overtorment: yeah the way it works right now is that when you spend from your wallet and its not yet confirmed - total balance is reduced immediately (that's what users expect from their experience with traditional banking apps). 
and when there is an incoming transaction - its not added to total balance until it is confirmed.
no one ever complained.

PR #1622, opened 2020-09-03
FIX: Improve wallet description translation

eliaspoveda:
Changed
"A wallet represents a pair of a secret (private key) and an addressyou can share to receive coins"
to
"A wallet represents a pair of keys, one private and one you can share to receive coins"


ncoelho (approved):
Thanks!

Issue #1617, opened 2020-09-03
Android -> decrypt storage works but throws an error 


Overtorment:
```
 [TypeError: _reactNativeSecureKeyStore.default.setResetOnAppUninstallTo is not a function. (In '_reactNativeSecureKeyStore.default.setResetOnAppUninstallTo(value)', '_reactNativeSecureKeyStore.default.setResetOnAppUninstallTo' is undefined)]
```

![image](https://user-images.githubusercontent.com/1913337/92118917-fd8b8500-edee-11ea-96a2-f7cd90cafaeb.png)

marcosrdz:
https://github.com/BlueWallet/BlueWallet/compare/Android---decrypt-storage-works-but-throws-an-error-%231617-?expand=1

marcosrdz:
feature is iOS only.

PR #1629, opened 2020-09-04
FIX:  Android -> decrypt storage works but throws an error #1617

marcosrdz:


GladosBlueWallet:

 [android in browser] https://appetize.io/app/rzmdhuzmefe0gpq6c5pdfmevpm

[download apk](http://filestorage.bluewallet.io:1488/1599188162.apk)

class/app-storage.js
```diff
@@ -77,11 +78,13 @@ export class AppStorage {
   }
 
   async setResetOnAppUninstallTo(value) {
-    await this.setItem(AppStorage.DELETE_WALLET_AFTER_UNINSTALL, value ? '1' : '');
-    try {
-      RNSecureKeyStore.setResetOnAppUninstallTo(value);
-    } catch (Error) {
-      console.warn(Error);
+    if (Platform.OS === 'ios') {
```
Overtorment, line 81: id probably just write 'if (Platform.OS !== 'ios') return;' and the diff would be a oneliner, but ok

PR #1635, opened 2020-09-04
Implement list of custom electrum servers

misovan:
This PR implements the following features:

- add or remove custome electrum server to the new list managed in Netwrok > Electrum server settings
- select server from the list to be connected to (after restart)
- remove server from the list
- delete all custom servers and return to default bluewallet servers

Rational is that users might want to connect to their electrum server while using wallet in their private network but be able to quickly switch to another server when on public network (might still be their own but exposed e.g. through ssh tunnel to their device or public one they trust)


marcosrdz:
Im getting a looped Error in connect() function when I click Save
Simulator Screen Shot - iPhone 11 - 2020-09-04 at 11 56 01

misovan:
Thanks fo giving a try. I will try to simulate it on fresh install (did not happen to me and I believe I did not touch connect function)

misovan:
OK, I was able to simulate the error that occured when incorrect host or port were submitted. In such case, BlueElectrum.testConnection was periodically trying to reconnect to wrong address. I fixed the bug in the function and on top I slightly revamped the original form to add new electrum server in electrum settings - it now allows selecting either SSL (TLS) or unencrypted port and it got some basic validations.

Screenshot_20200905-001335_BlueWallet

marcosrdz:
@misovan do you mind granting me write access to your branch so I can make UI modifications?

misovan:
Hi @marcosrdz may I kindly ask you for quick guidance regarding remaining DeepScan check? It reports multiple issues in files I did not change in this PR and then one in my code I am not sure how to efficiently address: 
https://deepscan.io/dashboard/#view=project&pid=6979&iid=39254385&tid=5209&subview=pull-request&prid=471778

Thank you upfront for your help.

marcosrdz:
> Hi @marcosrdz may I kindly ask you for quick guidance regarding remaining DeepScan check? It reports multiple issues in files I did not change in this PR and then one in my code I am not sure how to efficiently address: 
> 
> https://deepscan.io/dashboard/#view=project&pid=6979&iid=39254385&tid=5209&subview=pull-request&prid=471778
> 
> 
> 
> Thank you upfront for your help.

Hello! I'll be working on updating this branch really soon. I'll address the deepscan result.

Overtorment:
fyi sometimes deepscan is barking at the wrong tree and is safe to ignore.
Ill be review this later.

blue_modules/BlueElectrum.js
```diff
@@ -525,9 +541,10 @@ module.exports.testConnection = async function (host, tcpPort, sslPort) {
     await client.connect();
     await client.server_version('2.7.11', '1.4');
     await client.server_ping();
-    client.close();
+    await client.close();
```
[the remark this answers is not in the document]
Overtorment: anyway I think this part was refactored massively

screen/settings/electrumSettings.js
```diff
@@ -164,54 +347,79 @@ export default class ElectrumSettings extends Component {
...
-                onChangeText={text => this.setState({ host: text.trim() })}
-                numberOfLines={1}
-                style={styles.inputText}
-                editable={!this.state.isLoading}
-                placeholderTextColor="#81868e"
-                autoCorrect={false}
-                underlineColorAndroid="transparent"
-              />
+            <View style={styles.addServerTitle}>
+              <BlueText style={styles.settingText}>Add custom server</BlueText>
```
Overtorment, line 366: add to localization, and other strings

Overtorment (changes requested):
whole concept feels to be overenginered.
it can be greately simplified if you treat electrums list more like a history list:
that is, logic with `AppStorage.ELECTRUM_HOST` stays and doesnt get replaced with `AppStorage.ELECTRUM_SERVER_LIST`. every time user inputs a working unique electrum - it gets pushed to history list, which should be rendered as a list the way it is rendered now. 
"Add custom server" & "Use SSL" are unnecessary

PS. sorry it took so long to review. Lets finish it up and move to more testing

misovan:
OK, thanks for review. I'll try to get back to that in few days. 

1. I'll remove "selected" property from the list and adapt related logic to use previous ELECTRUM_HOST, etc constants from local storage. 
2. I'll revamp the UI so the list is presented as a history. 
3. I'll think of whether to keep "Remove" option alongside server item or perhaps replace it by "Remove all" that would clean the UI from abundance of buttons.
4. After this new concept gets ack, I'll deal with localization and other necessary cleaning

misovan:
Ok, I have updated the pullrequest so it now implements more simple electrum server list history with 3 features: 
- new server is added to list on save
- server from list can be selected and becomes active server (after restart)
- history can be cleaned

Open topics:
- Translation and some small styling after ack the concept by reviewer
- Open whether to keep existing feature to get back to default servers through submitting empty form or make it rather special button

marcosrdz:
DeepScan error

marcosrdz:
I think Clear All should have a confirmation alert because you might accidentally press it and not have the sever information readily available to reinsert.

marcosrdz:
Perhaps a "Reset to default" button to have it connect back at a random server?

marcosrdz (changes requested):
Deep scan error

misovan:
> I think Clear All should have a confirmation alert because you might accidentally press it and not have the sever information readily available to reinsert.

I'll do that.

> Perhaps a "Reset to default" button to have it connect back at a random server?

I'll do that. I'll then refactor save method and add basic validation to form fields.

class/app-storage.js
```diff
@@ -677,4 +678,4 @@ export class AppStorage {
   sleep = ms => {
     return new Promise(resolve => setTimeout(resolve, ms));
   };
-}
+}
```
Overtorment, line 681: unnecessary change

.gitignore
```diff
@@ -65,6 +65,6 @@ artifacts/
 
 # Editors
 .vscode/
-*.mx
-*.realm
-*.realm.lock
\ No newline at end of file
+/.editorconfig
```
Overtorment, line 68: looks like a result of unfortunate conflict resolve. 
lines

```
*.mx
*.realm
*.realm.lock 
```

got lost

loc/en.json
```diff
@@ -223,78 +223,85 @@
...
-    "privacy_quickactions": "Wallet Shortcuts",
-    "privacy_quickactions_explanation": "Touch and hold the BlueWallet app icon on your Home Screen to quickly view your wallet's balance.",
-    "privacy_clipboard_explanation": "Provide shortcuts if an address, or invoice, is found in your clipboard.",
-    "push_notifications": "Push Notifications",
-    "retype_password": "Re-type password",
-    "save": "Save",
-    "saved": "Saved",
-    "success_transaction_broadcasted" : "Success! You transaction has been broadcasted!"
-  },
+    "settings": {
```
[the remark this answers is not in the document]
Overtorment: only changed (or added) lines should be marked. in this case I sense `\n\r` issue, but not sure

screen/settings/electrumSettings.js
```diff
@@ -66,42 +65,127 @@ export default class ElectrumSettings extends Component {
     });
   };
 
+  selectServer = async server => {
+    const { host, port, sslPort } = server;
```
Overtorment, line 69: I think you can throw away body of that function and replace it with

```
this.setState({ host: server.host, port: server.port, sslPort: server.sslPort }, () => {
  this.save();
});

```

screen/settings/electrumSettings.js
```diff
@@ -66,42 +65,127 @@ export default class ElectrumSettings extends Component {
...
+      { text: 'Cancel', onPress: () => console.log('Cancel Pressed'), style: 'cancel' },
+      { text: 'OK', onPress: () => this.clearHistory() },
+    ]);
+  }
+
+  clearHistory = async () => {
+    this.setState({ isLoading: true }, async () => {
+      await AsyncStorage.setItem(AppStorage.ELECTRUM_SERVER_HISTORY, JSON.stringify([]));
+      try {
+        await DefaultPreference.setName('group.io.bluewallet.bluewallet');
```
Overtorment, line 114: you dont need to send ELECTRUM_SERVER_HISTORY to DefaultPreference.
Its only needed to let IOS Widget know which electrum we are using so widget can fetch fee estimations etc

screen/settings/electrumSettings.js
```diff
@@ -66,42 +65,127 @@ export default class ElectrumSettings extends Component {
...
+
+          if (!this.serverExists({ host, port, sslPort })) {
+            serverHistory.push({
+              host,
+              port,
+              sslPort,
+            });
+
+            await AsyncStorage.setItem(AppStorage.ELECTRUM_SERVER_HISTORY, JSON.stringify(serverHistory));
+            await DefaultPreference.set(AppStorage.ELECTRUM_SERVER_HISTORY, JSON.stringify(serverHistory));
```
Overtorment, line 176: dont send ELECTRUM_SERVER_HISTORY to ios widget

screen/settings/electrumSettings.js
```diff
@@ -288,4 +400,20 @@ const styles = StyleSheet.create({
...
+    paddingVertical: 10,
+  },
+  serverHistoryItem: {
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+    paddingVertical: 20,
+    borderBottomColor: BlueCurrentTheme.colors.formBorder,
+    borderBottomWidth: 1,
+  },
 });
```
Overtorment, line 419: file `scripts/maccatalystpatches/applypatchesformaccatalyst.sh` should not be changed

Overtorment (changes requested):
I like what I see, probably another round of fixes and we are close to merge!

Overtorment:
and please do a git rebase so we are not carrying history of 21 commits here

misovan:
- I incorporated the comments
- In case of multiple async-storage operations, they are now enclosed into single promise
- I made my best to rebase and simplify commit history to 3 commits (hopefully that one is ok)

Overtorment:
I reworked it into https://github.com/BlueWallet/BlueWallet/pull/2423

misovan:
> I reworked it into #2423

Thanks, I had cleaned issues with code marked as changed perhaps due to LF/CRLF in translation file however after pushing it, github showed it again as changed... may be something related to my Visual Studio on windows setup, I'll investigate.  To clean up the mess I as well planned  to move the code to clean branch, but did not make it during last couple of days.

I am sorry for unwanted additional work and I wish you & all the team peaceful Christmas.

Overtorment:
Happy Xmas!

PR #1763, opened 2020-09-13
FIX: create interval for tx list in useEffect hook

limpbrains:
- each render of `WalletTransactions` creates new interval
- in `setInterval` the state is changing which triggers render
- after a few cycles we have hundreds re-renders each of them triggers `wallet.getTransactions`
- app freezes

I've moved interval creation to `useEffect`, now it will be executed only once on component mount.

### DIFF
--- a/screen/wallets/transactions.js
+++ b/screen/wallets/transactions.js
@@ -213,7 +213,6 @@ const WalletTransactions = () => {
       backgroundColor: colors.background,
     },
   });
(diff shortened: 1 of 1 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/ek0hf4anttmu6h9yedcva4z550

[download apk](http://filestorage.bluewallet.io:1488/1599997513.apk)

marcosrdz (approved):

PR #1888, opened 2020-09-21
FIX: change TimeElapsed on transactions refresh to force items render

limpbrains:
Because BlueTransactionListItem is a pure component we need to change timeElapsed to force it render

### DIFF
--- a/screen/wallets/transactions.js
+++ b/screen/wallets/transactions.js
@@ -310,6 +310,7 @@ const WalletTransactions = () => {
       noErr = false;
       alert(err.message);
       setIsLoading(false);
+      setTimeElapsed(prev => prev + 1);
     }
     if (noErr && smthChanged) {
       console.log('saving to disk');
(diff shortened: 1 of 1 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/e7ftrbybr4gn7qv01eekgpv8b8

[download apk](http://filestorage.bluewallet.io:1488/1600686757.apk)

screen/wallets/transactions.js
```diff
@@ -310,6 +310,7 @@ const WalletTransactions = () => {
       noErr = false;
       alert(err.message);
       setIsLoading(false);
+      setTimeElapsed(prev => prev + 1);
```
Overtorment, line 313: will it cause performance issues?

limpbrains: I don't think so. Using pure components in FlatList is important because when you fast scroll it, it renders a lot of items.
Here we re-rendering not during the scroll, so should be fast

Overtorment (approved):

marcosrdz (approved):

PR #1892, opened 2020-09-22
TST: unit test for payjoin

Overtorment:


Overtorment:
@lukechilds @Kukks @junderw can you pls take a look?

PR #1926, opened 2020-09-25
OPS: bump payjoin dep

Overtorment:

PR #1937, opened 2020-09-29
Ops bump payjoin dep

Kukks:
Supersedes #1926 in that the tests now succeed, passes the payment script to the payjoin client and also checks if BIP21 signals that the server has output substitution disabled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "lottie-react-native": "3.5.0",
     "metro-react-native-babel-preset": "0.63.0",
     "path-browserify": "1.0.1",
-    "payjoin-client": "git+https://github.com/bitcoinjs/payjoin-client.git#bip78",
+    "payjoin-client": "^1.0.0-beta",
     "pbkdf2": "3.1.1",
     "prettier": "2.1.1",
     "process": "0.11.10",
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "lottie-react-native": "3.5.0",
     "metro-react-native-babel-preset": "0.63.0",
     "path-browserify": "1.0.1",
-    "payjoin-client": "^1.0.0-beta",
+    "payjoin-client": "git+https://github.com/bitcoinjs/payjoin-client.git#bip78",
     "pbkdf2": "3.1.1",
     "prettier": "2.1.1",
     "process": "0.11.10",
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "lottie-react-native": "3.5.0",
     "metro-react-native-babel-preset": "0.63.0",
     "path-browserify": "1.0.1",
-    "payjoin-client": "git+https://github.com/bitcoinjs/payjoin-client.git#31d2118a4c0d00192d975f3a6da2a96238f8f7a5",
+    "payjoin-client": "git+https://github.com/bitcoinjs/payjoin-client.git#bip78",
     "pbkdf2": "3.1.1",
     "prettier": "2.1.1",
     "process": "0.11.10",
--- a/class/deeplink-schema-match.js
+++ b/class/deeplink-schema-match.js
@@ -330,6 +330,7 @@ class DeeplinkSchemaMatch {
     let address = uri || '';
     let memo = '';
     let payjoinUrl = '';
+    let payjoinOutputSubstitutionDisabled = false;
     try {
       parsedBitcoinUri = DeeplinkSchemaMatch.bip21decode(uri);
       address = 'address' in parsedBitcoinUri ? parsedBitcoinUri.address : address;
@@ -344,9 +345,12 @@ class DeeplinkSchemaMatch {
         if ('pj' in parsedBitcoinUri.options) {
           payjoinUrl = parsedBitcoinUri.options.pj;
         }
+        if ('pjos' in parsedBitcoinUri.options) {
+          payjoinOutputSubstitutionDisabled = parsedBitcoinUri.options.pjos === "0";
+        }
       }
     } catch (_) {}
-    return { address, amount, memo, payjoinUrl };
+    return { address, amount, memo, payjoinUrl, payjoinOutputSubstitutionDisabled };
   }
 }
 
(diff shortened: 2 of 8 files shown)


junderw:
ummm... we might delete that branch.

please don't use it.

at the very least use a commit hash.

ideally we should publish to npm.

junderw:
Please use `"payjoin-client": "1.0.0-beta",` as the dependency.
You can install via npm with `npm install [--save|--save-dev] payjoin-client@beta`

Overtorment:
bad news, payjoin client now uses some methods in `URLSearchParams` that are not supported in react-native runtime (but probably supported in nodejs cli, so its hard to test for it, tests will pass).
In particular, this one is throwing  `node_modules/react-native/Libraries/Blob/URL.js` :

```
// Small subset from whatwg-url: https://github.com/jsdom/whatwg-url/tree/master/lib
// The reference code bloat comes from Unicode issues with URLs, so those won't work here.
export class URLSearchParams {
...
  set(name, value) {
    throw new Error('not implemented');
  }
...
```

can you pls rewrite it so `URLSearchParams` is not used?



image

Kukks:
> bad news, payjoin client now uses some methods in `URLSearchParams` that are not supported in react-native runtime (but probably supported in nodejs cli, so its hard to test for it, tests will pass).
> In particular, this one is throwing `node_modules/react-native/Libraries/Blob/URL.js` :
> 
> ```
> // Small subset from whatwg-url: https://github.com/jsdom/whatwg-url/tree/master/lib
> // The reference code bloat comes from Unicode issues with URLs, so those won't work here.
> export class URLSearchParams {
> ...
>   set(name, value) {
>     throw new Error('not implemented');
>   }
> ...
> ```
> 
> can you pls rewrite it so `URLSearchParams` is not used?
> 
> image

What if you included a polyfill: https://github.com/ungap/url-search-params ?

Overtorment:
> > bad news, payjoin client now uses some methods in `URLSearchParams` that are not supported in react-native runtime (but probably supported in nodejs cli, so its hard to test for it, tests will pass).
> > In particular, this one is throwing `node_modules/react-native/Libraries/Blob/URL.js` :
> > ```
> > // Small subset from whatwg-url: https://github.com/jsdom/whatwg-url/tree/master/lib
> > // The reference code bloat comes from Unicode issues with URLs, so those won't work here.
> > export class URLSearchParams {
> > ...
> >   set(name, value) {
> >     throw new Error('not implemented');
> >   }
> > ...
> > ```
> > 
> > 
> > can you pls rewrite it so `URLSearchParams` is not used?
> > image
> 
> What if you included a polyfill: https://github.com/ungap/url-search-params ?

extra dep from random nobody on interwebz. plus might require some work just to make in work in RN.
feels like your use-case can just be monkey coded..?

junderw:
Isn't this............ exactly why we made the payjoin client composable?

You can write your own requester class and pass it in instead...

junderw:
https://github.com/bitcoinjs/payjoin-client/blob/9241a826768b673f8d371f2d60db975d4139cccf/src/request.js#L35-L67

junderw:
If you look at the requester we provide it's trivial to re-implement using whatever request library react works with.

It's one of the reasons why we made the class interface so simple: IPayjoinRequester

https://github.com/bitcoinjs/payjoin-client/blob/9241a826768b673f8d371f2d60db975d4139cccf/src/request.d.ts#L20-L29

Overtorment:
> https://github.com/bitcoinjs/payjoin-client/blob/9241a826768b673f8d371f2d60db975d4139cccf/src/request.js#L35-L67

looks like Ill just need to copy-paste it completely, and pass it to payjoin client only to get rid this utils call:

```
    } else {
      this.payjoinRequester = new request_1.PayjoinRequester(
        utils_1.getEndpointUrl(opts.payjoinUrl, opts.payjoinParameters),
      );
    }
```

its the `utils.getEndpointUrl()` that is using `new URL()` and calling `.set()`
feels hacky.

junderw:
Actually, on second thought you just need to pass in `customFetch` (It's the second optional opt for the requester)

junderw:
Oh I see. It's not a problem with the async code but the URL parsing...

ugh, I hate JavaScript.

junderw:
This should be fixed by rewriting getEndpointUrl to not use standard Url...

Does querystring standard library play well with react?

junderw:
This needs to be fixed on the payjoin-client end and can not be fixed via composing.

Overtorment:
> This should be fixed by rewriting getEndpointUrl to not use standard Url...
> 
> Does querystring standard library play well with react?

its a roulette, hard to tell in advance. you code, with tests, everything works perfectly, and then falls apart in RN 

image

junderw:
I added some comments to the PR on the payjoin-client repo.

junderw:
I published a new version `1.0.0-beta2` under the same `beta` tag to npm.

Overtorment:
image

Overtorment:
can `encodeURIComponent` be used instead?

Overtorment:
https://github.com/bitcoinjs/payjoin-client/pull/3

Overtorment (changes requested):
``` 
screen/send/confirm.js#L5
'getEndpointUrl' is defined but never used   
```

otherwise lgtm. 
now Im going to test the build and do an actual payjoin

Overtorment:
everything worked well
image

package.json
```diff
@@ -109,7 +109,7 @@
     "lottie-react-native": "3.5.0",
     "metro-react-native-babel-preset": "0.63.0",
     "path-browserify": "1.0.1",
-    "payjoin-client": "git+https://github.com/bitcoinjs/payjoin-client.git#31d2118a4c0d00192d975f3a6da2a96238f8f7a5",
+    "payjoin-client": "git+https://github.com/bitcoinjs/payjoin-client.git#bip78",
```
Overtorment, line 112: when will this be released?

Overtorment:
pj dep bumped in other PR.
@Kukks, do I need to port smth from this branch? 

for example, currently we dont have `payjoinOutputSubstitutionDisabled` or `payjoinVersion`.

also, things seem to work even without `getEndpointUrlFunc()` which is a bit weird, no..?

PR #1942, opened 2020-10-02
DOC: fix command for running ios

eyal352:
fix for pod-install command, current podinstall command returns 404


ncoelho (approved):

Overtorment:
oops, fixed in 76bdc0b3

Issue #976, opened 2020-04-12
BW SIGNER: One BW (watch-only) creates tx, other BW (with priv keys, but always offline) signs it


Overtorment:
Do all through PSBT, inspired by https://github.com/paritytech/parity-signer
This is _not_ a separate app

Similar to coldcard integration

Overtorment:
@marcosrdz you might wana do this feature since you did awesome coldcard integration?

bitcoinheiro:
Very nice. I believe this is already available on iOS but still lacking on Android. I see import tx option on iOS but not on Android apk

PR #1962, opened 2020-10-08
ADD: multisig can cosign for other coordinator; REF: BC-UR is now dec…

Overtorment:
…oded centrally

With this PR it is now possible for BW to join consigning if its being coordinated by someone else:

image

see "Co-sign transaction". It opens a camera and reads PSBT from camera. TBH this is almost the same as just 'import transaction', but 'import transaction' is opening file browser, while in real life transaction can be on Cobo on QR, and its not handy  to save it as file.
If BW has any keys, it will offer to additionally cosign this transaction:

image

this should allow for example, to have 3 iphones with BW, and coordinate and sign transaction via QR-airgapped flow.

ALSO, BC-UR decode was refactored, and this potentially could break everything we have with camera: import wallet, import tx etc etc, so should be carefully tested

### DIFF
--- a/class/wallets/multisig-hd-wallet.js
+++ b/class/wallets/multisig-hd-wallet.js
@@ -865,4 +865,33 @@ export class MultisigHDWallet extends AbstractHDElectrumWallet {
 
     return sigsHave;
   }
+
+  /**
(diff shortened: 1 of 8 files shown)


Overtorment:
related: https://github.com/BlueWallet/BlueWallet/issues/976

marcosrdz (approved):

PR #1967, opened 2020-10-10
ADD: Privacy Settings Screen

marcosrdz:
IMAGE 2020-10-09 22:42:48


GladosBlueWallet:

 [android in browser] https://appetize.io/app/9ykkutu974qnm5e7axrmqw008c

[download apk](http://filestorage.bluewallet.io:1488/1602298557.apk)

screen/settings/SettingsPrivacy.js
```diff
@@ -0,0 +1,74 @@
...
+    },
+    scroll: {
+      ...styles.scroll,
+      backgroundColor: colors.background,
+    },
+    scrollBody: {
+      ...styles.scrollBody,
+      backgroundColor: colors.background,
+    },
+  };
```
limpbrains, line 38: Many styles that are not used. I can see only stylesWithThemeHook.scroll
Also `const styles = StyleSheet.create...` can be removed

blue_modules/appStateChange.js
```diff
@@ -0,0 +1,87 @@
...
+      return !!JSON.parse(clipboardAccessAllowed);
+    } catch {
+      return false;
+    }
+  };
+
+  AppStateChange.setReadClipboardAllowed = async value => {
+    setHasClipboardAccessAsked(JSON.stringify(true));
+    setIsClipboardAccessAllowed(JSON.stringify(value));
+  };
```
limpbrains, line 37: There is an inconsistency with regular and async functions.
`setReadClipboardAllowed` is an async func, but you are calling it from components as a regular one. Looks like all of them are actually async.

It is easy to lose exception, if something dies inside `setHasClipboardAccessAsked` user will never know.

It is better to have await everywhere and on a top level use try/catch and show some generic error if something goes wrong

blue_modules/appStateChange.js
```diff
@@ -0,0 +1,87 @@
+import { useAsyncStorage } from '@react-native-community/async-storage';
+import Clipboard from '@react-native-community/clipboard';
+import { Alert } from 'react-native';
+import loc from '../loc';
+
+function AppStateChange() {
```
limpbrains, line 6: 90% of BW code uses classes, I think it is better to have a consistent approach everywhere. Do you have a specific reason to use function as a constructor here ?

loc/en.json
```diff
@@ -273,6 +275,11 @@
     "password_explain": "Create the password you will use to decrypt the storage",
     "passwords_do_not_match": "Passwords do not match",
     "plausible_deniability": "Plausible deniability",
+    "privacy": "Privacy",
+    "privacy_read_clipboard": "Read Clipboard",
+    "privacy_read_clipboard_alert": "BlueWallet will display shortcuts for handling an invoice or address found in your clipboard.",
+    "privacy_additional_permissions": "Additional Permissions",
```
limpbrains, line 281: Does the name "Additional Permissions" fit here? More like "Application system permissions" or "System application settings"
I don't know

limpbrains (changes requested):

GladosBlueWallet:

 [android in browser] https://appetize.io/app/vwxyvcakb9g5h28mp7mrv9396m

[download apk](http://filestorage.bluewallet.io:1488/1602320282.apk)

loc/en.json
marcosrdz: Updated.

blue_modules/appStateChange.js
marcosrdz: I'd figured id write it using functions so "future proof it" since we've had to rewrite code to take advantage of hooks.

blue_modules/appStateChange.js
marcosrdz: updated

screen/settings/SettingsPrivacy.js
marcosrdz: updated

GladosBlueWallet:

 [android in browser] https://appetize.io/app/1ydm2uv6wbt766ht3kdze20w2w

[download apk](http://filestorage.bluewallet.io:1488/1602325428.apk)

blue_modules/appStateChange.js
Overtorment: btw i tried to stick to classes early on, but in the end, I feel like combining modules/classes where it makes sense is okay, taking the best of both worlds. so if module exposes a comfortable interface, its internals don't matter much

blue_modules/appStateChange.js
```diff
@@ -0,0 +1,38 @@
...
+    if (isAllowed) {
+      return Clipboard.getString();
+    } else {
+      return '';
+    }
+  };
+}
+
+AppStateChange.default = new AppStateChange();
+export default AppStateChange;
```
Overtorment, line 38: btw why is it called `AppStateChange` ? feels like it should be named just `clipboard`

blue_modules/appStateChange.js
```diff
@@ -0,0 +1,38 @@
...
+      }
+      return !!JSON.parse(clipboardAccessAllowed);
+    } catch {
+      await setIsClipboardAccessAllowed(JSON.stringify(true));
+      return true;
+    }
+  };
+
+  AppStateChange.setReadClipboardAllowed = value => {
+    setIsClipboardAccessAllowed(JSON.stringify(value));
```
Overtorment, line 24: cast to bool with `!!value`

Overtorment (changes requested):
mostly ltgm

blue_modules/appStateChange.js
marcosrdz: AsyncStorage only accepts strings.

GladosBlueWallet:

 [android in browser] https://appetize.io/app/9mk0ueqxqk3d4vv01nq2zz8554

[download apk](http://filestorage.bluewallet.io:1488/1602352910.apk)

blue_modules/BlueClipboard.js
```diff
@@ -0,0 +1,38 @@
+import { useAsyncStorage } from '@react-native-community/async-storage';
```
Overtorment, line 1: `blue_modules/BlueClipboard` is a tautology

`blue_modules/clipboard.js` better, no? and no caps

blue_modules/BlueClipboard.js
```diff
@@ -0,0 +1,38 @@
...
+      }
+      return !!JSON.parse(clipboardAccessAllowed);
+    } catch {
+      await setIsClipboardAccessAllowed(JSON.stringify(true));
+      return true;
+    }
+  };
+
+  BlueClipboard.setReadClipboardAllowed = value => {
+    setIsClipboardAccessAllowed(JSON.stringify(value));
```
Overtorment, line 24: `setIsClipboardAccessAllowed(JSON.stringify(!!value))` so no sly ass can make `setReadClipboardAllowed('cyka')`

Overtorment (changes requested):

GladosBlueWallet:

 [android in browser] https://appetize.io/app/bva5mpgvpv7mp834jz3ce8ujm0

[download apk](http://filestorage.bluewallet.io:1488/1602356050.apk)

PR #1993, opened 2020-10-18
support lnurl fallback scheme.

fiatjaf:
There was partial support mixed in with the code but it didn't work actually.
This PR also uses the `Lnurl` class to replace some old code in the lnurl-withdraw handler.

### DIFF
--- a/class/lnurl.js
+++ b/class/lnurl.js
@@ -20,9 +20,9 @@ export default class Lnurl {
   }
 
   static findlnurl(bodyOfText) {
-    var res = /^(http.*[&?]lightning=)?((lnurl)([0-9]{1,}[a-z0-9]+){1})/.exec(bodyOfText.toLowerCase());
+    var res = /^(?:http.*[&?]lightning=)?(lnurl1[02-9ac-hj-np-z]+)/.exec(bodyOfText.toLowerCase());
     if (res) {
-      return res[2];
+      return res[1];
     }
     return null;
   }
--- a/class/lnurl.js
+++ b/class/lnurl.js
@@ -20,9 +20,9 @@ export default class Lnurl {
   }
 
   static findlnurl(bodyOfText) {
-    var res = /,*?((lnurl)([0-9]{1,}[a-z0-9]+){1})/.exec(bodyOfText.toLowerCase());
+    var res = /^(http.*[&?]lightning=)?((lnurl)([0-9]{1,}[a-z0-9]+){1})/.exec(bodyOfText.toLowerCase());
     if (res) {
-      return res[1];
+      return res[2];
     }
     return null;
   }
@@ -36,7 +36,7 @@ export default class Lnurl {
   }
 
   static isLnurl(url) {
-    return url.toLowerCase().startsWith('lnurl1');
+    return Lnurl.findlnurl(url) !== null
   }
 
   async fetchGet(url) {
--- a/class/deeplink-schema-match.js
+++ b/class/deeplink-schema-match.js
@@ -3,6 +3,7 @@ import AsyncStorage from '@react-native-community/async-storage';
 import RNFS from 'react-native-fs';
 import url from 'url';
 import { Chain } from '../models/bitcoinUnits';
+import Lnurl from './lnurl';
 import Azteco from './azteco';
 const bitcoin = require('bitcoinjs-lib');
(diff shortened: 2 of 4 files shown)


Overtorment (approved):
lgtm

class/lnurl.js
```diff
@@ -20,9 +20,9 @@ export default class Lnurl {
   }
 
   static findlnurl(bodyOfText) {
-    var res = /,*?((lnurl)([0-9]{1,}[a-z0-9]+){1})/.exec(bodyOfText.toLowerCase());
+    var res = /^(http.*[&?]lightning=)?((lnurl)([0-9]{1,}[a-z0-9]+){1})/.exec(bodyOfText.toLowerCase());
```
junderw, line 23:
```js
var res = /^(?:http.*[&?]lightning=)?(lnurl1[02-9ac-hj-np-z]+)/.exec(bodyOfText.toLowerCase());
if (res) {
  return res[1];
}
```

1. Add ?: to the beginning of () to make a non-capturing group.
2. No need for () around lnurl
3. lnurl is bech32, the separator is 1, not "unlimited numbers"
4. lnurl is bech32, this is a limited set of characters `[02-9ac-hj-np-z]` and not "alphanumerical"
5. {1} is meaningless, so is the group it modifies
6. Since lnurl is now the first captured group, we return res[1] not res[2]

junderw: Also, You can verify the regex for bech32 with this code using the const alphabet from bech32 library

```js
const bech32 = 'qpzry9x8gf2tvdw0s3jn54khce6mua7l';
let lastLetter = '';
console.log(
  bech32
    .split('')
    .sort()
    .reduce((final, c, idx) => {
      if (final === '') return c;
      if (idx === bech32.length - 1) return final + c;
      const endLetter = final.slice(-1);

      if (endLetter !== '-') {
        lastLetter = endLetter;
        if (lastLetter.codePointAt(0) === c.codePointAt(0) - 1) {
          lastLetter = c;
          return final + '-';
        } else {
          return final + c;
        }
      } else {
        if (lastLetter.codePointAt(0) === c.codePointAt(0) - 1) {
          lastLetter = c;
          return final;
        } else {
          return final + lastLetter + c;
        }
      }
    }, '')
);
```

junderw: regex is hated by most programmers because if you take the normal dev approach of "fiddle around with it until one test case works" has about a 99.999999999% chance of having a bug in it, because regex is complicated, and user provided strings can be anything (usually unforseen as to how crazy the strings can become)

plus regex is hard to parse in your brain when reading the code.

I worked on translation assistance software for years... I also hate regex. lol

junderw (changes requested):
The regex comments I made

class/lnurl.js
junderw: That being said, no guarantee my suggestion has no bugs in it. :-P

Overtorment:
There is a saying. Once programmer has a problem and he thinks, okay I can
solve it with regexp, starting from that moment he has 2 problems kek

Overtorment:
@fiatjaf are you going to work on this?

fiatjaf:
For some reason I thought @junderw's review was already a fix. I've copied his regex now. It seems to work. I trust it because he guaranteed his suggestion has no bugs.

junderw (approved):
> he guaranteed his suggestion has no bugs.

LOL, no. I did not guarantee anything such as that.

But I do guarantee it has LESS bugs that what was originally there. :-)

fiatjaf:
2020-10-26-212525_683x651_scrot

You can't deny this.

junderw:
> You can't deny this.

Yes I can. I could deny it by saying someone logged into my account to write that comment.

Whether the deniable is credible or not is another question. :-P

junderw:
(Not to mention that anyone who has a basic understanding of regex and how most developers perceive regex that sees someone say "I guarantee my regex has no bugs" can deduce with high certainty that they are being sarcastic. (even without the ":-P" at the end))

fiatjaf:
Sorry, I faked the screenshot, I thought it would be funny and you would notice it (or maybe you did, but other people reading may be confused).

Never trust a screenshot!

marcosrdz (approved):

limpbrains (approved):

PR #2047, opened 2020-11-02
FIX: hodlhodl - my contracts - sorted by creation time (closes #2037)

### LINKED ISSUE #2037: Scroll down in My Contracts
Greetings everyone. I am using the app on IOS and i's updated. I just noticed it seems to be impossible to scroll down in hodlhodl My Contracts. So I cannot see my most recent contracts now that they fill the whole screen.

Also I'd recommend that the order presenting contracts should be inverted, i.e. most recent contracts appearing first. Makes more sense for usability IMHO.

Best,
J

Overtorment:


marcosrdz (approved):

GladosBlueWallet:

PR #2170, opened 2020-11-20
FIX: better fee estimation based on mempool (closes #2154)

### LINKED ISSUE #2154: Switch to mempool-based fee estimation
we can switch to mempool fee estimation. dont forget to cover estimate calculations with unit tests

```
$ echo '{"id": "1", "method": "mempool.get_fee_histogram", "params": []}' | timeout 2 nc electrum3.bluewallet.io  50001 
{"jsonrpc": "2.0", "result": [[99, 100428], [86, 112465], [72, 137337], [59, 123113], [48, 147850], [46, 168821], [35, 188240], [28, 192644], [23, 199732], [18, 232113], [12, 279858], [7, 268914], [4, 304525], [2, 765331], [1, 414467]], "id": "1"}
```

```
Result

The histogram is an array of [fee, vsize] pairs, where vsizen is the cumulative virtual size of mempool transactions with a fee rate in the interval [feen-1, feen], and feen-1 > feen.

Fee intervals may have variable size. The choice of appropriate intervals is currently not part of the protocol.
```

https://electrumx-spesmilo.readthedocs.io/en/latest/protocol-methods.html#mempool-get-fee-histogram

Overtorment:


Overtorment:
for next block estimation it is closer to what mempool space shows
image

the problem is estimation for 18th block and 144 block (thats our targets for medium and slow)
as currently mempool is filled only for 14 blocks, so theoretically minumal fee will work for 18th and 144th block, so we render 1



so maybe it should be calculated as a fraction of fast fees
which is nextblock
or use mempool nextblock estimator as "medium", bitcoind estimatefee(1 block) as fast, and slow as 1, always


and I just checked, mempool peaked 90mb once this year, so mempool estimation for 1 day (144 blocks) wont really work


@limpbrains what are your thoughts on this?

GladosBlueWallet:

 [android in browser] https://appetize.io/app/ntjrv7efng6mn8ub194ehqbc1r

[download apk](http://filestorage.bluewallet.io:1488/1605903822.apk)

limpbrains:
My concern - is this fee estimation algorithm really better? Do you have an experiment to prove it?

I understand that current method is far from perfect, but this solution is too naive and I'm not sure it will not harm users. 

IMHO good fee estimation should be based on both mempool and blockchain history and couldn't be done on BW, it should be server-side. And it should be tested against historical data.

Maybe it's better to use some third-party service for now and think how to create inhome solution. Current electrum estimation can be used as a reserve. 

Mempool peaked at 90mb in October, according to https://jochen-hoenicke.de/queue/#0,30d

GladosBlueWallet:

 [android in browser] https://appetize.io/app/t0mgfwm66wqwn87twd9rpn3acw

[download apk](http://filestorage.bluewallet.io:1488/1606042498.apk)

Overtorment:
afaik, current call we are using `estimatefee` does exactly this - takes in account past blocks fees, and applies some slow decay mechanism for it.
the decay is quite slow - even if mempool clears it still sugests  100-200 sat/b for next block.
maybe we can invent  hybrid estimation with this `fee histogram`.
from what I witnessed, next block estimation based on mempoolis okayish - at least it matches what mempool.space offers.
so maybe we could take proportions from `estimatefee` (between 1 block, 18 and 144 blocks estimation) and appy them to next block fee derived from mempool, and it will look legit

limpbrains:
> so maybe we could take proportions from `estimatefee` (between 1 block, 18 and 144 blocks estimation) and appy them to next block fee derived from mempool, and it will look legit

I agree, this sounds good!

GladosBlueWallet:

 [android in browser] https://appetize.io/app/de41y3akcct49d3791ke1wkt94

[download apk](http://filestorage.bluewallet.io:1488/mempool-fees-estimation-1612195304.apk)

marcosrdz (approved):

ncoelho (approved):

limpbrains (approved):

GladosBlueWallet:

PR #2203, opened 2020-11-24
OPS: updated payjoin dep

Overtorment:


GladosBlueWallet:

 [android in browser] https://appetize.io/app/pg8j21t87jw38hu44d0vb599fc

[download apk](http://filestorage.bluewallet.io:1488/1606235724.apk)

screen/send/confirm.js
```diff
@@ -46,6 +46,18 @@ export default class Confirm extends Component {
...
+   * we need to look into `recipients`, find destination address and return its outputScript
+   * (needed for payjoin)
+   *
+   * @return {string}
+   */
+  getPaymentScript() {
+    for (const recipient of this.state.recipients) {
+      return bitcoin.address.toOutputScript(recipient.address);
+    }
+  }
```
limpbrains, line 59: You just take the first value from the recipients list and evaluate the outputscript.  are you sure that this is the destination address?
Looks like this function doesn't match it's description.

limpbrains (changes requested):

screen/send/confirm.js
Overtorment: pretty sure. in payjoin, there is only one destination. in batch send payjoin wont work.
plus, I tested it on real payjoin-enabled server  https://btc.donate.kukks.org

GladosBlueWallet:

PR #2219, opened 2020-11-28
REF: no autconnect in BlueElectrum

limpbrains:
Some wallets depend on BlueElectum, BlueElectum starts server connection when imported, this produces unwanted side effects. Connection should be closed before test is completed. 
To solve this I moved connection initialization from BlueElectum to App.js
Now tests are required to run BlueElectum.mainConnect() before testing.

Another issue - lots of jest warnings because of console.log after tests are complete. To solve this I added `await BlueElectrum.waitTillDisconnected();` helper to use in tests

### DIFF
--- a/tests/.eslintrc
+++ b/tests/.eslintrc
@@ -1,0 +1,7 @@
+{
+  "env":{
+    "es6": true,
+    "jasmine": true,
+    "jest": true
+  }
+}
--- a/tests/e2e/init.js
+++ b/tests/e2e/init.js
@@ -1,4 +1,4 @@
-/* global jasmine, jest, beforeAll, afterAll, beforeEach, device */
+/* global device */
 
 const detox = require('detox');
 const config = require('../../package.json').detox;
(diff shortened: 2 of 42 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/c48qez3zp3v7g2gx16vvckb338

[download apk](http://filestorage.bluewallet.io:1488/1606575375.apk)

Overtorment:
maybe youre having issues because youre _importing_ BlueElectum instead of _requiring_ it. I never have any issues with it, and no jest warnings at all. for example:

image


image



as for `waitTillDisconnected`. killing socket is a momentary operation, unlike connection. so `BlueElectrum.forceDisconnect();` should be enough, and I really dont have any jest warnings on any itnegration tests, otherwise Id fix them a long time ago (in fact I made `BlueElectrum.forceDisconnect();` specifically to fix those warnings when I had them). maybe youre experiencing some race conditions because youre importing BlueElectum instead of requirong (requiring same file more than once does nothing as it is cached - this is how we ensure there is only one tcp connection to electrum server)

limpbrains:
`rm -rf node_modules` helped me. I guess I had some old code, which prints `closecloseclose` after connection close and jest was arguing about it. Now I seed no errors

tests/.eslintrc
```diff
@@ -0,0 +1,7 @@
+{
```
Overtorment, line 1: this looks like a valid refactoring. maybe worth porting to a brand new pr

tests/e2e/init.js
```diff
@@ -1,4 +1,4 @@
-/* global jasmine, jest, beforeAll, afterAll, beforeEach, device */
+/* global device */
```
Overtorment, line 1: and this

PR #2224, opened 2020-11-30
FIX: UI would hang if await took too long due to network conditions

marcosrdz:


GladosBlueWallet:

 [android in browser] https://appetize.io/app/d8jn96ncxw9e626xgcx5etgrxw

[download apk](http://filestorage.bluewallet.io:1488/1606704001.apk)

limpbrains (changes requested):
this.reCalcTx depends on state.networkTransactionFees, state.fee and on wallet UTXO and should be called after you fetch them all

GladosBlueWallet:

 [android in browser] https://appetize.io/app/4ghuqtygu2mcxbrnc274xv74f0

[download apk](http://filestorage.bluewallet.io:1488/1606748680.apk)

screen/send/details.js
```diff
@@ -367,16 +367,17 @@ export default class SendDetails extends Component {
...
-      }
-    } catch (_) {} // either sleep expired or recommendedFees threw an exception
+    const recommendedFeesPromise = Promise.race([NetworkTransactionFees.recommendedFees(), this.context.sleep(2000)]).then(
+      recommendedFees => {
+        if (recommendedFees.fastestFee) {
+          AsyncStorage.setItem(NetworkTransactionFee.StorageKey, JSON.stringify(recommendedFees));
+          this.setState({
+            fee: recommendedFees.fastestFee.toString(),
+            networkTransactionFees: recommendedFees,
+          });
```
limpbrains, line 377: this.reCalcTx also should be called after saving networkTransactionFees to state

limpbrains (changes requested):

GladosBlueWallet:

 [android in browser] https://appetize.io/app/fdccn11kdyy7zg33ncmanhx10r

[download apk](http://filestorage.bluewallet.io:1488/1606762297.apk)

limpbrains (commented):
I can't see how the last version flow different from the original one. There are 2 big promises resolved one after another.

I was wrong previously. Second version (commit 9863d31e82bd37ea02747ba9b8a2c355810da2a6) is correct one and looks better - you have 2 promises resolved simultaneously. 

My idea was that `reCalcTx` can be executed after each of 2 promises, not after all of them. This will give user more instant feedback. But now when I'm looking again at it, putting `reCalcTx` after Promise.all looks fine.

Also you can use `await` to resolve `Promise.all([recommendedFeesPromise, fetchUTXOPromise])`, it will look nicer

marcosrdz:
> I can't see how the last version flow different from the original one. There are 2 big promises resolved one after another.
> 
> 
> 
> I was wrong previously. Second version (commit 9863d31e82bd37ea02747ba9b8a2c355810da2a6) is correct one and looks better - you have 2 promises resolved simultaneously. 
> 
> 
> 
> My idea was that `reCalcTx` can be executed after each of 2 promises, not after all of them. This will give user more instant feedback. But now when I'm looking again at it, putting `reCalcTx` after Promise.all looks fine.
> 
> 
> 
> Also you can use `await` to resolve `Promise.all([recommendedFeesPromise, fetchUTXOPromise])`, it will look nicer

The goal is to NOT use await for this process since I've witnessed cases where the UI hangs for a while waiting for either Await to finish. We want the process to continue running without the UI hanging, thus the use of "then" instead of await.

limpbrains:
@marcosrdz are you saying async/await works differently than Promise.then ? I thought async/await is just a Promise syntax sugar.
I believe UI can freeze because of unhandled exception, not because of async/await

Overtorment:
this was already a working code, and I spent time making sure it is working with bad/no network at all; 
promise.race + sleep ensures that it wont hang forever

PR #2285, opened 2020-12-08
Update broadcast.js

Ziya-Sadr:
hash seems to be incorrect for this field. I changed it to Hex.
I'm not a developer. so if I messed up sth, I'd like you to tell me how to change it correctly :)


Overtorment:
thanks

PR #2289, opened 2020-12-08
Capitalization – Part One

mcsaeid:
Please note that these changes are not specific to one English style guide. These are what all style guides agree upon. Most of these words or phrases are titles. Let me know which one does not make sense for you, and I will provide links that you can read and make an informed decision. The rest will be changed in part two.


Overtorment (changes requested):
json is broken. you can validate it on https://jsonparseronline.com

mcsaeid:
Thank you for the tip.

What do you think now?

Overtorment:
its a valid json now

Overtorment:
some strings werent capitalized probably because contextually, they looked better without capitalization. for example:

image

image

image


image

mcsaeid:
Contextually, titles _should be_ capitalized regardless of how they look. That is why we have rules in writing. Styles make sense as long as there is consistency. Here is an example of **inconsistency**. In the second screenshot, you have “window” and “mix / max,” both of which are not capitalized, but then you have “Cash.”

Aside from how they all look, it is not common at all. If you look closely at every major app or platform, wherever capitalization is called for, it is there—not because it looks right but because there is a certain style guide that was followed.

Instagram

Even in the cases that you mentioned and provided screenshots for, as someone who specializes in writing and editing, I still believe revision is the way to go. However, the final call is yours. I am only here to make things better.

marcosrdz (approved):

Overtorment (approved):

Overtorment:
@ncoelho

ncoelho (approved):

Overtorment:
pls fix conflicts

PR #2291, opened 2020-12-08
REF: Removing Extra Spaces

mcsaeid:
No spaces precede or follow slashes.


Overtorment:
not sure, someone put those spaces there, probably on purpose. @ncoelho thoughts?

mcsaeid:
Probably, but I don’t think that makes any difference on a technical level. Style-wise, the correct approach is to remove the spaces. Let’s see what Nuno thinks.

ncoelho (approved):

PR #2322, opened 2020-12-10
REF: Updating en.json

mcsaeid:
Since there were multiple merges, many of the changes were undone, including punctuation marks, capitalization, etc. But this is everything—nearly perfect.


loc/en.json
```diff
@@ -101,47 +101,47 @@
...
     "for": "For:",
     "lightning_invoice": "Lightning Invoice",
-    "has_been_paid": "This invoice has been paid for",
+    "has_been_paid": "This invoice has been paid.",
     "open_direct_channel": "Open direct channel with this node:",
     "please_pay": "Please pay",
     "preimage": "Preimage",
-    "sats": "sats",
-    "wasnt_paid_and_expired": "This invoice was not paid for and has expired."
+    "sats": "satoshis.",
```
Overtorment, line 115: no, sats are sats

loc/en.json
```diff
@@ -101,47 +101,47 @@
...
+    "wasnt_paid_and_expired": "This invoice was not paid and has expired."
   },
   "plausibledeniability": {
     "create_fake_storage": "Create Encrypted Storage",
     "create_password": "Create a password",
-    "create_password_explanation": "Password for fake storage should not match the password for your main storage.",
-    "help": "Under certain circumstances, you might be forced to disclose a password. To keep your coins safe, BlueWallet can create another encrypted storage with a different password. Under pressure, you can disclose this password to a 3rd party. If entered in BlueWallet, it will unlock a new “fake” storage. This will seem legit to the third party, but it will secretly keep your main storage with coins safe.",
-    "help2": "The new storage will be fully functional, and you can store some minimum amounts there so it looks more believable.",
+    "create_password_explanation": "Password for the fake storage should not match the password for your main storage.",
```
Overtorment, line 121: this probably breaks e2e tests. need to adjust spec file accordingly

loc/en.json
```diff
@@ -243,13 +243,13 @@
...
-    "electrum_host": "Host, for example {example}",
+    "electrum_error_connect": "Can’t connect to the provided Electrum server",
+    "electrum_host": "Host, for example, {example}",
     "electrum_port": "TCP port, usually {example}",
     "electrum_port_ssl": "SSL port, usually {example}",
-    "electrum_saved": "Your changes have been saved successfully. Restart may be required for changes to take effect.",
+    "electrum_saved": "Your changes have been saved successfully. Restarting BlueWallet may be required for the changes to take effect.",
     "electrum_settings": "Electrum Settings",
-    "electrum_settings_explain": "Set to blank to use default",
+    "electrum_settings_explain": "Leave blank to use default settings.",
```
Overtorment, line 252: user literally has to clean up input fields to revert to default settings. is that ok..?

loc/en.json
```diff
@@ -261,50 +261,50 @@
...
     "general_adv_mode": "Advanced Mode",
-    "general_adv_mode_e": "When enabled, you will see advanced options such as different wallet types, the ability to specify the LNDHub instance you wish to connect to and custom entropy during wallet creation.",
+    "general_adv_mode_e": "When enabled, you will see advanced options such as different wallet types, the ability to specify the LNDHub instance you wish to connect to, and custom entropy during wallet creation.",
     "general_continuity": "Continuity",
     "general_continuity_e": "When enabled, you will be able to view selected wallets, and transactions, using your other Apple iCloud connected devices.",
-    "groundcontrol_explanation": "GroundControl is a free opensource push notifications server for Bitcoin wallets. You can install your own GroundControl server and put its URL here to not rely on BlueWallet's infrastructure. Leave blank to use default",
+    "groundcontrol_explanation": "GroundControl is a free, open-source push notifications server for Bitcoin wallets. You can install your own GroundControl server and put its URL here to not rely on BlueWallet’s infrastructure. Leave blank to use default settings.",
```
Overtorment, line 267: in this case to use defautl groundcontrol __server__

loc/en.json
```diff
@@ -261,50 +261,50 @@
...
     "network": "Network",
     "network_broadcast": "Broadcast Transaction",
     "network_electrum": "Electrum Server",
     "not_a_valid_uri": "Not a valid URI",
     "notifications": "Notifications",
     "password": "Password",
-    "password_explain": "Create the password you will use to decrypt the storage",
-    "passwords_do_not_match": "Passwords do not match",
+    "password_explain": "Create the password you will use to decrypt the storage.",
+    "passwords_do_not_match": "Passwords do not match.",
```
Overtorment, line 282: also probably breaks e2e tests

loc/en.json
```diff
@@ -422,11 +422,11 @@
...
     "export_coordination_setup": "Export Coordination Setup",
     "cosign_this_transaction": "Co-sign this transaction?",
-    "lets_start": "Let's start",
+    "lets_start": "Let’s start",
     "create": "Create",
     "provide_key": "Provide key",
-    "native_segwit_title": "Best practice",
-    "wrapped_segwit_title": "Best compatibility",
+    "native_segwit_title": "Best Practice",
+    "wrapped_segwit_title": "Best Compatibility",
```
Overtorment, line 429: this is the screen where its used, are you sure it must be capitalized here?

image

loc/en.json
```diff
@@ -441,41 +441,41 @@
...
+    "input_path": "Insert Derivation Path",
     "input_path_explain": "Skip to use the default one ({default})",
     "view_edit_cosigners_title": "Edit Cosigners",
     "ms_help": "Help",
-    "ms_help_title": "How Multisig Vaults work. Tips and tricks",
-    "ms_help_text": "A wallet with multiple keys, to exponentially increase security or for shared custody.",
-    "ms_help_title1": "Multiple devices are advised",
-    "ms_help_1": "The Vault will work with other BlueWallet apps and PSBT compatible wallets, like Electrum, Specter, Coldcard, Cobo vault, etc.",
+    "ms_help_title": "How Multisig Vaults Work: Tips and Tricks",
+    "ms_help_text": "A wallet with multiple keys, for increased security or sharing accounts",
```
Overtorment, line 468: "shared custody" is kinda commonly used term in this multisig monkey biz

loc/en.json
```diff
@@ -486,11 +486,11 @@
   "cc": {
     "change": "Change",
     "coins_selected": "Coins Selected ({number})",
-    "empty": "This wallet doesn't have any coins at the moment",
-    "freeze": "Freeze",
+    "empty": "This wallet doesn’t have any coins at the moment.",
+    "freeze": "Frozen",
```
Overtorment, line 490: also probably breaks e2e, tests are looking for string "Freeze". probably needs code readjustment so frozen UTXO has its own localization key

Overtorment (changes requested):

loc/en.json
mcsaeid: Got it. The way I was going about it was “sats” should be the norm currency (when showing the balance), but in the case of messages and everyday conversation—for instance, here, “Please pay [x] satoshis.”—we would opt for satoshis instead of sats. No worries, though. Sats **is** great.

loc/en.json
mcsaeid: Can you point out what I should usually look out for and change in the _bluewallet.spec.js_ file so that these breaks don’t happen in the future? I am trying to get the hang of it. Thanks.

loc/en.json
mcsaeid: Exactly. If the user wishes to use the default settings, they have to _leave_ the field _blank_ before proceeding. I have seen it being used in other apps and instances in general, but I also did a quick search before making the change just to make sure. I believe _leave blank_ is the standard term to go with.

loc/en.json
mcsaeid: Perfect. I will re-edit this accordingly to make it more clear and understandable. Thank you.

loc/en.json
mcsaeid: Got it. So, I should make sure the spec file has the same changes as this current file, right?

loc/en.json
mcsaeid: There is no must here in this case, but since these can be categorized as semi-titles, I thought it would look better if they were all capitalized. One thing that I am sure about is the acronyms in the parentheses. Because every letter represents a larger entity/word, all acronyms must be written capitalized, e.g. P2WSH or P2SH.

loc/en.json
mcsaeid: Got it. Let me change it back.

loc/en.json
mcsaeid: Yeah. Is that something I can look for and change? In the end, it should be “frozen.” I even checked Electrum to make sure.

mcsaeid:
I believe I made the necessary changes. Let me know what you think when you get a chance.

loc/en.json
Overtorment: should be 'frozen', but code uses same loc key for several places, so that's a bumer.

loc/en.json
Overtorment: image
look for something like

image

Overtorment (changes requested):
e2e + conflicts

mcsaeid:
> 
> 
> e2e + conflicts

So, if I got it right, I should fix the spec.json file but do it in the same PR. How can I do that? I apprecaite the help.

loc/en.json
mcsaeid: How does changing the code for this particular case work? Do you need find all the instances and correct them (as easy as that), or do you have to do/add something additional?

Overtorment:
> > e2e + conflicts
> 
> So, if I got it right, I should fix the spec.json file but do it in the same PR.

yes, looks like you've succeeded

Overtorment (approved):
lgtm

mcsaeid:
Any thoughts on how to prevent the tests from failing? These may require more coding knowledge than I currently have. @limpbrains @Overtorment

limpbrains:
in bluewallet.spec.js
`No, and don't ask me again` should be changed to `No, and don’t ask me again`

ncoelho (approved):

PR #2332, opened 2020-12-11
REF: bluewallet.spec.js

mcsaeid:


Overtorment:
dont update e2e tests in a separate PR, it will fail as well. do all the work in single pr

PR #2403, opened 2020-12-20
FIX: Fallback to English if case isn't found

marcosrdz:
When switching back to English, dayjs wouldnt change.

### DIFF
--- a/loc/index.js
+++ b/loc/index.js
@@ -128,6 +128,8 @@ const setDateTimeLocale = async () => {
   }
   if (localeForDayJSAvailable) {
     dayjs.locale(lang.split('_')[0]);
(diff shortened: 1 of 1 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/2qt21vftaax38hea3hjzmky520

[download apk](http://filestorage.bluewallet.io:1488/1608450092.apk)

limpbrains (approved):
Dayjs doesn't have a default locale?

marcosrdz:
It does but for example...when you switch from Russian to English, dayjs stays in Russian. Meaning that tx rows date label remain in Russian. This fixes that bug.

PR #2419, opened 2020-12-24
FIX: show more accurate precalculation fee on "Not enough balance." exception

### LINKED ISSUE #2404: Incorrect price calculation with coin control
Sending 1000$ will cost less than sending 100$. Banking allures :-)
The fee drops (actually) exactly from 410$ to 411$. AND
Until 410$: The fast & medium option shows the same price. Only by 411$ AND 412$, the medium price option is higher, than the fast. From 413$ to 1000$ the fast price option is as expected more expensive.

Also, price jumps between 0$, 5$, 10$.
Type: HD SegWit (BIP84 Benach32 Native) with 24 words imported from https://iancoleman.io/bip39/

I observed this in CHF and switched to $ to open the ticket.

Important: I have 4 Coins: $406.82 / $12.45 and Freezed: 2'081.52 / $11'466.48. I think this may show the error. (406+12 are close to the magic 411$).

YES! - This is hot: I unlocked the freezes. Then, the price drop moves up from 411$ to 13'950$. Same behavior there. The medium option for 13'950$ is higher than the fast prices.

I allways just changed the amount. No changes in inputs or outputs. Allow fee bump On/Off shows no impact.

I first beleaved, that diffrent wallet types result in different behavior. As the phenomenon is not observable with my watch-ony wallet, imported from CoboVault. And is not observale with Legacy (P2PKH), SegWit (P2SH), Legacy (BIP 44PSPKH), Multisig Vault. But there i have either not used coin control or zero or only one UTXO.

limpbrains:
When the user tries to send more funds than a wallet balance, BW still tries to precalculate fee to show something. 
To do this all transaction output values was replaced with dust value. In result, this tx always smaller than maximum possible option, because it has smaller number of inputs. 

Now BW will create tx with the same amount of outputs, but one of them will be MAX. It will allow showing more accurate fee estimation.

Fixes #2404

### DIFF
--- a/class/wallets/legacy-wallet.js
+++ b/class/wallets/legacy-wallet.js
@@ -276,8 +276,8 @@ export class LegacyWallet extends AbstractWallet {
     if (!changeAddress) throw new Error('No change address provided');
 
     let algo = coinSelectAccumulative;
-    if (targets.length === 1 && targets[0] && !targets[0].value) {
-      // we want to send MAX
+    // if targets has output without a value, we want send MAX to it
+    if (targets.some(i => !('value' in i))) {
       algo = coinSelectSplit;
     }
 
@@ -285,7 +285,7 @@ export class LegacyWallet extends AbstractWallet {
 
     // .inputs and .outputs will be undefined if no solution was found
     if (!inputs || !outputs) {
-      throw new Error('Not enough balance. Try sending smaller amount');
+      throw new Error('Not enough balance. Try sending smaller amount or decrease the fee.');
     }
 
     return { inputs, outputs, fee };
--- a/class/wallets/multisig-hd-wallet.js
+++ b/class/wallets/multisig-hd-wallet.js
@@ -3,8 +3,6 @@ import bip39 from 'bip39';
 import b58 from 'bs58check';
 import { decodeUR } from 'bc-ur';
 const BlueElectrum = require('../../blue_modules/BlueElectrum');
-const coinSelectAccumulative = require('coinselect/accumulative');
-const coinSelectSplit = require('coinselect/split');
 const HDNode = require('bip32');
 const bitcoin = require('bitcoinjs-lib');
 const createHash = require('create-hash');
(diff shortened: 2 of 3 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/55y9qgth1j96g156czt0v20q1r

[download apk](http://filestorage.bluewallet.io:1488/1608818199.apk)

class/wallets/legacy-wallet.js
```diff
@@ -276,16 +276,16 @@ export class LegacyWallet extends AbstractWallet {
     if (!changeAddress) throw new Error('No change address provided');
 
     let algo = coinSelectAccumulative;
-    if (targets.length === 1 && targets[0] && !targets[0].value) {
-      // we want to send MAX
+    // if targets has output without a value, we want send MAX to it
```
Overtorment, line 279: this change should apply to multisig wallet as well

Overtorment: in fact, this logic in multisig wallet is duplicated for some reason. maybe some refactoring is due

Overtorment (changes requested):

class/wallets/legacy-wallet.js
limpbrains: refactored

GladosBlueWallet:

 [android in browser] https://appetize.io/app/xvhbf46g5c0u63zb5krz9gphwg

[download apk](http://filestorage.bluewallet.io:1488/1608826175.apk)

class/wallets/multisig-hd-wallet.js
```diff
@@ -809,22 +807,9 @@ export class MultisigHDWallet extends AbstractHDElectrumWallet {
     if (targets.length === 0) throw new Error('No destination provided');
     if (this.howManySignaturesCanWeMake() === 0) skipSigning = true;
 
-    if (!changeAddress) throw new Error('No change address provided');
+    const { inputs, outputs, fee } = this.coinselect(utxos, targets, feeRate, changeAddress);
```
Overtorment, line 810: bueno.

Overtorment (approved):

marcosrdz (approved):

GladosBlueWallet:

PR #2495, opened 2021-01-13
FIX: disable RPC batching for Fulcrum electrum server

Overtorment:


marcosrdz (approved):

GladosBlueWallet:

 [android in browser] https://appetize.io/app/bt9r21a805ckrpduuzyc2ap6pm

[download apk](http://filestorage.bluewallet.io:1488/1610559011.apk)

GladosBlueWallet:

cculianu:
Hey guys wanted to let you know that I'm working on implementing batching in Fulcrum and when it's done and released.

I may making it configurable so that the admin can disable it (maybe), or I may have it always-on.  Not sure yet, depending on whether I can find ways to totally DoS the server with it or not.  I may also just limit the complexity of batch requests.

I have a question: how large are the batches potentially coming-in from BlueWallet and what do they typically contain in them?

Overtorment:
@cculianu batching or not should not affect DDOSability of the server. take a look at electrumx, they assign weight per operation, and it doesnt matter if this operation is inside batch or not, same ip client that exceeds some quota (of weights) gets disconnected.

bw's batches are typically between 50 and 100 operations in a batch. this was decided empyrical, also looking at  gap_limit (hint: most consumer electrum-based wallets will have gap limit ~2), and operations are ususally get scripthash balance and get scripthash history

cculianu:
Skimming the sources for BlueWallet it does seem like I see some batchSize of 200 in there for at least 1 of the calls.  Is that the case? Also do you ever send more than 1 batch at a time or is it just 1 batch at a time?

Overtorment:
one batch at a time, otherwise electrumx would disconnect us as an anti-ddos measure. i remember i was experimenting and tuning batch size numbers specifically to make electrumx anti-ddos happy

cculianu:
Any memory on what the maximum batch size I can expect is?  I am thinking of limiting them to either `200` or at most `345`.  Let me know if that would work flawlessly with BlueWallet.

Of course after I'm done I'll just install BlueWallet myself and tinker with it..  anyway but it does save me time to ask you.

Thanks in advance for answering my questions.

Overtorment:
i dont remember having batchsizes greater than 200

cculianu:
Ok, feature was added as of Fulcrum version 1.6.0. Please see #4422.

PR #2500, opened 2021-01-14
DEL: Reset keychain after uninstall

marcosrdz:


GladosBlueWallet:

 [android in browser] https://appetize.io/app/d8a0up2hhvvzn34c0qcvtnw7a0

[download apk](http://filestorage.bluewallet.io:1488/1610639448.apk)

screen/wallets/walletMigrate.js
```diff
@@ -15,41 +10,6 @@ export default class WalletMigrate {
 
   // 0: Let's start!
   async start() {
-    if (Platform.OS === 'ios' && !isCatalyst) {
```
Overtorment, line 18: image

Overtorment (approved):
lgtm

limpbrains:
How does this work? I see only removing of this feature.

marcosrdz:
> How does this work? I see only removing of this feature.

keychain and app live separately. this allows you to remove the app from your device, but keep the data. sadly, there were multiple bugs after the app grew and started using App Groups (for widgets and watch app) which resulted in this feature being triggered,with wrong value, without the user requesting it.

Overtorment:
@limpbrains so the idea is that this feature randomly wipes keystores of random newb users, and better to just remove it instead of having a headache where the bug is

limpbrains (approved):

ncoelho (approved):

GladosBlueWallet:

PR #2541, opened 2021-01-28
ADD: AEZEED mnemonics support (from LND; only primary native-segwit c…

### LINKED ISSUE #1543: Support AEZEED mnemonic seeds (compatibility with LND onchain)
https://github.com/bitcoinjs/aezeed

Overtorment:
…hain atm; closes #1543)


Overtorment:
I was thinking, should I build-in funds recovery support

but eventually decided no: aezeed has 10 deriviations internally (10 times bigger than bip84), and nature of LND operation is such that indexes can easily go forward for hundreds. so proper funds recovery from crashed LND would require scanning thousands and thousands of addresses, which is better done via specialized software
so with this you can keep an eye on your LND mnemonic, top it up or send from it

also, get it's xpub and share it with someone if you want someone to top-up your LND

Overtorment:
@junderw can you pls take a look?
I had to tweak a bit the package you made (same as with bip38): the package relies on `scryptsy`, which relies on `crypto` only present in nodejs
so i had to patch it to use pure-js `pbkdf2` (the only thing needed from `crypto` module) and also patch aezeed to use my own scrypt implementation when it runs in react-native environment (kotlin/swift x100 faster here when it runs in RN).

this feels suboptimal. do you think there is something we can do on `bitcoinjs` org level..?

GladosBlueWallet:

 [android in browser] https://appetize.io/app/nz0vz2u75h7qm44m75caxf1zzr

[download apk](http://filestorage.bluewallet.io:1488/add-aezeed-support-1611839509.apk)

junderw:
Can't you stub out the dependencies using package-lock.json?

You can add a dependency as another dependency (ie. tell npm to download pack-xyz and use it to resolve any imports or requires to pack-abc) and you can do it for subdependencies.

It requires manually editing the package-lock.json though.

As long as your library has the same interface as scryptsy etc. it should be fine.

Overtorment:
@junderw sounds even more hacky.
would be could to just be able to inject `scryptsy` dependency in aezeed & bip38 packages

Overtorment:
oh, @junderw, whats the deal with www.npmjs.com/package/junderw-crc32c  ?

junderw:
Re: crc32

https://github.com/junderw/js-crc32/commit/b06adcad53c4aa05693d2ad3bcf65997975896bd

This is the only commit I've changed.

The reason why is because aezeed uses an unpopular variant of the crc32 checksum algorithm, and the only library I could find that uses it was poorly written.

The library I forked was well tested and well written. The only difference was the constant, but it seemed like they weren't open to adding such a niche feature.

So I just added the new constant and renamed/published the package.

junderw:
> would be cool to just be able to inject scryptsy dependency in aezeed & bip38 packages

I disagree. Even if aezeed/bip38 supported some sort of scryptsy injection method, that would require setup / teardown code to get it working.

Admittedly that would be less code than what you currently have, but using package-lock is 0 code. You just literally tell bip38 to view package x as if it was package y, and all you need to do is to ensure package y has the same interface.

limpbrains:
I was able to recover full balance of my LND node. Works good

limpbrains (approved):

Overtorment:
hmmm.  I agree, less code is better, but manually editing `package-lock.json` is prone to errors. someone someday will decide to install deps from scratch by deleting package-lock.json and will bang his head against a table because of broken functionality.
another solution would be patches for node_modules.

so many options. don't know which is worse

marcosrdz (approved):

PR #2617, opened 2021-02-12
REF: load from disk

Overtorment:


GladosBlueWallet:

 [android in browser] https://appetize.io/app/bbzztexzwb8j675uy38c8yzfkm

[download apk](http://filestorage.bluewallet.io:1488/ref-load-from-disk-1613176947.apk)

class/app-storage.js
```diff
@@ -221,109 +216,104 @@ export class AppStorage {
...
-        const wallets = data.wallets;
-        for (const key of wallets) {
-          // deciding which type is wallet and instatiating correct object
-          const tempObj = JSON.parse(key);
-          let unserializedWallet;
-          switch (tempObj.type) {
-            case PlaceholderWallet.type:
+    }
+    if (data !== null) {
+      const realm = await this.getRealm();
```
marcosrdz, line 228: We should reset the realm storage if this fails. It could get corrupted forcing wallets to not load even though they are two separate things

marcosrdz (changes requested):

class/app-storage.js
Overtorment: elaborate

GladosBlueWallet:

 [android in browser] https://appetize.io/app/69w3x8kym7hp20bcd5b3gc9few

[download apk](http://filestorage.bluewallet.io:1488/ref-load-from-disk-1613221682.apk)

class/app-storage.js
marcosrdz: So lets say the user is playing around with APKs and installs a release with an older realm than the one in the last APK installed. Realm will not be able to handle the downgrade of db, we should call 

deleteAll() so the Realm db is reset. Right now theres no catch for it..it just becomes unusable cause the code doesnt proceed to other actions

Overtorment: its unrelated to this ticket. 
it falls into category "what if i install a really outdated build that wont work with current db format". frankly, this is not our problem, no one ever promised forward compatibility

marcosrdz (approved):

limpbrains (approved):

GladosBlueWallet:

PR #2621, opened 2021-02-13
FIX: Delivered unprocessed notifications would not process if it was …

marcosrdz:
…received with the app on background

Issue: if u receive a notification while being on a different app, and then returned to BW, it wouldnt acknowledge that u received a notification thus wallets wouldnt fetch txs


GladosBlueWallet:

 [android in browser] https://appetize.io/app/hygzm4qnkbtuqkx9rhhu30w0wr

[download apk](http://filestorage.bluewallet.io:1488/notificationsprocess-1613194577.apk)

limpbrains (approved):

GladosBlueWallet:

 [android in browser] https://appetize.io/app/5hxtjc8h8fuh0w0hqvu04eh1um

[download apk](http://filestorage.bluewallet.io:1488/notificationsprocess-1614009412.apk)

blue_modules/notifications.js
```diff
@@ -68,7 +68,7 @@ function Notifications(props) {
 
           // if user is staring at the app when he receives the notification we process it instantly
           // so app refetches related wallet
-          if (payload.foreground) props.onProcessNotifications();
+          props.onProcessNotifications();
```
Overtorment, line 71: you changed the line, but didnt change the comment that explains it.
anyway, there was a reason for that IF, i just don't exactly remember.
i need to think about it and reread this code

GladosBlueWallet:

Overtorment:
i reverted it via rebase & force push

App.js
```diff
@@ -192,10 +192,9 @@ const App = () => {
     await new Promise(resolve => setTimeout(resolve, 200));
     // sleep needed as sometimes unsuspend is faster than notification module actually saves notifications to async storage
     const notifications2process = await Notifications.getStoredNotifications();
-
+    const deliveredNotifications = await Notifications.getDeliveredNotifications();
```
Overtorment, line 195: whats the reson for this move?

blue_modules/notifications.js
Overtorment: ok, i think the reason for that IF is that app can get a notification even while its not launched - iOS executes small part of code even if app is unloaded from memory.
ofcourse, there are no wallets at the moment in RAM, so all we can do is save notification to asyncstorage and hope it will be processed later.
and that IF allows processing ON THE SPOT only notification user is seeing.
otherwise it goes to asyncstorage queue, and should be processed upon app state change (UNSUSPEND). if its not happening - debug there

PR #2712, opened 2021-02-26
FIX: Set biometric flag off when enabling encryption

marcosrdz:


GladosBlueWallet:

 [android in browser] https://appetize.io/app/6cv9zw5dyj9wq4jw2hz61pz5yc

[download apk](http://filestorage.bluewallet.io:1488/setbiometricflag-1614311792.apk)

Overtorment (changes requested):
why do we disable it?
storage encryption doesn't means we should not verify biometrics prior to showing wallet export/other risky stuff

marcosrdz:
> why do we disable it?
> storage encryption doesn't means we should not verify biometrics prior to showing wallet export/other risky stuff

You're right.

Overtorment:
@marcosrdz  dont forget to delete branches if you're closing the pr

Issue #2779, opened 2021-03-11
Lightning and Vault Icons


mcsaeid:
Lightning and Vault icons seem to be cut off from the top. I have noticed that the height for the Bitcoin icon is 381 pixels, which is the only one that is displayed correctly.

![1](https://user-images.githubusercontent.com/63400670/110800706-b85e7900-8291-11eb-9fb2-89471c3abe42.jpg)

![2](https://user-images.githubusercontent.com/63400670/110800716-bbf20000-8291-11eb-9c29-2084869c1ac1.jpg)

![3](https://user-images.githubusercontent.com/63400670/110800729-beecf080-8291-11eb-8bfb-f8d7a55bedce.jpg)

Since I don’t know how to fix this with code (but you probably do), I have edited the PNG files so that all three have the same height. The benefit now is that they all look similar in size.

![Icons](https://user-images.githubusercontent.com/63400670/110801726-c5c83300-8292-11eb-9f03-e5f5ab0e2723.jpg)

Please advise on how to upload the files and make a pull request.

![Disabled](https://user-images.githubusercontent.com/63400670/110800756-c4e2d180-8291-11eb-98b7-d9c8722b9de2.jpg)

ncoelho:
https://github.com/BlueWallet/BlueWallet/pull/2780

PR #2780, opened 2021-03-11
FIX: Wallet logomarks images

ncoelho:
Closes https://github.com/BlueWallet/BlueWallet/issues/2779

Screenshot 2021-03-11 at 15 46 21

PR #2781, opened 2021-03-11
FIX: Lightning and Vault Icons

mcsaeid:
Lightning and Vault icons seem to be cut off from the top. I have noticed that the height for the Bitcoin icon is 381 pixels, and it is the only one that is displayed correctly. Since I don’t know how to fix this with code (but you probably do), here are all of the icons with the same height. The benefit now is that they all look the same size.

Explained more in #2779.


ncoelho:
ahah, sorry was with everything opened to change fast...😄

https://github.com/BlueWallet/BlueWallet/pull/2780

mcsaeid:
Haha, no problem! Great job. The files I uploaded are a bit bigger, though—400 x 380 pixels. Will the logos still look good on larger screens?

What if SVGs were used?

Overtorment:
so its safe to close..?
closing.

Issue #990, opened 2020-04-15
Hierarchy navigator screen


Overtorment:
Need something like this:

![image](https://user-images.githubusercontent.com/1913337/79360291-9ce0d600-7f3b-11ea-9808-5a0e520ce14b.png)

A list of receive/change addresses with balances and transactions num.
There should be an option for each address to see (and copy) public key, priv key, sign/verify message, encrypt/decrypt message.
also, maybe button to request payment with a specific amount to this exact address, and option to assign text label to an address.

jaonoctus:
Any progress on this feature yet? Should I send some PR?

Overtorment:
no progress yet. I was itchy to build it myself but eventually got busy with other stuff.
PR would be nice! @ncoelho do we have designs for such screen..?

jaonoctus:
@ncoelho @Overtorment Maybe we should use a similar approach to that of Coin Control screen?

ncoelho:
I played with this concept. Could get some feedback before, not super polished and may be missing some things.

<img width="757" alt="Screenshot 2021-01-30 at 18 07 43" src="https://user-images.githubusercontent.com/1238429/106363013-34c27b80-6326-11eb-9543-7b1e8360936f.png">

Overtorment:
needs to display indexes for both internal (change) & external (receive) addresses.
needs to display number of transactions on each address - address might be empty (0 btc), but still have some transactions which means it is "used"
needs to display which index is the "tip" - next free address for internal & external hierarchy, and an option to show + 20 (or more..?) addresses in case user wants to manually give away his addresses to his payers

ncoelho:
Ok, I will adapt the proposed design to be able to handle more data upfront. Will update the issue later.

ncoelho:
Ok, this seems more solid to me. Any more feedback, let me know. 

<img width="779" alt="Screenshot 2021-02-01 at 16 54 05" src="https://user-images.githubusercontent.com/1238429/106482852-328b2900-64ae-11eb-8bb6-534980a825fb.png">

Overtorment:
maybe gray-out whole row for used addresses..?
![image](https://user-images.githubusercontent.com/1913337/106489086-4d5c9e00-64ac-11eb-9568-a00803f4f3b4.png)

jaonoctus:
Where can I add the action for this screen? Is "wallet details" (`screen/wallets/details.js`) a good place?

[...]

PR #2837, opened 2021-03-23
ADD: Hierarchy Navigator Screen

jaonoctus:
Solves #990

## Addresses Page

- [x] Create "Show addresses" page.
- [x] Add "Show addresses" button on wallet details page.
- [x] Fetch wallet "receive" and "change" (external/internal) addresses.
- [x] Show address index.
- [x] Fetch and show address balance.

## Address Actions

- [x] Copy Address to clipboard.

### DIFF
--- a/Navigation.js
+++ b/Navigation.js
@@ -36,6 +36,7 @@ import ExportMultisigCoordinationSetup from './screen/wallets/exportMultisigCoor
 import ViewEditMultisigCosigners from './screen/wallets/viewEditMultisigCosigners';
 import WalletXpub from './screen/wallets/xpub';
 import SignVerify from './screen/wallets/signVerify';
+import WalletAddresses from './screen/wallets/addresses';
 import BuyBitcoin from './screen/wallets/buyBitcoin';
 import HodlHodl from './screen/wallets/hodlHodl';
 import HodlHodlViewOffer from './screen/wallets/hodlHodlViewOffer';
@@ -400,6 +401,17 @@ const SignVerifyStackRoot = () => {
   );
 };
 
+const WalletAddressesStack = createStackNavigator();
+const WalletAddressesStackRoot = () => {
+  const theme = useTheme();
+
+  return (
+    <WalletAddressesStack.Navigator name="WalletAddressesRoot" screenOptions={defaultStackScreenOptions} initialRouteName="WalletAddresses">
+      <WalletAddressesStack.Screen name="WalletAddresses" component={WalletAddresses} options={WalletAddresses.navigationOptions(theme)} />
+    </WalletAddressesStack.Navigator>
+  );
+};
+
 const WalletExportStack = createStackNavigator();
 const WalletExportStackRoot = () => {
   const theme = useTheme();
@@ -501,6 +513,7 @@ const Navigation = () => {
       <RootStack.Screen name="ViewEditMultisigCosignersRoot" component={ViewEditMultisigCosignersRoot} options={{ headerShown: false }} />
       <RootStack.Screen name="WalletXpubRoot" component={WalletXpubStackRoot} options={{ headerShown: false }} />
       <RootStack.Screen name="SignVerifyRoot" component={SignVerifyStackRoot} options={{ headerShown: false }} />
+      <RootStack.Screen name="WalletAddressesRoot" component={WalletAddressesStackRoot} options={{ headerShown: false }} />
       <RootStack.Screen name="BuyBitcoin" component={BuyBitcoin} options={BuyBitcoin.navigationOptions(theme)} />
       <RootStack.Screen name="Marketplace" component={Marketplace} options={Marketplace.navigationOptions(theme)} />
       <RootStack.Screen name="SelectWallet" component={SelectWallet} options={{ headerLeft: null }} />
--- a/components/addresses/AddressItem.js
+++ b/components/addresses/AddressItem.js
@@ -1,0 +1,59 @@
+import React from 'react';
+import { StyleSheet } from 'react-native';
+import { useTheme } from '@react-navigation/native';
+import { ListItem } from 'react-native-elements';
+import Clipboard from '@react-native-clipboard/clipboard';
+import PropTypes from 'prop-types';
+import { AddressTypeBadge } from './AddressTypeBadge';
+import { formatBalance } from '../../loc';
+
+const AddressItem = ({ item, balanceUnit }) => {
+  const { colors } = useTheme();
+
+  const styles = StyleSheet.create({
+    container: {
+      borderBottomColor: colors.lightBorder,
+      backgroundColor: colors.elevated,
+    },
+    list: {
+      color: colors.buttonTextColor,
+    },
+  });
+
+  const shortId = `${item.address.substring(0, 9)}...${item.address.substr(item.address.length - 9)}`;
+
+  const copyAddressToClipboard = () => {
+    Clipboard.setString(item.address);
+  };
+
+  const balance = formatBalance(item.balance, balanceUnit, true);
+
+  const render = () => {
+    return (
+      <ListItem key={`${item.key}`} button onPress={copyAddressToClipboard} containerStyle={styles.container}>
+        <ListItem.Content style={styles.list}>
+          <ListItem.Title style={styles.list}>
+            {item.index}.{shortId}
+          </ListItem.Title>
+          <ListItem.Subtitle style={styles.list}>{balance}</ListItem.Subtitle>
+        </ListItem.Content>
+        <AddressTypeBadge isInternal={item.isInternal} />
+      </ListItem>
+    );
+  };
+
+  return render();
+};
+
+AddressItem.propTypes = {
+  item: PropTypes.shape({
+    key: PropTypes.string,
+    index: PropTypes.number,
+    address: PropTypes.string,
+    isInternal: PropTypes.bool,
+    transactions: PropTypes.number,
+    balance: PropTypes.number,
+  }),
+  balanceUnit: PropTypes.string,
+};
+export { AddressItem };
--- a/components/addresses/AddressItem.js
+++ b/components/addresses/AddressItem.js
@@ -20,8 +20,6 @@ const AddressItem = ({ item, balanceUnit }) => {
     },
   });
 
-  const shortId = `${item.address.substring(0, 9)}...${item.address.substr(item.address.length - 9)}`;
-
   const copyAddressToClipboard = () => {
     Clipboard.setString(item.address);
   };
@@ -32,8 +30,8 @@ const AddressItem = ({ item, balanceUnit }) => {
     return (
       <ListItem key={`${item.key}`} button onPress={copyAddressToClipboard} containerStyle={styles.container}>
         <ListItem.Content style={styles.list}>
-          <ListItem.Title style={styles.list}>
-            {item.index}.{shortId}
+          <ListItem.Title style={styles.list} numberOfLines={1} ellipsizeMode="middle">
+            {item.index}.{item.address}
           </ListItem.Title>
           <ListItem.Subtitle style={styles.list}>{balance}</ListItem.Subtitle>
         </ListItem.Content>
--- a/components/addresses/AddressTypeBadge.js
+++ b/components/addresses/AddressTypeBadge.js
@@ -1,43 +1,43 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { StyleSheet } from 'react-native';
 import { useTheme } from '@react-navigation/native';
-import { Badge } from 'react-native-elements';
+import { StyleSheet, View, Text } from 'react-native';
 import loc from '../../loc';
 
 const styles = StyleSheet.create({
-  badge: {
-    borderWidth: 0,
-    paddingVertical: 8,
-    paddingHorizontal: 6,
+  container: {
+    paddingTop: 2,
+    paddingBottom: 2,
+    paddingLeft: 8,
+    paddingRight: 8,
+    borderRadius: 20,
+  },
+  badgeText: {
+    fontSize: 12,
   },
 });
 
 const AddressTypeBadge = ({ isInternal }) => {
   const { colors } = useTheme();
 
-  const oStyles = {
-    change: {
-      badge: { backgroundColor: colors.changeBackground },
-      text: { color: colors.changeText },
-    },
-    receive: {
-      badge: { backgroundColor: colors.receiveBackground },
-      text: { color: colors.receiveText },
-    },
-  };
+  const stylesHook = StyleSheet.create({
+    changeBadge: { backgroundColor: colors.changeBackground },
+    receiveBadge: { backgroundColor: colors.receiveBackground },
+    changeText: { color: colors.changeText },
+    receiveText: { color: colors.receiveText },
+  });
 
-  const getBadgeText = () => {
-    return isInternal ? loc.addresses.type_change : loc.addresses.type_receive;
-  };
+  const badgeLabel = isInternal ? loc.addresses.type_change : loc.addresses.type_receive;
 
-  const getStyle = () => {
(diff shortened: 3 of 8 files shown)


ncoelho:
Tested, definitely some things missing. I would suggest to not make this a huge PR and effort in general, to break it down by functionality, so one feature doesn't block the other and we can keep working on it as we go. 

For instance, for first release we should aim to:
1. List of addresses
2. Retrieve new addresses
3. Incorporate sign/verify from the other PR

wdyt?

Overtorment:
i think splitting things in several PRs is a good idea, so we can deliver faster.

just

> List of addresses
> Retrieve new addresses

is good enough for initial release. @jaonoctus maybe you can deliver that in a separate PR so we can review it & merge..?

jaonoctus:
@Overtorment @ncoelho sure!

screen/wallets/addresses.js
```diff
@@ -0,0 +1,220 @@
...
+    setIsFetchingMore(true);
+    await fetchAddresses();
+    setIsFetchingMore(false);
+  };
+
+  useFocusEffect(
+    useCallback(() => {
+      Privacy.enableBlur();
+
+      const singleAddressTypes = ['legacy', 'segwitP2SH', 'segwitBech32'];
```
limpbrains, line 113: please use `LegacyWallet.type` instead of string

screen/wallets/addresses.js
```diff
@@ -0,0 +1,220 @@
...
+
+      const isWatchOnlyHD = wallet.type === 'watchOnly' && wallet._hdWalletInstance !== undefined;
```
limpbrains, line 115: watch-only wallet has `wallet.isHd()` for similar purposes

limpbrains (changes requested):
Take a look at package-lock.json it was completely rewritten. It should be removed from PR

Overall it works good, let's merge it!

components/addresses/AddressTypeBadge.js
```diff
@@ -0,0 +1,63 @@
...
+  const getStyle = () => {
+    const oStyles = {
+      change: {
+        badge: { backgroundColor: '#fdf2da' },
+        text: { color: '#f38c47' },
+      },
+      receive: {
+        badge: { backgroundColor: '#d1f9d6' },
+        text: { color: '#50c7a7' },
+      },
```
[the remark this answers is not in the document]
jaonoctus: image

:sweat_smile: will fix it

ncoelho:
Can't see the text for change/receive on iOS

Screenshot 2021-04-07 at 12 35 41

screen/wallets/addresses.js
jaonoctus: @limpbrains some wallet types return this:

image

- `wallet.type`: `HDsegwitBech32`

limpbrains: Yeah. This is for watch-only wallet only

jaonoctus:
## Light mode

image

## Dark mode

image

class/wallets/abstract-hd-electrum-wallet.js
```diff
@@ -1131,4 +1131,10 @@ export class AbstractHDElectrumWallet extends AbstractHDWallet {
   getMasterFingerprintHex() {
     return AbstractHDElectrumWallet.seedToFingerprint(this.secret);
   }
+
+  async getBalanceByAddress(address) {
+    const balance = await BlueElectrum.getBalanceByAddress(address);
```
Overtorment, line 1136: wrong. user won't expect any network work in hierarchy navigator screen.
balance for each address is in
```
    this._balances_by_external_index = {}; //  0 => { c: 0, u: 0 } // confirmed/unconfirmed
    this._balances_by_internal_index = {};
```

for  HD wallets, or simply `this.getBalance()`  for single-address wallets

screen/wallets/addresses.js
```diff
@@ -0,0 +1,226 @@
...
+      const step = 5;
+      const limit = latestIndex + step;
+
+      const receivingAddresses = [];
+      const changeAddresses = [];
+
+      const fetchAddress = async (index, type) => {
+        let address;
+
+        if (type === 'receive') {
```
Overtorment, line 76: whats this `'receive'`? looks like a magic constant, which is a bad practice

Overtorment: bool would probably be better, just like in derivation path its either 0 or 1 to identify external/internal, so `type` could be renamed to `isInternal` where false indicates receive aka external address, and true indicated change aka internal

screen/wallets/addresses.js
```diff
@@ -0,0 +1,226 @@
...
+});
+
+const WalletAddresses = () => {
+  const [showAddresses, setShowAddresses] = useState(false);
+
+  const [showLoadMoreButton, setShowLoadMoreButton] = useState(false);
+
+  const [addresses, setAddresses] = useState([]);
+
+  const [latestIndex, setLatestIndex] = useState(5);
```
Overtorment, line 42: where this 5 comes from? 
HD wallets have
```
    this.next_free_address_index
    this.next_free_change_address_index
```

so its safe to render all addresses up to those indexes.

there is also `this.gap_limit`, so you can even render all addresses up to `(this.next_free_address_index + this.gap_limit)`, but rendering anything above that will allow user to shoot himself in the foot, as BW wont be able to find transactions/balances for everything above gaplimit. this is btw how HD all wallets work, this is not specific to bluewallet.

Navigation.js
```diff
@@ -400,6 +401,17 @@ const SignVerifyStackRoot = () => {
   );
 };
 
+const WalletAddressesStack = createStackNavigator();
```
Overtorment, line 404: im really bad at this, but... @limpbrains @marcosrdz do we need another navigator for just a single screen?

screen/wallets/details.js
```diff
@@ -533,6 +541,13 @@ const WalletDetails = () => {
                 </>
               )}
 
+              {wallet.type !== LightningCustodianWallet.type && (
```
Overtorment, line 544: lets try using  `(wallet instanceof AbstractHDElectrumWallet)` here

screen/wallets/details.js
```diff
@@ -240,6 +240,14 @@ const WalletDetails = () => {
       },
     });
 
+  const navigateToAddresses = () =>
+    navigate('WalletAddressesRoot', {
+      screen: 'WalletAddresses',
+      params: {
+        walletId: wallet.getID(),
```
Overtorment, line 247: `walletId` -> `walletID`

Overtorment (changes requested):

Navigation.js
marcosrdz: We do if it is its own modal

Overtorment: so is it good here?

marcosrdz: Yup

class/wallets/abstract-hd-electrum-wallet.js
jaonoctus: @Overtorment `watchOnly` wallet does not have those functions.

```
console.log(wallet._balances_by_external_index); // undefined
console.log(wallet._balances_by_internal_index); // undefined
```

jaonoctus: nvm. 

I can use `wallet._hdWalletInstance._balances_by_external_index` instead.

screen/wallets/details.js
jaonoctus: will not work for `[LegacyWallet.type, SegwitP2SHWallet.type, SegwitBech32Wallet.type]`

Overtorment: but do they need address explorer tho? those are single-address wallets

class/wallets/abstract-hd-electrum-wallet.js
Overtorment: btw its a good idea to create a getter method for those, so internal representation is concealed from outer code.
watchOnly class already has some getters that pass-through calls to internal `wallet._hdWalletInstance` if needed

screen/wallets/details.js
jaonoctus: For upcoming features. `Sign/Verify` will be moved to this screen, for example.

That said, what about `wallet instanceof AbstractHDElectrumWallet || wallet instanceof LegacyWallet`?

Overtorment: hmm. id prefer for single-address wallets to have a separate sign/verify button on walletDetails screen.
address explorer is a feature that is specific to HD wallets, and it would require a bunch of ugly code hacks to make this screen work with single-address wallets.
anyway, don't worry about it too much, industry consensus is that single-address wallets are deprecated.

btw `wallet instanceof AbstractHDElectrumWallet || wallet instanceof LegacyWallet` makes no sense, our hierarchy is not very well thought through,  `AbstractHDElectrumWallet` is a direct descendant of `LegacyWallet`

screen/wallets/addresses.js
jaonoctus: Initially only 5 receiving (external) and 5 change (internal) addresses will be rendered.

And when the user clicks on the "load more" btn, 5 more addresses of each will be added to the list.

Its an arbitrary value, tho

components/addresses/AddressItem.js
```diff
@@ -0,0 +1,58 @@
...
+
+  const render = () => {
+    return (
+      <ListItem key={`${item.key}`} button onPress={copyAddressToClipboard} containerStyle={styles.container}>
+        <ListItem.Content style={styles.list}>
+          <ListItem.Title style={styles.list}>
+            {item.index}.{shortId}
+          </ListItem.Title>
+          <ListItem.Subtitle style={styles.list}>
+            {loc.formatString(loc.addresses.address_balance, { balance: item.balance })}
```
marcosrdz, line 37: Use formatBalance

screen/wallets/addresses.js
```diff
@@ -0,0 +1,131 @@
...
+      fetchAddresses();
+
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []),
+  );
+
+  const render = () => {
+    if (showAddresses) {
+      return <FlatList data={addresses} renderItem={item => <AddressItem {...item} />}/>;
+    }
```
marcosrdz, line 113: npm run lint:fix

marcosrdz (changes requested):

screen/wallets/addresses.js
```diff
@@ -0,0 +1,131 @@
...
+
+const fetchAddress = ({ wallet, index, isInternal }) => {
+  let address;
+  let balance = 0;
+
+  const getAllBalance = ({ c, u } = { c: 0, u: 0 }) => c + u;
+
+  if (isInternal) {
+    address = wallet._getInternalAddressByIndex(index);
+    balance = getAllBalance(wallet._balances_by_internal_index[index]);
```
Overtorment, line 19: will not work for watch-only. i sugggest making a wrapper gettet function, similar to `WatchOnlyWallet._getInternalAddressByIndex()`

screen/wallets/addresses.js
```diff
@@ -0,0 +1,131 @@
...
+import { ActivityIndicator, FlatList, StyleSheet, View } from 'react-native';
+import { useFocusEffect, useRoute } from '@react-navigation/native';
+import Privacy from '../../blue_modules/Privacy';
+import { BlueStorageContext } from '../../blue_modules/storage-context';
+import loc from '../../loc';
+import navigationStyle from '../../components/navigationStyle';
+import { AddressItem } from '../../components/addresses/AddressItem';
+import { WatchOnlyWallet } from '../../class';
+
+const fetchAddress = ({ wallet, index, isInternal }) => {
```
Overtorment, line 11: i personally not a fan of passing arguments to function as object.
traditional arguments work just fine, and any modern IDE typehints and autocomplete etc

also pls rename `fetchAddress` to smth like `getAddress`, historically we use "fetch" for something that does async network work

Overtorment (changes requested):

screen/wallets/addresses.js
jaonoctus: @Overtorment it is working, actually.

https://github.com/BlueWallet/BlueWallet/blob/f605fb04cd914627d0bd22fe0212c59e5c15dfc1/screen/wallets/addresses.js#L79

screen/wallets/addresses.js
jaonoctus: The `fetch..` name is because I was making async calls before. Forgot to rename it after the refactor

Overtorment:
running tests here https://github.com/BlueWallet/BlueWallet/pull/2954

Overtorment (approved):
tests pass. lgtm. 
@marcosrdz @limpbrains your move

components/addresses/AddressTypeBadge.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { StyleSheet } from 'react-native';
+import { useTheme } from '@react-navigation/native';
+import { Badge } from 'react-native-elements';
```
marcosrdz, line 5: Some devices fail to render the text with this RN Elements badge component. Look at electrumSettings.js for a similar implementation

screen/wallets/addresses.js
```diff
@@ -0,0 +1,131 @@
...
+
+      getAddresses();
+
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, []),
+  );
+
+  const render = () => {
+    if (showAddresses) {
+      return <FlatList data={addresses} renderItem={item => <AddressItem {...item} balanceUnit={balanceUnit} />} />;
```
marcosrdz, line 112: Add content container style with background color.

screen/wallets/addresses.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    return (
+      <View style={styles.loading}>
```
marcosrdz, line 116: Add  style with background color.

marcosrdz (changes requested):

components/addresses/AddressTypeBadge.js
jaonoctus: @marcosrdz i'm using it as it is on the coin control screen.

https://github.com/BlueWallet/BlueWallet/blob/ac0c6eed549888e5fc54f29e756d58766ecc35cf/screen/send/coinControl.js#L3

https://github.com/BlueWallet/BlueWallet/blob/ac0c6eed549888e5fc54f29e756d58766ecc35cf/screen/send/coinControl.js#L43-L50

jaonoctus: So I need to change it to something like this, right?

https://github.com/BlueWallet/BlueWallet/blob/ac0c6eed549888e5fc54f29e756d58766ecc35cf/screen/settings/electrumSettings.js#L223-L227

marcosrdz: Yup. Some OnePlus devices have custom fonts that prohibitS that Badge component from rendering reliably. A workaround is to just use a view and give it a border radius

marcosrdz: Screen Shot 2021-04-13 at 3 38 34 PM


Example

limpbrains:
@jaonoctus better way to truncate text for addresses list #2959

limpbrains (approved):
Need to start work on the next version and implement all the missing features from original issue

jaonoctus:
@limpbrains surething!

screen/wallets/addresses.js
jaonoctus: What bg color?

Overtorment:
@marcosrdz verdict..?

marcosrdz (approved):

Overtorment:
is it still WIP..?

ncoelho:
Just polishing some details and styling. Or just merge and I will do that on from master

PR #2850, opened 2021-03-25
add support for USD pegged currencies ANG AWG

pointbiz:
The Caribbean has many currencies pegged to the USD. I'm adding support specifically for two Dutch Caribbean currencies ANG and AWG. I'm promoting the use of BlueWallet heavily in Curacao and it's much easier mentally for retailers and consumers to think in terms of florins than USD.

You can see the fixed rate of 1.79 for both currencies on Wikipedia.
https://en.wikipedia.org/wiki/Netherlands_Antillean_guilder
https://en.wikipedia.org/wiki/Aruban_florin

### DIFF
--- a/ios/WalletInformationWidget/Widgets/Shared/Fiat/FiatUnit.swift
+++ b/ios/WalletInformationWidget/Widgets/Shared/Fiat/FiatUnit.swift
@@ -29,10 +29,11 @@ struct FiatUnit: Codable {
         return nil
       }
       return WidgetDataStore(rate: rateString, lastUpdate: lastUpdatedString, rateDouble: rateDouble)
-  } else if fixedRate != nil {
-      guard let bpi = json["bpi"] as? Dictionary<String, Any>, let preferredCurrency = bpi[endPointKey] as? Dictionary<String, Any>, let rateDouble = preferredCurrency["rate_float"] as? Double * Double(fixedRate), let time = json["time"] as? Dictionary<String, Any>, let lastUpdatedString = time["updatedISO"] as? String else {
+  } else if let fixedRate = fixedRate {
+      guard let bpi = json["bpi"] as? Dictionary<String, Any>, let preferredCurrency = bpi[endPointKey] as? Dictionary<String, Any>, let rateDouble = preferredCurrency["rate_float"] as? Double, let time = json["time"] as? Dictionary<String, Any>, let lastUpdatedString = time["updatedISO"] as? String else {
         return nil
       }
+      rateDouble = rateDouble * Double(fixedRate)
       return WidgetDataStore(rate: String(rateDouble), lastUpdate: lastUpdatedString, rateDouble: rateDouble)
   } else {
     guard let rateKey = rateKey, let rateDict = json[rateKey] as? [String: Any], let rateDouble = rateDict["price"] as? Double, let lastUpdated = json["timestamp"] as? Int else {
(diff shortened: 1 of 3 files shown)


marcosrdz:
Needs to be applied to FiatUnits.plist and FiatUnit.swift

pointbiz:
OK. I updated the FiatUnits.plist to make "rateKey" mandatory so there is a unique key now that pegged currencies rely on USD for the endPointKey. The dictionary structure is different in swift from JS so I had to make that change.

I updated FiatUnit.swift to put a new condition for the pegged currencies but I'm not familiar with swift and not confident in my syntax.

I also modified the lambda query to search on rateKey instead of endPointKey because rateKey is unique.

I'm nervous that the file using FiatUnit.swift for selecting a new preferred currency needs to look at rateKey instead of endPointKey but I couldn't find anywhere that needed that change.

ios/WalletInformationWidget/Widgets/Shared/Fiat/FiatUnit.swift
```diff
@@ -28,6 +29,11 @@ struct FiatUnit: Codable {
         return nil
       }
       return WidgetDataStore(rate: rateString, lastUpdate: lastUpdatedString, rateDouble: rateDouble)
+  } else if fixedRate != nil {
```
marcosrdz, line 32: Use let fixedRate = fixedRate

ios/WalletInformationWidget/Widgets/Shared/Fiat/FiatUnit.swift
```diff
@@ -28,6 +29,11 @@ struct FiatUnit: Codable {
...
+      guard let bpi = json["bpi"] as? Dictionary<String, Any>, let preferredCurrency = bpi[endPointKey] as? Dictionary<String, Any>, let rateDouble = preferredCurrency["rate_float"] as? Double * Double(fixedRate), let time = json["time"] as? Dictionary<String, Any>, let lastUpdatedString = time["updatedISO"] as? String else {
+        return nil
+      }
+      return WidgetDataStore(rate: String(rateDouble), lastUpdate: lastUpdatedString, rateDouble: rateDouble)
   } else {
     guard let rateKey = rateKey, let rateDict = json[rateKey] as? [String: Any], let rateDouble = rateDict["price"] as? Double, let lastUpdated = json["timestamp"] as? Int else {
```
marcosrdz, line 38: Compiler throwing error here

marcosrdz (changes requested):

ios/WalletInformationWidget/Widgets/Shared/Fiat/FiatUnit.swift
pointbiz: OK. I see let returns nil or not nil. I've made the change as you specified.

ios/WalletInformationWidget/Widgets/Shared/Fiat/FiatUnit.swift
pointbiz: OK. I misunderstood guard. I see all the statements must return booleans. I've fixed it.

limpbrains:
@pointbiz 
Sorry, I don't see why you need to add extra logic for these currencies. Both ANG and AWG are supported by coindesk API. You can just add them to the list and that is all.

limpbrains:
I've added ANG and AWG in #2857

pointbiz:
> I've added ANG and AWG in #2857

I didn't realize Coindesk supported those 2 currencies. So, the extra logic doesn't make sense now that I know that. 

@limpbrains  Thank you! Please post your donation address, so I can thank you that way too. 

I'll close this PR soon.

limpbrains:
> @limpbrains Thank you! Please post your donation address, so I can thank you that way too.
> 
`bc1q7xpn77t9l0c2tgjfsq82kv3y5g3569s6j5p8zc` thanks !

PR #2851, opened 2021-03-25
Ref import procedure

Overtorment:

PR #2963, opened 2021-04-15
REF: refresh wallet if it never been done

limpbrains:
Since #2851 it is possible to have freshly imported wallet with 0 balance and no transactions.
We want to automatically start refreshing when the user opens this wallet. 
So instead of checking for positive balance, lest check wallet last tx fetch time

### DIFF
--- a/screen/wallets/transactions.js
+++ b/screen/wallets/transactions.js
@@ -147,10 +147,9 @@ const WalletTransactions = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [walletID]);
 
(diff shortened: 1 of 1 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/ygnuabe9n1cb3rvm6cbjhrwvmc

[download apk](http://filestorage.bluewallet.io:1488/limpbrains-refresh-tx-1618475068.apk)

Overtorment (approved):
okay. why not.
we can always rollback.

marcosrdz (approved):

GladosBlueWallet:

PR #3027, opened 2021-04-27
Electrum cache

Overtorment:
in this PR we introduce a long-overdue electrum cache for `multiGetTransactionByTxid` call.
only transactions with >7 confirmations are cached.
this should greatly improve network fetch for batched transactions: sometimes exchange withdrawals have _thousands_ of inputs, and our code has to fetch each one of them to correctly calculate input amounts and thus, transaction value. they are now cached and repeated wallet refresh will take less time (in my case for my wallet, from 100 sec to 3 sec).
cache resides in a separate realm database file.

### DIFF
--- a/BlueComponents.js
+++ b/BlueComponents.js
@@ -44,7 +44,6 @@ import Lnurl from './class/lnurl';
 import { BlueStorageContext } from './blue_modules/storage-context';
 import ToolTipMenu from './components/TooltipMenu';
 
-/** @type {AppStorage} */
 const { height, width } = Dimensions.get('window');
 const aspectRatio = height / width;
 let isIpad;
--- a/UnlockWith.js
+++ b/UnlockWith.js
@@ -7,7 +7,6 @@ import { SafeAreaView } from 'react-native-safe-area-context';
 import { StackActions, useNavigation, useRoute } from '@react-navigation/native';
 import { BlueStorageContext } from './blue_modules/storage-context';
 import ReactNativeHapticFeedback from 'react-native-haptic-feedback';
-/** @type {AppStorage} */
 
 const styles = StyleSheet.create({
   root: {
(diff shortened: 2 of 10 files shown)


limpbrains (commented):
What about cache expiration? Without it this can cause problems

Overtorment:
> What about cache expiration? Without it this can cause problems

so the idea is that we cache only responses to "give me expanded transaction by this txid", which frankly should never change since it is written to the blockchain
we do NOT cache if such tx is <7 confirmations, so you could say we will handle well reorgs up to 7 blocks, which again should never happen on bitcoin blockchain (and if they do - we will probably have bigger problems to worry about, like nuclear war or alien invasion)

GladosBlueWallet:

 [android in browser] https://appetize.io/app/z65438e4j85ba157rrwubv5je0

[download apk](http://filestorage.bluewallet.io:1488/electrum-cache-1619552209.apk)

GladosBlueWallet:

 [android in browser] https://appetize.io/app/e80kygwja4yvh5t6tyd8nhh1r4

[download apk](http://filestorage.bluewallet.io:1488/electrum-cache-1619553100.apk)

marcosrdz (approved):

GladosBlueWallet:

 [android in browser] https://appetize.io/app/cqjtyaunnrhpan59d32zq0y42g?device=pixel4

[download apk](http://filestorage.bluewallet.io:1488/electrum-cache-1619566766.apk)

GladosBlueWallet:

limpbrains:
I mean cache requires storage space. It could be limited. Cache something forever is a not very good idea. Even if wallet will be deleted, cache remains.

Overtorment:
i haven't thought about limited space. 
hmm. on the first sight space consumption is not that big, with some heavy usage it might be up to 10-20 mb. maybe.

PR #3068, opened 2021-05-03
Added Problem need to fix of current version, the goal, and To do list

wuyih9:
Hi,

I have added "Problem need to fix", "Requirement", "Goal" and "To do list" to the "Motivation to build it" section on README.md. I think it could be more specific and more clear for the new contributors to understand this project.


jaonoctus (commented):
I think you should create an issue instead

Overtorment:
thanks, but its easier to just remove that section

PR #3168, opened 2021-05-17
REF: Hierarchy navigator screen

jaonoctus:
#990 

## UI/UX Changes

image

## SignVerify directly from an address

image

### DIFF
--- a/components/addresses/AddressItem.js
+++ b/components/addresses/AddressItem.js
@@ -115,12 +115,14 @@ const AddressItem = ({ item, balanceUnit, walletID, isWatchOnly }) => {
             </ListItem.Title>
             <View style={styles.subtitle}>
               <Text style={[stylesHook.list, styles.balance, stylesHook.balance]}>{balance}</Text>
-              <Text style={[stylesHook.list, styles.balance, stylesHook.balance]}>
-                {loc.addresses.transactions}: {item.transactions}
-              </Text>
             </View>
           </ListItem.Content>
-          <AddressTypeBadge isInternal={item.isInternal} hasTransactions={hasTransactions} />
+          <View style={styles.labels}>
+            <AddressTypeBadge isInternal={item.isInternal} hasTransactions={hasTransactions} />
+            <Text style={[stylesHook.list, styles.balance, stylesHook.balance]}>
+              {loc.addresses.transactions}: {item.transactions}
+            </Text>
+          </View>
         </ListItem>
       </View>
     );
@@ -135,7 +137,7 @@ const styles = StyleSheet.create({
     marginHorizontal: 40,
   },
   list: {
-    flex: 6,
+    flex: 8,
   },
   index: {
     fontSize: 15,
--- a/components/addresses/AddressItem.js
+++ b/components/addresses/AddressItem.js
@@ -76,7 +76,7 @@ const AddressItem = ({ item, balanceUnit, onPress }) => {
           containerStyle={stylesHook.container}
           onLongPress={showToolTipMenu}
         >
-          <ListItem.Content style={stylesHook.list}>
+          <ListItem.Content style={[styles.list, stylesHook.list]}>
             <ListItem.Title style={stylesHook.list} numberOfLines={1} ellipsizeMode="middle">
               <Text style={[styles.index, stylesHook.index]}>{item.index + 1}</Text>{' '}
               <Text style={[stylesHook.address, styles.address]}>{item.address}</Text>
@@ -101,6 +101,9 @@ const styles = StyleSheet.create({
   address: {
     fontWeight: 'bold',
     marginHorizontal: 40,
+  },
+  list: {
+    flex: 6,
   },
   index: {
     fontSize: 15,
--- a/components/addresses/AddressItem.js
+++ b/components/addresses/AddressItem.js
@@ -76,7 +76,12 @@ const AddressItem = ({ item, balanceUnit, onPress }) => {
               <Text style={[styles.index, stylesHook.index]}>{item.index + 1}</Text>{' '}
               <Text style={[stylesHook.address, styles.address]}>{item.address}</Text>
             </ListItem.Title>
-            <ListItem.Subtitle style={[stylesHook.list, styles.balance, stylesHook.balance]}>{balance}</ListItem.Subtitle>
+            <View style={styles.subtitle}>
+              <Text style={[stylesHook.list, styles.balance, stylesHook.balance]}>{balance}</Text>
+              <Text style={[stylesHook.list, styles.balance, stylesHook.balance]}>
+                {loc.addresses.transactions}: {item.transactions}
+              </Text>
+            </View>
           </ListItem.Content>
           <AddressTypeBadge isInternal={item.isInternal} />
         </ListItem>
@@ -98,6 +103,12 @@ const styles = StyleSheet.create({
   balance: {
     marginTop: 8,
     marginLeft: 14,
+  },
+  subtitle: {
+    flex: 1,
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+    width: '100%',
   },
 });
 
--- a/components/addresses/AddressItem.js
+++ b/components/addresses/AddressItem.js
@@ -9,7 +9,7 @@ import TooltipMenu from '../TooltipMenu';
 import Clipboard from '@react-native-clipboard/clipboard';
 import Share from 'react-native-share';
 
-const AddressItem = ({ item, balanceUnit, walletID }) => {
+const AddressItem = ({ item, balanceUnit, walletID, isWatchOnly }) => {
   const { colors } = useTheme();
   const tooltip = useRef();
   const listItem = useRef();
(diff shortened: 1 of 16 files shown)


components/addresses/AddressItem.js
```diff
@@ -89,16 +131,25 @@ const AddressItem = ({ item, balanceUnit, onPress }) => {
 
 const styles = StyleSheet.create({
   address: {
-    fontWeight: '600',
+    fontWeight: 'bold',
     marginHorizontal: 40,
   },
+  list: {
+    flex: 6,
```
marcosrdz, line 138: 6? How does that behave on large screens?

jaonoctus: I have not tested on larger devices

ncoelho:
Changed the labels and transaction to grow when needed.

Screenshot 2021-05-18 at 15 06 30

ncoelho (approved):

limpbrains (approved):

components/addresses/AddressItem.js
```diff
@@ -44,41 +71,58 @@ const AddressItem = ({ item, balanceUnit, onPress }) => {
...
+        onPress: handleCopyPress,
+      },
+      {
+        id: 'share',
+        text: loc.receive.details_share,
+        onPress: handleSharePress,
+      },
+    ];
+
+    if (!isWatchOnly) {
```
Overtorment, line 88: we have `wallet.allowSignVerifyMessage()` for that

Overtorment: cc @limpbrains

components/addresses/AddressItem.js
```diff
@@ -44,41 +71,58 @@ const AddressItem = ({ item, balanceUnit, onPress }) => {
     Share.open({ message: item.address }).catch(error => console.log(error));
   };
 
+  const getAvailableActions = () => {
+    const actions = [
+      {
+        id: 'copyToClipboard',
```
Overtorment, line 77: tbh, "copy" & "share" makes no sense to me here since user can just tap address and go to receive screen where he can do same actions

components/addresses/AddressItem.js
```diff
@@ -44,41 +71,58 @@ const AddressItem = ({ item, balanceUnit, onPress }) => {
...
           button
-          onPress={onPress}
+          onPress={navigateToReceive}
           containerStyle={stylesHook.container}
           onLongPress={showToolTipMenu}
         >
-          <ListItem.Content style={stylesHook.list}>
+          <ListItem.Content style={[styles.list, stylesHook.list]}>
             <ListItem.Title style={stylesHook.list} numberOfLines={1} ellipsizeMode="middle">
               <Text style={[styles.index, stylesHook.index]}>{item.index + 1}</Text>{' '}
```
Overtorment, line 113: lets start from zero? makes more sense, since we should eventually allow exporting path for addresses, like `m/48'/0'/0'/2'/0/0` and they always start with zero.
that would be consistent

components/addresses/AddressItem.js
```diff
@@ -44,41 +71,58 @@ const AddressItem = ({ item, balanceUnit, onPress }) => {
...
               <Text style={[stylesHook.address, styles.address]}>{item.address}</Text>
             </ListItem.Title>
-            <ListItem.Subtitle style={[stylesHook.list, styles.balance, stylesHook.balance]}>{balance}</ListItem.Subtitle>
+            <View style={styles.subtitle}>
+              <Text style={[stylesHook.list, styles.balance, stylesHook.balance]}>{balance}</Text>
```
Overtorment, line 117: do we have to always display it, even if its `0 BTC`? its kind of tiresome to scroll through a bunch of addresses and they all show zero btc.

cc @ncoelho

components/addresses/AddressItem.js
```diff
@@ -44,41 +71,58 @@ const AddressItem = ({ item, balanceUnit, onPress }) => {
...
+            </View>
           </ListItem.Content>
-          <AddressTypeBadge isInternal={item.isInternal} />
+          <View style={styles.labels}>
+            <AddressTypeBadge isInternal={item.isInternal} hasTransactions={hasTransactions} />
+            <Text style={[stylesHook.list, styles.balance, stylesHook.balance]}>
```
Overtorment, line 122: same for transactions. do we have to show this text if address has 0 txs..?
cc @ncoelho

Overtorment (changes requested):
mostly lgtm, but UI nuances should be discussed with @ncoelho

components/addresses/AddressItem.js
jaonoctus: Agree. The original PR was starting with zero.

components/addresses/AddressItem.js
marcosrdz: Those are shortcuts already in production. Doesnt require the receive modal to load.

jaonoctus: Ideally these actions would be in a bottom modal similar to coinControl screen, right?

marcosrdz: No. They use the standard tool tip to provide shortcuts when long pressing a row. 
I use other apps to track addresses, so being able to long press, copy, switch, rinse & repeat quickly is a time saver.

components/addresses/AddressItem.js
```diff
@@ -89,16 +133,25 @@ const AddressItem = ({ item, balanceUnit, onPress }) => {
 
 const styles = StyleSheet.create({
   address: {
-    fontWeight: '600',
+    fontWeight: 'bold',
     marginHorizontal: 40,
   },
+  list: {
+    flex: 8,
```
marcosrdz, line 140: Still confused as to why such a large number. Usually we use 1.

components/addresses/AddressItem.js
ncoelho: If more features are added that need interaction (like signing, private key, etc) we can then change to a modal. For now I think this is ok and shortcuts behaviour never hurts.

components/addresses/AddressItem.js
ncoelho: Yes, otherwise the layout changes between the rows, which is more annoying.

components/addresses/AddressTypeBadge.js
```diff
@@ -9,27 +9,46 @@ const styles = StyleSheet.create({
...
     receiveBadge: { backgroundColor: colors.receiveBackground },
+    usedBadge: { backgroundColor: colors.buttonDisabledBackgroundColor },
     changeText: { color: colors.changeText },
     receiveText: { color: colors.receiveText },
+    usedText: { color: colors.alternativeTextColor },
   });
 
-  const badgeLabel = isInternal ? loc.addresses.type_change : loc.addresses.type_receive;
+  // eslint-disable-next-line prettier/prettier
+  const badgeLabel = hasTransactions
```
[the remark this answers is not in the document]
ncoelho: Yes. Doesn't feel great there. There's some solutions:
1. Don't use the "used" label but keep the different title colors
2. Don't show "used" on change
3. Have a section title
4. Use tabs

For some reason option 2 makes more sense to me.

components/addresses/AddressItem.js
ncoelho: I think it is ok, we can revisit later

marcosrdz (approved):

components/addresses/AddressTypeBadge.js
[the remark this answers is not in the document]
jaonoctus: @ncoelho I like the "tabs" solution better. Do you have any design in mind?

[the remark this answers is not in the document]
ncoelho: Something like this, but probably means adding a new lib?

Screenshot 2021-05-26 at 17 49 12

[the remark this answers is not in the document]
jaonoctus: @ncoelho cool!

Not needed. The component could have a `currentTab` state, which will be `external` as default.

Once you click on the other tab, the state will toggle/switch the state to `internal`.

And the List will reactively filter and show the necessary addresses.

wdyt? I can send a new commit with these changes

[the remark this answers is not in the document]
ncoelho: let's try it, It does simplify it I think

components/addresses/AddressItem.js
jaonoctus: @marcosrdz why you have changed it to `item.index + 1` ?

marcosrdz: List would start at 0. Instead of 1. Programmer friendly, instead of user friendly.

components/addresses/AddressTypeBadge.js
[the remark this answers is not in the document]
jaonoctus: cc @ncoelho just need to add your :sparkles: BEAUTIFUL styles :sparkles: now.

3e987df

image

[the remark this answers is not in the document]
marcosrdz: https://reactnativeelements.com/docs/button_group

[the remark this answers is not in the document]
jaonoctus: @marcosrdz  Already working after 9493c54

Overtorment:
is it ready for review & merge?

Overtorment (approved):

jaonoctus:
@Overtorment yessir

Issue #3177, opened 2021-05-18
Storage issues


Overtorment:
chasing a bug that potentially causes storage to drop.
potential leads:

https://sentry.io/organizations/bluewallet/issues/2215853608/events/679ed7a6dcab401eaee2f6ec117566b6/?project=1295736
![image](https://user-images.githubusercontent.com/1913337/118708797-153b5a00-b814-11eb-977b-3df5649e41fb.png)

https://sentry.io/organizations/bluewallet/issues/1965998088/events/5a8b3c7fd4cc4823a2e3faa8d10b7bef/?project=1295736&query=+app.device%3A472f8cb5adc48c9c74ef4853bfdc73d6edc2aed3&statsPeriod=14d

![image](https://user-images.githubusercontent.com/1913337/118708853-271cfd00-b814-11eb-84e6-ff55e5fd9661.png)

all user issues:  https://sentry.io/organizations/bluewallet/issues/?project=1295736&query=+app.device%3A472f8cb5adc48c9c74ef4853bfdc73d6edc2aed3&statsPeriod=14d

PR #3178, opened 2021-05-18
REF: storage improvements

Overtorment:
long story short, we have this issue with random storage drop: https://github.com/BlueWallet/BlueWallet/issues/3177

in attempt to fix it, we will try to use less secure keystore and more asyncstorage.

we have this AppStorage singleton, which was supposed to handle all data persisnce when bw was created, but in retrospective it was a bad idea:
there were no advantages in keeping data management centralized (and all we do is simple key-value anyway), and all surrounding code turned into legacy mess.
probably a better approach is to let each piece of functionality (like a separate module like currency) to handle its data on its own.

so in this PR we are trying to achieve exactly that: 

1) some keys constants moved where they make more sense 

2) three storage keys that used to be stored in keystore migrated to asyncstorage (AppStorage.HANDOFF_STORAGE_KEY, AppStorage.DO_NOT_TRACK, AppStorage.ADVANCED_MODE_ENABLED). i wanted to add biometrics as well, but refrained due to minor security concerns.

3) experimental change in `Biometric.isBiometricUseEnabled` - @marcosrdz claims that because of that it will be possible to make app stuck in forever locked state (how, btw? disable biometrics in OS settings and then try relogin in BW..?)

PS. further refactoring of storage would be nice, but atm I'm out of ideas how to make it better.

### DIFF
--- a/BlueApp.js
+++ b/BlueApp.js
@@ -15,6 +15,7 @@ const startAndDecrypt = async retry => {
     console.log('App already has some wallets, so we are in already started state, exiting startAndDecrypt');
     return true;
   }
+  await BlueApp.migrateKeys();
   let password = false;
   if (await BlueApp.storageIsEncrypted()) {
     do {
--- a/blue_modules/BlueElectrum.js
+++ b/blue_modules/BlueElectrum.js
@@ -1,7 +1,7 @@
 /* global alert */
 import AsyncStorage from '@react-native-async-storage/async-storage';
 import { Alert } from 'react-native';
-import { AppStorage, LegacyWallet, SegwitBech32Wallet, SegwitP2SHWallet } from '../class';
+import { LegacyWallet, SegwitBech32Wallet, SegwitP2SHWallet } from '../class';
 import DefaultPreference from 'react-native-default-preference';
 import RNWidgetCenter from 'react-native-widget-center';
 import loc from '../loc';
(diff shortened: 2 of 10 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/r8cvq8hkn8jbzx22h7yj35kag0?device=pixel4

[download apk](http://filestorage.bluewallet.io:1488/ref-storage-1621371719.apk)

limpbrains:
So this is not about "all my wallets suddenly disappeared" bug ?

Overtorment:
> So this is not about "all my wallets suddenly disappeared" bug ?

it is

limpbrains:
> > So this is not about "all my wallets suddenly disappeared" bug ?
> 
> it is

But app storage still use RNSecureKeyStore for wallets. How does this patch affects wallets storage? I don't get it.

Overtorment:
> > > So this is not about "all my wallets suddenly disappeared" bug ?
> > 
> > 
> > it is
> 
> But app storage still use RNSecureKeyStore for wallets. How does this patch affects wallets storage? I don't get it.

affected user had keystore exceptions `{"message":"error saving key"}`, and for some reason with the same timestamp, trying to write to both handoff and biometrics key (see https://github.com/BlueWallet/BlueWallet/issues/3177) so working theory is that its some kind of race condition. maybe simultaneous write to keystore fucks it up, so the idea is to store less stuff in there. less reads, less writes. wallets data should still go to keystore, although maybe we should change rn package that provides it eventually.
plus, I tried to do a lil bit of refactoring

limpbrains (approved):

Overtorment:
@marcosrdz did you test your test-case where you think it might result in user being stuck on lock-screen..?
we need your blessing

marcosrdz (approved):
didnt get locked.

GladosBlueWallet:

PR #3213, opened 2021-05-23
REF: storage write

Overtorment:
even tho i cant reproduce race condition on keystore write, I think its a good idea to add protection against it

### DIFF
--- a/class/app-storage.js
+++ b/class/app-storage.js
@@ -1,6 +1,7 @@
 /* global alert */
 import AsyncStorage from '@react-native-async-storage/async-storage';
 import RNSecureKeyStore, { ACCESSIBLE } from 'react-native-secure-key-store';
+import * as Keychain from 'react-native-keychain';
 import {
   HDLegacyBreadwalletWallet,
   HDSegwitP2SHWallet,
@@ -20,10 +21,12 @@ import {
   SLIP39LegacyP2PKHWallet,
   SLIP39SegwitBech32Wallet,
 } from './';
+import { randomBytes } from './rng';
 const encryption = require('../blue_modules/encryption');
 const Realm = require('realm');
 const createHash = require('create-hash');
 let usedBucketNum = false;
+let savingInProgress = 0; // its both a flag and a counter of attempts to write to disk
 
 export class AppStorage {
   static FLAG_ENCRYPTED = 'data_encrypted';
@@ -88,11 +91,36 @@ export class AppStorage {
     }
   };
 
+  /**
+   * @throws Error
+   * @param key {string}
+   * @returns {Promise<*>|null}
+   */
+  getItemWithFallbackToRealm = async key => {
+    let value;
+    try {
+      return await this.getItem(key);
+    } catch (error) {
+      console.warn('error reading', key, error.message);
+      console.warn('fallback to realm');
+      const realmKeyValue = await this.openRealmKeyValue();
+      const obj = realmKeyValue.objectForPrimaryKey('KeyValue', key); // search for a realm object with a primary key
+      value = obj?.value;
+      realmKeyValue.close();
+      if (value) {
+        console.warn('successfully recovered', value.length, 'bytes from realm for key', key);
+        return value;
+      }
+      return null;
+    }
+  };
+
   storageIsEncrypted = async () => {
     let data;
     try {
-      data = await this.getItem(AppStorage.FLAG_ENCRYPTED);
+      data = await this.getItemWithFallbackToRealm(AppStorage.FLAG_ENCRYPTED);
     } catch (error) {
+      console.warn('error reading `' + AppStorage.FLAG_ENCRYPTED + '` key:', error.message);
       return false;
     }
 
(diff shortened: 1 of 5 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/mg6j9gaecfpxen0qk3p66rkxpm?device=pixel4

[download apk](http://filestorage.bluewallet.io:1488/ref-storage3-1621784508.apk)

jaonoctus (approved):

marcosrdz (approved):

class/app-storage.js
```diff
@@ -414,6 +415,12 @@ export class AppStorage {
    * @returns {Promise} Result of storage save
    */
   async saveToDisk() {
+    if (savingInProgress) {
+      console.warn('saveToDisk is in progress, skip');
+      if (++savingInProgress > 10) alert('Critical error. Last actions were not saved');
+      return;
+    }
+    savingInProgress = 1;
```
limpbrains, line 423: I don't think it is save to skip saving if another one is in process. 
Second saving should wait for the first one to complete and then be executed.
It is theoretically possible to lose data without it. Especially on a slow device.

class/app-storage.js
```diff
@@ -468,17 +475,18 @@ export class AppStorage {
...
     }
     try {
-      return await this.setItem('data', JSON.stringify(data));
+      await this.setItem('data', JSON.stringify(data));
+      await this.setItem(AppStorage.FLAG_ENCRYPTED, this.cachedPassword ? '1' : '');
     } catch (error) {
+      console.error(error.message);
       alert(error.message);
+    } finally {
+      savingInProgress = 0;
```
limpbrains, line 489: What if exception happens outside of `try {}` ? Then `savingInProgress` will not be set to `0` and save function will be forever broken.
The whole logic of `saveToDisk` func should be inside that `try {}` so that in set `savingInProgress` to `0` in 100% cases.

limpbrains (changes requested):

Overtorment:
added fallback to realm.
now, critical data gets saved to the dedicated realm database file. this file is encrypted with the current device id https://github.com/react-native-device-info/react-native-device-info#getuniqueid
as far as I can tell this is quite secure for ios devices, and somewhat secure for android devices.

in an emergency when bw cant read keystore it reads same data from realm.

GladosBlueWallet:

 [android in browser] https://appetize.io/app/vdq96n8uyy9wu6ug1jh9kq9788?device=pixel4

[download apk](http://filestorage.bluewallet.io:1488/ref-storage3-1621859667.apk)

class/app-storage.js
```diff
@@ -414,71 +483,87 @@ export class AppStorage {
...
+      const realmkeyValue = await this.openRealmKeyValue();
+      this.saveToRealmKeyValue(realmkeyValue, 'data', JSON.stringify(data));
+      this.saveToRealmKeyValue(realmkeyValue, AppStorage.FLAG_ENCRYPTED, this.cachedPassword ? '1' : '');
+      realmkeyValue.close();
     } catch (error) {
-      alert(error.message);
+      console.error('save to disk exception:', error.message);
+      alert('save to disk exception: ' + error.message);
+    } finally {
+      savingInProgress = 0;
```
limpbrains, line 566: `savingInProgress` could never reach 10 if it's value get nullified after successful run.

limpbrains (changes requested):

class/app-storage.js
Overtorment: thats kind of the idea. ive put a check and alert as a last resort for a case when for some reason save requests piled up and can not run.

limpbrains: - there are 9 pending tasks in the queue
- first executed and `savingInProgress` nullified
- 8 new tasks created
- now there are 16 pending tasks, but `savingInProgress` == 8

Overtorment: tasks are kinda interchangeable - if you get one successful you don't need the other ones.
or how do you propose to handle it? instead of nullifying decrease the counter?

limpbrains: I mean queue is growing, but you can't see that. 
Yes decrease the counter.

Let's merge it, I will propose my solution later.
I thought of array of promises. When function starts it looks for pending promises in this array. If there are any it waits for them to resolve. Then it starts and add it's own promise to the array. And so on.

limpbrains (approved):

class/app-storage.js
Overtorment: i wanted only to avoid _concurrent_ writes, and just some kind of alerting mechanism to notify user that something is completely fucked (which is not expected, really, we don't have any places in the app where saveToDisk would be firing a lot)

GladosBlueWallet:

 [android in browser] https://appetize.io/app/wh6hey8uq44mzumwy5hf52x3xm?device=pixel4

[download apk](http://filestorage.bluewallet.io:1488/ref-storage3-1622064139.apk)

limpbrains (approved):

marcosrdz (approved):

GladosBlueWallet:

PR #3403, opened 2021-07-18
Typescript support

demerzel3:
This PR adds support for typescript files in build and tests, I also converted a very small file to typescript to test the configuration and as a proof of concept.

I've used the [official guide](https://reactnative.dev/docs/typescript) as a starting point for the configuration.

Some notes:
* we are using babel to build the ts/tsx files, which means that types are not checked at build time
* type checks are delegated to a separate `tslint` command that does nothing more than running `tsc` (the typescript compiler) on the code
* I've added `tslint` as a pre-requisite of the `npm run test` command, it should be enough to avoid merging code that fails the type checks
* I've converted only a very simple file to typescript, if you'd like something else converted as part of this PR just let me know, I've tried to do the bare minimum to make it work but happy to expand on that
* I think it'd be great to add a pre-commit hook to run linting and type checking before each commit (early feedback on errors is always preferable imho), let me know if you'd like me to include it in this PR


package-lock.json
```diff
@@ -19002,7 +19348,7 @@
       }
     },
     "scryptsy": {
-      "version": "file:blue_modules/scryptsy",
```
limpbrains, line 19005: npm incorrectly changes source for scryptsy package in package-lock, this needs to be reverted

limpbrains (changes requested):

package-lock.json
marcosrdz: Daily struggle 😥

demerzel3: interesting 🤔 I'm wondering if switching to yarn would fix this.

marcosrdz: there must be an npm solution for this. its not a deal breaker. we don't wanna keep more and more dependencies than we already have.

demerzel3: @limpbrains I pushed a fixup with the revert

BlueComponents.js
```diff
@@ -282,7 +282,7 @@ export class BlueWalletNavigationHeader extends Component {
     };
   }
 
-  handleCopyPress = _item => {
+  handleCopyPress = () => {
```
Overtorment, line 285: the point of this change is that tslint is not bitching about unused `_item` ..?

demerzel3: Thanks for pointing this out, it made me think harder about the eslint configuration. Turns out that the new configuration for `@typescript-eslint/no-unused-vars` is stricter than the one we had for `no-unused-vars` and this is why it surfaced more errors (that I tried to fix).
I pushed a couple of fixup with the configuration change to align it to the old one and a revert of the unnecessary changes.

Overtorment (changes requested):
lets try to convert at least one screen file to ts to see how it will look.

is it possible to convert to ts something from /classes/? I have a feeling that because they are too much dependent on each other, its not possible to convert only one. its all or nothing..?

in general this looks pretty good, we would be able to continue write js where we want, and introduce TS gradually.

@marcosrdz @limpbrains what are your thoughts?

BlueComponents.js
```diff
@@ -1841,9 +1841,9 @@ export class DynamicQRCode extends Component {
         index: state.index - 1,
       }));
     } else {
-      this.setState(state => ({
+      this.setState({
         index: total - 1,
-      }));
+      });
```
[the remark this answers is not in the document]
demerzel3: This was part of the fixes for the unused vars, it's reverted now.

blue_modules/ur/index.js
```diff
@@ -9,7 +9,7 @@ import {
   CryptoPSBT,
   CryptoAccount,
   Bytes,
-} from '@keystonehq/bc-ur-registry';
+} from '@keystonehq/bc-ur-registry/dist';
```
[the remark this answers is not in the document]
demerzel3: yup as it stands this raises a number of TS errors as `bc-ur-registry` is written in typescript but with different version/configuration. We could probably make it work by copying the version and configuration from their repo, but I'd avoid introducing this kind of constraint honestly, using the built JS version under `dist` is easier.

class/lnurl.js
```diff
@@ -162,7 +162,7 @@ export default class Lnurl {
       fixed: min === max,
       min,
       max,
-      domain: data.callback.match(new RegExp('https://([^/]+)/'))[1],
+      domain: data.callback.match(/https:\/\/([^/]+)\//)[1],
```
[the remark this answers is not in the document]
demerzel3: this is to make this eslint rule happy: https://eslint.org/docs/rules/prefer-regex-literals.
It's probably part of the `"plugin:@typescript-eslint/recommended"` preset. I can disable the rule explicitly and revert this change if you prefer, I don't have a strong preference honestly.

components/DynamicQRCode.js
```diff
@@ -80,9 +80,9 @@ export class DynamicQRCode extends Component {
         index: state.index - 1,
       }));
     } else {
-      this.setState(state => ({
+      this.setState({
         index: total - 1,
-      }));
+      });
```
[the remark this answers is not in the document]
demerzel3: same as above, this has been reverted already.

ios/Widgets/Shared/Fiat/FiatUnit.swift
```diff
@@ -16,7 +16,7 @@ struct FiatUnit: Codable {
 }
 
 func fiatUnit(currency: String) -> FiatUnit? {
-  return Bundle.main.decode([String: FiatUnit].self, from: "fiatUnit.json").first(where: {$1.endPointKey == currency})?.value
+  return Bundle.main.decode([String: FiatUnit].self, from: "fiatUnits.json").first(where: {$1.endPointKey == currency})?.value
```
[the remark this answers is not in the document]
demerzel3: This one was interesting. Upon renaming `fiatUnit.js` to `fiatUnit.ts` I got errors all over the place because for some reason the `fiatUnit.json` file was imported in place of the typescript one. 
I figured the easiest way to fix it was to use different names, unfortunately this meant changing stuff in the swift code as well.

class/lnurl.js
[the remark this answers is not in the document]
Overtorment: its ok

limpbrains (approved):
I'm for it!

demerzel3:
@Overtorment [I tried to convert a very simple screen](https://github.com/BlueWallet/BlueWallet/pull/3403/commits/d1bd68974bc2dd87d77453a9101fa1c57d005138) (`LightningSettings`) doing the bare minimum, I had to sprinkle `any` here and there to make the compiler happy but might have already uncovered a small bug (see [this comment](https://github.com/BlueWallet/BlueWallet/pull/3403/commits/d1bd68974bc2dd87d77453a9101fa1c57d005138#diff-fde68265c7a0805a10cb8721edb909c1a346711876ddcf2fefa4cb2e0c5eb7b8R142)).

Let me know if it's ok or you'd like to go deeper. Personally I think converting at least `navigationStyle` might be a good idea, but I don't want to expand the scope too much.

I'll dedicate some time later to try and convert something from `classes` as per your suggestion.

demerzel3:
@Overtorment latest update:
* I've converted `navigationStyle` and `themes`, pretty straightforward stuff and will make screen migrations easier https://github.com/BlueWallet/BlueWallet/pull/3403/commits/4f51ac2be1c9c85f6d73e1ddf81827b24a6ef10b
* I've also converted [AbstractWallet](https://github.com/BlueWallet/BlueWallet/pull/3403/commits/79c0d83b7f580033d764c9770936f30b55823bb9) and [PlaceholderWallet](https://github.com/BlueWallet/BlueWallet/pull/3403/commits/303ffed4cc4ff1980838845f97a1be86a264e666)

None of this is perfect, there are some `any` left but we can improve it over time as more and more code gets converted.

Overtorment:
running all e2e tests here: https://github.com/BlueWallet/BlueWallet/pull/3424

PR #3451, opened 2021-07-26
REF: better handling large transactions from electrum

Overtorment:


Overtorment:
a backstory: historically we rely on electrum server to decode transactions for us (when we pass verbose=true), so sometimes for transaction that is 400kb decoded version could be way over 1Mb.
our own electrum servers configured to work with up to 5Mb transactions, but the default electrumx configuration is only 1Mb.
so I added a workaround for such case - in case we encounter this limitation we re-request transaction with verbose=false and decode it ourselves. ofcourse we disable batching for this (because with batching we could again hit this limitation). so it will work, but slower.

so in this PR we are fixing a bit rare use case when user is using his own electrum server (with low response limit), and has quite large transactions in his history.

cc @ncoelho

GladosBlueWallet:

 [android in browser] https://appetize.io/app/gdwmma6w9uqw06x6n1ay46xear?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/7431)

limpbrains (approved):
Nice! Also this is should be faster

GladosBlueWallet:

PR #3454, opened 2021-07-26
REF: convert LegacyWallet to typescript

demerzel3:
Couple notes for the reviewer:
* the `getAddress` method of `LegacyWallet` usually returns `string` but can also return `false` in some cases; adding `false` to the type exposed some issues in the rest of the code that I tried to handle case by case in a sensible manner. in case you don't like my changes or don't want to risk changing code that is working I can rollback those changes and add an ignore comment or similar to make TS happy
* renaming `legacy-wallet.js` to `legacy-wallet.ts` changed the way the file is transpiled/handled by babel and made it so that now the BlueElectrum module is loaded eagerly rather than lazily. This caused some problems in the integration tests because the client was trying unexpectedly to connect to electrum; I patched the tests copying from the ones that were using BlueElectrum directly, but I'm not sure if the change in load mode might have impacted some other areas as well. Again if you think this can cause problems I can look for a way to retain the original load mode for BlueElectrum (although I have to say it doesn't look trivial...).

### DIFF
--- a/class/wallets/legacy-wallet.ts
+++ b/class/wallets/legacy-wallet.ts
@@ -3,10 +3,17 @@ import bitcoinMessage from 'bitcoinjs-message';
 import { randomBytes } from '../rng';
 import { AbstractWallet } from './abstract-wallet';
 import { HDSegwitBech32Wallet } from '..';
-const bitcoin = require('bitcoinjs-lib');
-const BlueElectrum = require('../../blue_modules/BlueElectrum');
-const coinSelect = require('coinselect');
-const coinSelectSplit = require('coinselect/split');
+import * as bitcoin from 'bitcoinjs-lib';
+import * as BlueElectrum from '../../blue_modules/BlueElectrum';
+import coinSelect from 'coinselect';
+import coinSelectSplit from 'coinselect/split';
+import { CreateTransactionResult, CreateTransactionUtxo, Transaction, Utxo } from './types';
+
+type CoinselectUtxo = {
+  vout: number;
+  value: number;
+  txId: string;
+};
 
 /**
  *  Has private key and single address like "1ABCD....."
@@ -16,13 +23,16 @@ export class LegacyWallet extends AbstractWallet {
   static type = 'legacy';
   static typeReadable = 'Legacy (P2PKH)';
 
+  _txs_by_external_index: Transaction[] = []; // eslint-disable-line camelcase
+  _txs_by_internal_index: Transaction[] = []; // eslint-disable-line camelcase
+
   /**
    * Simple function which says that we havent tried to fetch balance
    * for a long time
    *
    * @return {boolean}
    */
-  timeToRefreshBalance() {
+  timeToRefreshBalance(): boolean {
     if (+new Date() - this._lastBalanceFetch >= 5 * 60 * 1000) {
       return true;
     }
@@ -35,7 +45,7 @@ export class LegacyWallet extends AbstractWallet {
    *
    * @return {boolean}
    */
-  timeToRefreshTransaction() {
+  timeToRefreshTransaction(): boolean {
     for (const tx of this.getTransactions()) {
       if (tx.confirmations < 7 && this._lastTxFetch < +new Date() - 5 * 60 * 1000) {
         return true;
@@ -44,12 +54,12 @@ export class LegacyWallet extends AbstractWallet {
     return false;
   }
 
-  async generate() {
+  async generate(): Promise<void> {
     const buf = await randomBytes(32);
     this.secret = bitcoin.ECPair.makeRandom({ rng: () => buf }).toWIF();
   }
 
-  async generateFromEntropy(user) {
+  async generateFromEntropy(user: Buffer): Promise<void> {
     let i = 0;
     do {
       i += 1;
@@ -68,7 +78,7 @@ export class LegacyWallet extends AbstractWallet {
    *
    * @returns {string}
    */
-  getAddress() {
+  getAddress(): string | false {
     if (this._address) return this._address;
     let address;
     try {
@@ -79,7 +89,7 @@ export class LegacyWallet extends AbstractWallet {
     } catch (err) {
       return false;
     }
-    this._address = address;
+    this._address = address ?? false;
 
     return this._address;
   }
@@ -87,8 +97,10 @@ export class LegacyWallet extends AbstractWallet {
   /**
    * @inheritDoc
    */
-  getAllExternalAddresses() {
-    return [this.getAddress()];
+  getAllExternalAddresses(): string[] {
+    const address = this.getAddress();
+
+    return address ? [address] : [];
   }
 
   /**
@@ -97,9 +109,11 @@ export class LegacyWallet extends AbstractWallet {
    *
    * @returns {Promise.<void>}
    */
-  async fetchBalance() {
+  async fetchBalance(): Promise<void> {
     try {
-      const balance = await BlueElectrum.getBalanceByAddress(this.getAddress());
+      const address = this.getAddress();
+      if (!address) throw new Error('LegacyWallet: Invalid address');
+      const balance = await BlueElectrum.getBalanceByAddress(address);
       this.balance = Number(balance.confirmed);
       this.unconfirmed_balance = Number(balance.unconfirmed);
       this._lastBalanceFetch = +new Date();
@@ -113,9 +127,11 @@ export class LegacyWallet extends AbstractWallet {
    *
    * @return {Promise.<void>}
    */
-  async fetchUtxo() {
+  async fetchUtxo(): Promise<void> {
     try {
-      const utxos = await BlueElectrum.multiGetUtxoByAddress([this.getAddress()]);
+      const address = this.getAddress();
+      if (!address) throw new Error('LegacyWallet: Invalid address');
+      const utxos = await BlueElectrum.multiGetUtxoByAddress([address]);
       this.utxo = [];
       for (const arr of Object.values(utxos)) {
         this.utxo = this.utxo.concat(arr);
@@ -156,8 +172,8 @@ export class LegacyWallet extends AbstractWallet {
    * @param respectFrozen {boolean} Add Frozen outputs
    * @returns {[]}
    */
-  getUtxo(respectFrozen = false) {
-    let ret = [];
+  getUtxo(respectFrozen = false): Utxo[] {
+    let ret: Utxo[] = [];
     for (const u of this.utxo) {
       if (u.txId) u.txid = u.txId;
       if (!u.confirmations && u.height) u.confirmations = BlueElectrum.estimateCurrentBlockheight() - u.height;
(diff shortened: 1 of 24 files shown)


Overtorment:
id say dont rush converting everything to TS.
"f it works, don't touch it"

demerzel3:
This is a way of filling the gaps that I left in the last PR as regard to `AbstractWallet`: some of the fields have underspecified types and some are just plain wrong (see `utxo` for example). Besides, I think that if we want to switch to typescript starting from the models would serve us well as an approach, so I see this as valuable work overall. It's also a great way for me to get familiar with the codebase.
Of course I don't want to break anything with this, I will make sure that all existing tests still pass and I will revert any unwanted change in behavior that I may have introduced with the types (as soon as I have a bit of time to continue working on it 😄)

demerzel3:
@Overtorment I think this is ready for review now if you want to take a look, I've updated the main comment with a couple of notes.

GladosBlueWallet:

image

limpbrains (approved):
I very much like it!

I believe we need to refactor BlueElectrum so that it will have manual connect() method, that we can run in tests if it is needed.

tests/integration/watch-only-wallet.test.js
```diff
@@ -1,8 +1,7 @@
+import assert from 'assert';
+
 import { WatchOnlyWallet } from '../../class';
-const assert = require('assert');
-global.net = require('net'); // needed by Electrum client. For RN it is proviced in shim.js
-global.tls = require('tls'); // needed by Electrum client. For RN it is proviced in shim.js
-const BlueElectrum = require('../../blue_modules/BlueElectrum'); // so it connects ASAP
+import * as BlueElectrum from '../../blue_modules/BlueElectrum';
```
Overtorment, line 4: there might be a problem with this approach, as `require` caches the result,
so if module that's being required creates a connection (and also some variables in its scope),  this connection is reused every time someone requires this module.
this is how we ensured.

with `import` its a different deal afaik. everything is brand new and fresh when you import a module, so sockets and connections are duplicated.

not sure how it all works in this PR, need to finish reading the diff and try running it myself

tests/integration/notifications.test.js
```diff
@@ -1,6 +1,8 @@
+import assert from 'assert';
```
Overtorment, line 1: whats the purpose of changing `require` -> `import` ?

tests/integration/lightning-custodian-wallet.test.js
```diff
@@ -1,6 +1,24 @@
+import assert from 'assert';
 import Frisbee from 'frisbee';
+
 import { LightningCustodianWallet } from '../../class';
-const assert = require('assert');
+import * as BlueElectrum from '../../blue_modules/BlueElectrum';
```
Overtorment, line 5: lightning custodial wallet doesnt need `BlueElectrum`, it works through REST API

tests/setup.js
```diff
@@ -2,6 +2,9 @@
 
 import mockClipboard from '@react-native-clipboard/clipboard/jest/clipboard-mock.js';
 
+global.net = require('net'); // needed by Electrum client. For RN it is proviced in shim.js
```
Overtorment, line 5: good idea to move it here

tests/integration/hodlhodl.test.js
```diff
@@ -1,8 +1,20 @@
+import assert from 'assert';
+import * as bitcoin from 'bitcoinjs-lib';
+
 import { LegacyWallet, SegwitBech32Wallet, SegwitP2SHWallet } from '../../class';
 import { HodlHodlApi } from '../../class/hodl-hodl-api';
+import * as BlueElectrum from '../../blue_modules/BlueElectrum';
```
Overtorment, line 6: hodlhodl doesnt need `BlueElectrum`

tests/integration/aopp.test.js
```diff
@@ -2,6 +2,18 @@ import assert from 'assert';
 
 import { HDLegacyP2PKHWallet, HDSegwitP2SHWallet } from '../../class';
 import AOPP from '../../class/aopp';
+import * as BlueElectrum from '../../blue_modules/BlueElectrum'; // so it connects ASAP
```
Overtorment, line 5: AOPP doesnt need `BlueElectrum`

Overtorment (changes requested):
ok i see your troubles with BlueElectrum.
the easiest solution would be (I just tested it, it works) to remove mandatory connection from BlueElectrum.js (basically socket connects as soon as module is required, lets change that),
and add connection to BlueApp.js right after BlueElectrum is required.

```
overtorment@laptop3 ~/Documents/BlueWallet (master) $ git diff BlueApp.js blue_modules/BlueElectrum.js
diff --git a/BlueApp.js b/BlueApp.js
index acdce2a0..4b0a1fcf 100644
--- a/BlueApp.js
+++ b/BlueApp.js
@@ -5,6 +5,7 @@ import loc from './loc';
 const prompt = require('./blue_modules/prompt');
 const currency = require('./blue_modules/currency');
 const BlueElectrum = require('./blue_modules/BlueElectrum'); // eslint-disable-line @typescript-eslint/no-unused-vars
+BlueElectrum.connectMain();
 const BlueApp = new AppStorage();
 // If attempt reaches 10, a wipe keychain option will be provided to the user.
 let unlockAttempt = 0;
diff --git a/blue_modules/BlueElectrum.js b/blue_modules/BlueElectrum.js
index 4ff7325f..599d9a57 100644
--- a/blue_modules/BlueElectrum.js
+++ b/blue_modules/BlueElectrum.js
@@ -183,8 +183,6 @@ async function connectMain() {
   }
 }
 
-connectMain();
-
 async function presentNetworkErrorAlert(usingPeer) {
   if (await isDisabled()) {
     console.log(
```


otherwise LGTM

class/wallets/legacy-wallet.ts
```diff
@@ -16,13 +23,16 @@ export class LegacyWallet extends AbstractWallet {
   static type = 'legacy';
   static typeReadable = 'Legacy (P2PKH)';
 
+  _txs_by_external_index: Transaction[] = []; // eslint-disable-line camelcase
```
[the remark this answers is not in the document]
demerzel3: Not sure what you mean here, they are used in `fetchTransactions` ([here](https://github.com/demerzel3/BlueWallet/blob/82299ebaea12010d5e26d1712b14621ecf8a1890/class/wallets/legacy-wallet.ts#L346)) and `getTransactions` ([here](https://github.com/demerzel3/BlueWallet/blob/82299ebaea12010d5e26d1712b14621ecf8a1890/class/wallets/legacy-wallet.ts#L352-L354)).

class/wallets/legacy-wallet.ts
```diff
@@ -364,14 +411,25 @@ export class LegacyWallet extends AbstractWallet {
...
+    targets: {
+      address: string;
+      value?: number;
+    }[],
+    feeRate: number,
+    changeAddress: string,
+    sequence: number,
+    skipSigning = false,
+    masterFingerprint: number,
+  ): CreateTransactionResult<U> {
```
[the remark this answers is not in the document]
demerzel3: here `U` is anything that extends `CreateTransactionUtxo`, it is a way of expressing that the list of utxos that you pass in must at least have the same fields as `CreateTransactonUtxo`, but whatever extra attributes you pass in it will be kept intact in the result.

If `CreateTransactionUtxo` is this type:
```ts
type CreateTransactionUtxo = {
  txId: string
  value: number
}
```
Then you can do something like this and typescript won't complain:
```ts
// given the shape of the input `U` is inferred here to be: 
// {txId: string; value: number; foo: string}
const result = createTransaction(
    [{ txId: '123', value: 123, foo: 'foo' }], 
    ... whatever ...
)
// Since in the result `inputs` is of type `U[]` we can read `foo` without issues
console.log(result.inputs[0].foo)
```

class/wallets/legacy-wallet.ts
```diff
@@ -392,12 +450,13 @@ export class LegacyWallet extends AbstractWallet {
...
 
-    outputs.forEach(output => {
+    const sanitizedOutputs = outputs.map(output => ({
+      ...output,
       // if output has no address - this is change output
-      if (!output.address) {
-        output.address = changeAddress;
-      }
+      address: output.address ?? changeAddress,
+    }));
```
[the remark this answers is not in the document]
demerzel3: the existing code was fine but this is better because it changes the type of `address` from being optional (`string | undefined`) to always being present (`string`), which is a tad safer and saves some checks down the line.

typings/coinselect.d.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+declare module 'coinselect/split' {
+  type Utxo = {
+    vout: number;
+    value: number;
+    txId: string;
+  };
+
+  export default function coinSelectSplit<U extends Utxo>(
```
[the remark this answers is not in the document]
demerzel3: Same as a previous comment, `U` gets inferred from the input to be anything that extends the `Utxo` type; whatever extra attribute the input has will be preserved in the result (as you can see `inputs` is of type `U[]` in the return type).

tests/integration/notifications.test.js
demerzel3: just consistency 🤷‍♂️

tests/integration/watch-only-wallet.test.js
demerzel3: This is not my experience at all, as far as I know es6 modules are only evaluated once on the first import exactly like `require`, it shouldn't create any problem.

demerzel3:
@Overtorment I moved `BlueElectrum` connection to `BlueApp` as per your suggestion and reverted the changes to the tests that didn't need `BlueElectrum` in the first place.

tests/integration/watch-only-wallet.test.js
Overtorment: nice

tests/integration/hd-legacy-breadwallet.test.js
```diff
@@ -17,7 +16,7 @@ afterAll(async () => {
 beforeAll(async () => {
   // awaiting for Electrum to be connected. For RN Electrum would naturally connect
   // while app starts up, but for tests we need to wait for it
-  await BlueElectrum.waitTillConnected();
+  await BlueElectrum.connectMain();
```
Overtorment, line 19: you sure we dont need both `connectMain` & `waitTillConnected` ?

tests/integration/import.test.js
```diff
@@ -58,7 +57,7 @@ afterAll(async () => {
 beforeAll(async () => {
   // awaiting for Electrum to be connected. For RN Electrum would naturally connect
```
Overtorment, line 58: not precisely correct comment anymore

Overtorment:
running all tests here: https://github.com/BlueWallet/BlueWallet/pull/3576

Overtorment (approved):

PR #3465, opened 2021-07-28
FIX: Remove existing release notes when applying Catalyst patches.

marcosrdz:
Builds would get release notes from old commits rather than the latest ones

### DIFF
--- a/scripts/maccatalystpatches/applypatchesformaccatalyst.sh
+++ b/scripts/maccatalystpatches/applypatchesformaccatalyst.sh
@@ -1,3 +1,5 @@
+echo "Removing existing release notes"
+rm release-notes.txt release-notes.json
 echo "Applying patch for package.json"
 sed -i '' '/react-native-tor/d' ./package.json
 rm -fr node_modules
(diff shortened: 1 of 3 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/fkbf1j8rnufy4j5z6139ueavd0?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/7478)

scripts/maccatalystpatches/applypatchesformaccatalyst.sh
```diff
@@ -1,3 +1,5 @@
+echo "Removing existing release notes"
+rm release-notes.txt release-notes.json
```
Overtorment, line 2: how is this supposed to work?
those files should regenerate upon instal

Overtorment (changes requested):

scripts/maccatalystpatches/applypatchesformaccatalyst.sh
marcosrdz: Those files dont get over written when npm install runs so one could end up deploying release note files that are inaccurate. Such as from WIP branches

Overtorment (approved):

GladosBlueWallet:

PR #3619, opened 2021-08-19
Fix 3357 txs in realm

Overtorment:
a bit risky change. we change format of storing HD wallets transactions in realm. before, we json encoded all txs in a single json blob and wrote it in a single realm row.
that started failing when some users transactions exceeded 30 mb.

after that change we store 1 tx per row, so we now practically can store unlimited amount of transactions (again HD wallets only).

after the update, users will notice transactions disappearing and then re-fetching.

also, if user updated to this build its impossible to go back, as realm doesn't allow schema downgrade.

cc @ncoelho

### DIFF
--- a/class/app-storage.js
+++ b/class/app-storage.js
@@ -228,16 +228,16 @@ export class AppStorage {
     const password = this.hashIt(this.cachedPassword || 'fyegjitkyf[eqjnc.lf');
     const buf = Buffer.from(this.hashIt(password) + this.hashIt(password), 'hex');
     const encryptionKey = Int8Array.from(buf);
-    const path = this.hashIt(this.hashIt(password)) + '-wallets.realm';
+    const path = this.hashIt(this.hashIt(password)) + '-wallettransactions.realm';
 
     const schema = [
       {
-        name: 'Wallet',
-        primaryKey: 'walletid',
+        name: 'WalletTransactions',
         properties: {
           walletid: { type: 'string', indexed: true },
-          _txs_by_external_index: 'string', // stringified json
-          _txs_by_internal_index: 'string', // stringified json
+          internal: 'bool?', // true - internal, false - external
+          index: 'int?',
+          tx: 'string', // stringified json
         },
       },
     ];
@@ -317,7 +317,12 @@ export class AppStorage {
       }
     }
     if (data !== null) {
-      const realm = await this.getRealm();
+      let realm;
+      try {
+        realm = await this.getRealm();
+      } catch (error) {
+        alert(error.message);
+      }
       data = JSON.parse(data);
       if (!data.wallets) return false;
       const wallets = data.wallets;
@@ -410,7 +415,7 @@ export class AppStorage {
             break;
         }
 
-        this.inflateWalletFromRealm(realm, unserializedWallet);
+        if (realm) this.inflateWalletFromRealm(realm, unserializedWallet);
 
         // done
         if (!this.wallets.some(wallet => wallet.getSecret() === unserializedWallet.secret)) {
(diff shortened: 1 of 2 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/tqkyvaycf3zuj9p3ze5y5m3uvg?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/7843)

class/app-storage.js
```diff
@@ -317,7 +318,12 @@ export class AppStorage {
       }
     }
     if (data !== null) {
-      const realm = await this.getRealm();
+      let realm;
+      try {
+        realm = await this.getRealm();
+      } catch (error) {
+        alert(error.message);
+      }
```
limpbrains, line 326: Why realm became optional?

class/app-storage.js
```diff
@@ -451,23 +457,27 @@ export class AppStorage {
...
-          if (walletToInflate._hdWalletInstance) {
-            walletToInflate._hdWalletInstance._txs_by_external_index = txsByExternalIndex;
-            walletToInflate._hdWalletInstance._txs_by_internal_index = txsByInternalIndex;
-          } else {
-            walletToInflate._txs_by_external_index = txsByExternalIndex;
-            walletToInflate._txs_by_internal_index = txsByInternalIndex;
-          }
+    const transactionsForWallet = wallets.filtered(`walletid = "${walletToInflate.getID()}"`);
+    for (const tx of transactionsForWallet) {
+      if (tx.tx && tx.internal === false) {
```
limpbrains, line 462: if (!tx.tx) continue;

class/app-storage.js
```diff
@@ -233,16 +233,17 @@ export class AppStorage {
...
       {
         name: 'Wallet',
-        primaryKey: 'walletid',
         properties: {
           walletid: { type: 'string', indexed: true },
-          _txs_by_external_index: 'string', // stringified json
-          _txs_by_internal_index: 'string', // stringified json
+          internal: 'bool?', // true - internal, false - external
+          index: 'int?',
+          tx: 'string?', // stringified json
```
limpbrains, line 240: Why use optional type?

class/app-storage.js
Overtorment: this realm is only for storing transactions. i noticed that if exception is raised somewhere in getting this realm (like schema version mismatch), whole block of loadFromDisk skips, and user might end up with no wallets (when he starts the app), so this is like failsafe for some unprobable occasions

class/app-storage.js
Overtorment: the most important check here is `tx.internal`, don't think we can exit fast here

class/app-storage.js
Overtorment: the idea was, that when we decide to offload to realm other wallet types (besides HD), we would need a bit different schema, so we would add other optional fields. wdyt?

GladosBlueWallet:

image

class/app-storage.js
limpbrains: I don't have experience with Realm. IMO it is better to rename 'Wallets' to 'HDWalletsTransactions' and enforce strict scheme there. So you don't need to check for tx.tx and tx.internal. You know it is there for sure.
For different wallet types another realm scheme should be used.

Overtorment: hmm makes sense. will do

GladosBlueWallet:

image

GladosBlueWallet:

image

GladosBlueWallet:

 [android in browser] https://appetize.io/app/bq71hnk7huxwbpj0efh824tpt4?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/7884)

PR #3625, opened 2021-08-20
FIX: generate release notes if running from tag checkout

### LINKED ISSUE #3145: Release notes not generated when running from tag checkout
Running `git checkout v6.1.2;` after cloning this repo and building the app, will result in empty releasenotes.json file (it will contain only `""`)

The reason is that running file `./scripts/release-notes.sh` will return nothing.
as:
`git tag | sort | tail -n 1` will return `v6.1.2`
and running `git show-ref -s v6.1.2` return `55cb3daf4d02e8236a07104b0bff75d5b0a8a62f`
and running: `git  log --pretty=format:'* %s %b'  55cb3daf4d02e8236a07104b0bff75d5b0a8a62f..HEAD`
will return empty result.

fix: ensure the release-notes.sh script works when running after checkout of a tag.

limpbrains:
fix #3145

If run release-notes.sh from master, it will show notes staring from last tag.

But if you checkout specific tag, it will show notes starting from the previous tag to the current one

cc @emanuelb

### DIFF
--- a/scripts/release-notes.sh
+++ b/scripts/release-notes.sh
@@ -1,4 +1,11 @@
 #!/bin/sh
-TAG=`git tag | sort | tail -n 1`
+HEAD=`git rev-parse --abbrev-ref --symbolic-full-name HEAD`
+if [ $HEAD == "master" ]
+then
+    TAG=`git tag | sort | tail -n 1`
+else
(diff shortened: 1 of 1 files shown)


Overtorment (approved):

GladosBlueWallet:

Issue #3476, opened 2021-07-30
Support truncated BIP39 wordlist (up to 4 chars per word)


Overtorment:
as per spec, 4 chars is more than enough (sometimes 3) to lookup correct word in the wordlist.
we need to teach our abstract HD wallet to work with it.

might be worth looking into our bip39 lib, maybe its already supported.

testcases:

![image](https://user-images.githubusercontent.com/1913337/127673355-7e0842af-3748-4ad2-ab25-ce8e8dbcfb8c.png)

Overtorment:
related https://github.com/BlueWallet/BlueWallet/issues/1285

PR #3633, opened 2021-08-21
Add support for BIP39 truncated wordlist

demerzel3:
Closes https://github.com/BlueWallet/BlueWallet/issues/3476

### DIFF
--- a/class/wallets/abstract-hd-wallet.ts
+++ b/class/wallets/abstract-hd-wallet.ts
@@ -1,6 +1,8 @@
 import { LegacyWallet } from './legacy-wallet';
-const bip39 = require('bip39');
-const BlueElectrum = require('../../blue_modules/BlueElectrum');
+import * as bip39 from 'bip39';
+import { BIP32Interface } from 'bip32';
+import BlueElectrum from '../../blue_modules/BlueElectrum';
+import { Transaction } from './types';
 
 /**
  * @deprecated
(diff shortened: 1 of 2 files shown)


class/wallets/abstract-hd-wallet.ts
```diff
@@ -21,59 +35,94 @@ export class AbstractHDWallet extends LegacyWallet {
...
   }
 
-  setSecret(newSecret) {
+  setSecret(newSecret: string): this {
     this.secret = newSecret.trim().toLowerCase();
     this.secret = this.secret.replace(/[^a-zA-Z0-9]/g, ' ').replace(/\s+/g, ' ');
+
+    // try to match words to the default bip39 wordlist and complete partial words
+    const wordlist = bip39.wordlists[bip39.getDefaultWordlist()];
+    // TODO: since the wordlist is static this will be always the same and can be cached indefinitely.
```
Overtorment, line 79: why cache it? its already in out modules, so access is fast & cheap

class/wallets/abstract-hd-wallet.ts
```diff
@@ -21,59 +35,94 @@ export class AbstractHDWallet extends LegacyWallet {
...
+      } else {
+        lookupMapTemp.set(prefix4, [word]);
+      }
+    }
+    const lookupMap = new Map<string, string>();
+    lookupMapTemp.forEach((words, key) => {
+      if (words.length === 1) {
+        lookupMap.set(key, words[0]);
+      }
+    });
```
Overtorment, line 104: feels overengineered

class/wallets/abstract-hd-wallet.ts
demerzel3: I was thinking about caching the `Map` so that we don't have to recompute it every time an import is attempted, but it might not matter as importing wallets is not something that is done so frequently.

Overtorment: plus its only 2000 words, and easy to calculate hashmap

Overtorment (approved):
lgtm

limpbrains (approved):

PR #3685, opened 2021-08-27
REF: QRComponent

marcosrdz:


components/QRCodeComponent.js
```diff
@@ -0,0 +1,94 @@
...
+import Share from 'react-native-share';
+import loc from '../loc';
+import PropTypes from 'prop-types';
+
+const QRCodeComponent = ({
+  value,
+  isLogoRendered = true,
+  isMenuAvailable = true,
+  logoSize = 90,
+  size = (is.ipad() && 300) || 300,
```
limpbrains, line 16: Is it always 300 ?

limpbrains (approved):

components/QRCodeComponent.js
marcosrdz: Default yes. If a different size is required then it gets passed as prop

limpbrains: I mean, if default 300, then iPad checking can be removed

GladosBlueWallet:

PR #3792, opened 2021-09-11
FIX: clarify LndHub custom URI setting format

xanoni:
- Show better example for Electrum host IP setting
- Fix example LndHub URI shown in settings
- Use dedicated format string for LndHub URI setting


limpbrains (changes requested):
You don't need a dedicated string for LndHub. Just pass different example

GladosBlueWallet:

image

xanoni:
> You don't need a dedicated string for LndHub. Just pass different example

I am aware that the content of the string is the same, but it is confusing that the string used for a non-Electrum URI (not even host!) is called `electrum_host`.  If you expect the string to always remain the same, then I would at least rename the variable to something like `uri_or_host`.

limpbrains:
Agree, naming is confusing. If you want to change this PR to rename string key, I'm for it.

Thing is changing json and js files is not enough. You will also need go Transifex and change this string in all ~60 languages by hand. Because otherwise already existed translation of this string will be lost. That is why I prefer not to change anything

GladosBlueWallet:

image

GladosBlueWallet:

image

xanoni:
Just to clarify, the reason why I originally sent this PR is that the “https://“ was missing in the placeholder LndHub string and thus it told me my custom LndHub URI was invalid.

When I tried to find the right place to add that in the code, I realized that the variable naming was confusing, thus I also changed that.

I believe I updated all the languages already? Or are you talking about something else?

limpbrains:
> I believe I updated all the languages already? Or are you talking about something else?

Yes. We are using [Transifex](transifex.com/) to store translations. It is primary source of truth. It is not enough to edit language files.

I think PR should just change example and nothing more. 
Thanks

PR #3817, opened 2021-09-14
FIX: typos in lang file

limpbrains:
cc @mcsaeid


loc/en.json
```diff
@@ -114,14 +114,14 @@
     "payButton": "Pay",
     "placeholder": "Invoice",
     "open_channel": "Open Channel",
-    "funding_amount_placeholder": "Funding amount, for exampe 0.001",
+    "funding_amount_placeholder": "Funding amount, for example 0.001",
```
mcsaeid, line 117: Great job overall. As for this, please add BTC at the end to specify what this is even though it is obvious. Also, you will need a comma after “for example,” so it would be: “Funding amount, for example, 0.001 BTC.”

GladosBlueWallet:

 [android in browser] https://appetize.io/app/5gdvva4akb0rdrjn39h4z8mgg8?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/8311)

loc/en.json
marcosrdz: The unit would be based on the wallet setting. It won't be hardcoded.

mcsaeid: I understand that. The example, however, won’t change. It will always be “0.001,” but 0.001 what? BTC, of course.

marcosrdz (approved):

Overtorment (approved):

Issue #3832, opened 2021-09-16
Feature Request: Add WalletMatrix Support


dcentrica:
## Proposing a feature?

Adds an informational meta-file about BlueWallet's features to its public GitHub repo for auto-import into [WalletMatrix](https://walletmatrix.app).

WalletMatrix is a Bitcoin wallet feature discovery tool for folks to find wallets by feature, language and platform alongside contextual help content, a [developer API](https://walletmatrix.app/developers) and heaps more. There are several similar services which rely on manual and centralised data curation, but since wallet projects iterate so quickly, WalletMatrix automates an otherwise unsustainable update process.

Merging this PR exposes BlueWallet's features via a standard [`matrix.json`](https://walletmatrix.app/schema) file which allows WalletMatrix to automatically parse, consume and store BlueWallet's features, making them available to search almost immediately.

If this is not an appropriate repository, please let us know which one is and we'll happily re-issue the feature there.

**What do I need to do?**

 1. Merge an incoming PR (feel free to comment on it first!)
 2. Keep your new `matrix.json` up-to-date :-)

**Resources**

* [The Matrix Itself](https://walletmatrix.app/matrix/)
* [The Glossary](https://walletmatrix.app/glossary)
* [About WalletMatrix](https://walletmatrix.app/about/)
* `matrix.json` is currently based on our [v3 JSON schema](https://walletmatrix.app/schema/v3.json)
* See [an example `matrix.json` file here](https://github.com/dcentrica/walletmatrix-acme-wallet)
* Need a UUID for a new wallet? Just call our [`uuid`](https://walletmatrix.app/api/v1/uuid) API
* [Wallet Vendor Info](https://walletmatrix.app/vendors/)

**Questions?**

* [FAQ](https://walletmatrix.app/faq/)
* [Vendor Info](https://walletmatrix.app/vendors/)
* [Twitter](https://twitter.com/walletmatrix)
* Email: hello AT dcentrica DOT com

PR #3833, opened 2021-09-16
ADD: Add WalletMatrix support. See #3832

dcentrica:
Adds an informational meta-file about your wallet's features for auto-import into walletmatrix.app. See #3832

### DIFF
--- a/matrix.json
+++ b/matrix.json
@@ -1,0 +1,228 @@
+{
+    "version": "3.0",
+    "organisation": {
+        "name": "BlueWallet Services S.R.L",
+        "web": "https://bluewallet.io/",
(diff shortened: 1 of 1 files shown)


ncoelho:
nACK

I don't think this should be in the app repo.

phptek:
Thanks, is there an alternative repo against which a PR could be issued?

dcentrica:
@ncoelho Do you have a suggestion for where might be best to re-issue the PR? As you may recall from almost a year ago, we did attempt to host this on a private repo run by us and invited @Overtorment and others to manage it, but were convinced that keeping things open would be the best policy - hence a PR to BlueWallet's OSS repo. Cheers :+1:

PR #3849, opened 2021-09-18
FIX: do not check secret when adding new wallet

limpbrains:
wallet.getID() is enought, because it is already contains hash of secret + passphrase

closes #3848

@mcsaeid please test it

PR #3850, opened 2021-09-19
FIX: use getID to compare wallets during delete and loadFromDisk

limpbrains:
Basically same as #3849
- If you add 2 wallets with same seed, but different passphrase, and try to remove one wallet, both wallets will be removed
- same for loading from disk. one wallet will be removed if they both have same seed

PS
We use getID() quite often in the app. And it always calculates sha256, for it. How fast is it? Should we optimize it? Switch to constant id's, for example. Of cache it.

### DIFF
--- a/class/app-storage.js
+++ b/class/app-storage.js
@@ -426,7 +426,8 @@ export class AppStorage {
         }
 
         // done
-        if (!this.wallets.some(wallet => wallet.getSecret() === unserializedWallet.secret)) {
+        const ID = unserializedWallet.getID();
+        if (!this.wallets.some(wallet => wallet.getID() === ID)) {
           this.wallets.push(unserializedWallet);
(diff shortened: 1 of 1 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/3gfrzt7dt46p7vydhqjnppm0kr?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/8394)

marcosrdz (approved):

Overtorment:
purejs sha256 is quite fast, no need to optimize it. if you're worried you can put time measurement on selftest screen and measure it I'm runtime.
you're probably thinking about privkey->address conversion, and that is slow indeed (but mostly because when we need it we do it in packs of gap_limit, so it might take seconds, single conversion is still something like 100ms)

PR #3853, opened 2021-09-19
FIX: Iranian currency

mcsaeid:
The correct Iranian currency is IRT or Iranian Toman. Exir* also shows the amount in toman instead of rial (IRR).

Since IRR is one-tenth of an IRT, the amount shown in BlueWallet is incorrect. It should be ten times more†. People rarely use IRR in daily life, anyway.

bkp09a4E jpg small

cY1HDdf2 jpg small

*‌ https://github.com/BlueWallet/BlueWallet/blob/42be5c9b56f371ee2ce9109bdab6a98ecdd08f46/models/fiatUnit.ts#L62
† The correct amount shown in the screenshot should have been ۱۵۱٬۹۷۷٬۵۰۰٫۰۰ ﷼.

**Question**
How can we fix the thousands separators? They look off. For comparison, look at the Persian numbers above. I am also using the Arabic Thousands Separator (U+066C) instead of a comma.

### DIFF
--- a/models/fiatUnits.json
+++ b/models/fiatUnits.json
@@ -119,6 +119,12 @@
       "locale": "fa-IR",
       "source": "Exir"
     },
+    "IRR": {
+      "endPointKey": "IRR",
+      "symbol": "﷼",
+      "locale": "fa-IR",
+      "source": "Exir"
+    },
     "JPY": {
       "endPointKey": "JPY",
       "symbol": "¥",
--- a/models/fiatUnits.json
+++ b/models/fiatUnits.json
@@ -113,9 +113,9 @@
       "locale": "hi-HN",
       "source": "CoinDesk"
     },
-    "IRR": {
-      "endPointKey": "IRR",
-      "symbol": "﷼",
+    "IRT": {
+      "endPointKey": "IRT",
+      "symbol": "تومان",
       "locale": "fa-IR",
       "source": "Exir"
     },
--- a/tests/integration/Currency.test.js
+++ b/tests/integration/Currency.test.js
@@ -42,9 +42,9 @@ describe('currency', () => {
     // assert.ok(cur.BTC_LBP > 0);
 
     // test Exir rate source
-    await currency.setPrefferedCurrency(FiatUnit.IRR);
+    await currency.setPrefferedCurrency(FiatUnit.IRT);
     await currency.init(true);
     cur = JSON.parse(await AsyncStorage.getItem(currency.EXCHANGE_RATES));
-    assert.ok(cur.BTC_IRR > 0);
+    assert.ok(cur.BTC_IRT > 0);
   });
 });
--- a/tests/integration/Currency.test.js
+++ b/tests/integration/Currency.test.js
@@ -42,9 +42,9 @@ describe('currency', () => {
     // assert.ok(cur.BTC_LBP > 0);
 
     // test Exir rate source
-    await currency.setPrefferedCurrency(FiatUnit.IRT);
+    await currency.setPrefferedCurrency(FiatUnit.IRR);
     await currency.init(true);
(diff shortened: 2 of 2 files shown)


marcosrdz (commented):
Whats the resulting behavior for a user who had the now deleted currency, updates, and opens the app?

Overtorment:
dont ask, test it yourself. 
i'd expect a crash, so some kind of migration code is needed

marcosrdz:
If the user had a currency that is no longer part of the app, it still references it given that we save the FiatUnit object in AsyncStorage. The user would have to go into currency settings and select a different currency. No crashes. I don't mind this to be as much of an issue.

Overtorment:
i would expect it to crash here: https://github.com/BlueWallet/BlueWallet/blob/c0c4d7c4116fc6d9b1bc7af2dfd323791fd06b44/models/fiatUnit.js#L109

mcsaeid:
> Whats the resulting behavior for a user who had the now deleted currency, updates, and opens the app?

That is a good point. I hadn’t thought of that. I don’t have the technical know-how to test any of this or propose a solution, but a migration code seems like the safer bet.

marcosrdz:
@mcsaeid  i would add it as a new currency instead of replacement until mitigation code is added

mcsaeid:
> @mcsaeid i would add it as a new currency instead of replacement until mitigation code is added

What if you kept IRR but only changed the currency sign from rial* to toman†? Right now the problem with the IRR currency is that the amount shown is one-tenth of what it is supposed to be. Correcting the sign to toman would solve that. Having two currencies, IRR and IRT, where both show the same amount, one correct and one incorrect, doesn’t sit well with me. Let me know what you think.

Of course, IRR would later need to be changed to IRT, and that is where the migration code comes into play.

*‌ ﷼
† تومان

marcosrdz:
This would require context for any developer not familiar with this modification as it feels like a "band aid". This can also affect the iOS/macOS Widgets. The cleanest way is to have mitigation code that reverts your currency to the device default and then you can select the new currency.

models/fiatUnits.json
```diff
@@ -113,9 +113,9 @@
       "locale": "hi-HN",
       "source": "CoinDesk"
     },
-    "IRR": {
-      "endPointKey": "IRR",
-      "symbol": "﷼",
+    "IRT": {
+      "endPointKey": "IRT",
+      "symbol": "تومان",
```
mcsaeid, line 118: @marcosrdz In this case, aren’t rial and toman just symbols shown to the user?

I guess I don’t understand how keeping “IRR” and simply changing the symbol wouldn’t be the solution here.

marcosrdz: Widgets dont use this property to render price. They use system builtin symbols based on the passed currency (such as JPY or CAD). Taking this approach would result in one symbol being rendered in app and a different one on widgets

mcsaeid: I see.

How would go about creating the new currency? Also, would you say waiting and fixing the issue properly is the better approach? There is no rush is what I am saying.

marcosrdz: You can do this by simply adding the entry into the JSON file rather than replacing an existing one.

mcsaeid: Done. What about Currency.test.js?

marcosrdz: It can be replaced with the new currency or keep the current one. The test checks that the API returns an actual value so either currency works for the test purpose

marcosrdz (approved):

models/fiatUnits.json
mcsaeid: Good. Then it would be better to have the new currency in there.

marcosrdz:
Running tests here:

https://github.com/BlueWallet/BlueWallet/pull/3858

PR #3915, opened 2021-09-26
Update importDiscovery.js

marcosrdz:
Slight UX enhancements

### DIFF
--- a/screen/wallets/importDiscovery.js
+++ b/screen/wallets/importDiscovery.js
@@ -2,7 +2,7 @@ import React, { useContext, useEffect, useState, useRef, useMemo } from 'react';
 import { ActivityIndicator, Alert, LayoutAnimation, StatusBar, StyleSheet, Text, View, FlatList } from 'react-native';
 import IdleTimerManager from 'react-native-idle-timer';
 import { useNavigation, useRoute, useTheme } from '@react-navigation/native';
-
+import ReactNativeHapticFeedback from 'react-native-haptic-feedback';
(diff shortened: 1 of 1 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/kh09n3zbwbqe4v7bc990ggd93m?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/8513)

screen/wallets/importDiscovery.js
```diff
@@ -108,10 +110,15 @@ const ImportWalletDiscovery = () => {
...
       active={selected === index}
-      onPress={() => setSelected(index)}
+      onPress={() => {
+        setSelected(index);
+        ReactNativeHapticFeedback.trigger('selection', { ignoreAndroidSystemSettings: false });
+      }}
     />
   );
 
+  const keyExtractor = (w, index) => `${w.id}_${index}`;
```
limpbrains, line 120: Why do you add an index here?

GladosBlueWallet:

screen/wallets/importDiscovery.js
marcosrdz: FlatList was complaining about keys being reused.

limpbrains: If so, then there is a bug. Its either `import` reported same wallet 2 times, or 2 wallets have same id.
Can you reproduce this? with secret have you used?

marcosrdz: Sent a PR to remove this since its only reproducible on debug, not release. Because the bundler refreshes in a way that causes this flatlist behavior.

PR #3941, opened 2021-09-28
Update fs.js

marcosrdz:


blue_modules/fs.js
```diff
@@ -115,16 +115,18 @@ const showImagePickerAndReadImage = () => {
...
-        if (asset.uri) {
-          const uri = asset.uri.toString().replace('file://', '');
-          LocalQRCode.decode(uri, (error, result) => {
-            if (!error) {
-              resolve(result);
-            } else {
-              reject(new Error(loc.send.qr_error_no_qrcode));
-            }
-          });
+        if (!response.didCancel) {
```
Overtorment, line 118: Can this condition be reversed to make early return?

Overtorment (approved):

blue_modules/fs.js
marcosrdz: It can be. Visually seeing a "if this condition is met then read whats about to happen...oh nvm return" is misleading compared to "if this condition is met then proceed with scan". But im fine with changing it if need be

GladosBlueWallet:

 [android in browser] https://appetize.io/app/ckkc9cbxc8qyc11dn34m18d168?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/8581)

GladosBlueWallet:

PR #4088, opened 2021-10-21
ADD: Add unique ID in Settings

marcosrdz:
There are scenarios(which have been reported) where the app crashes as soon as you open it. This makes it impossible to get the unique id so we can filter the crash log. Adding this information in the app's bundle allows the user to obtain this information for a bug report. 

Simulator Screen Shot - iPhone 13 mini - 2021-10-21 at 00 43 39


GladosBlueWallet:

 [android in browser] https://appetize.io/app/4fg829tghuupethac72wex26fm?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/8848)

GladosBlueWallet:

marcosrdz:
> very thoughtful addition!
> 
> are you sure its the same as the one we pass to bugsnag?
> 
> also, is there a way to do the same for android?

It is. I looked at the docs for getUniqueId. Ill check for how to implement the same on Android

PR #4187, opened 2021-11-09
FIX: refresh lndhub wallet balance after paying for invoice by scanning qrcode on main screen

### LINKED ISSUE #3679: Lightning balance doesn't update after payment from main screen
* Moto G7 Power, Android 10
* BlueWallet app version 6.2.3
* self-test passes? Open settings->about->scroll down, tap "Run self-test" ☑

I use my own installation of LndHub so I can check its logs. After paying using Lightning from the main screen (the "Scan" button in the bottom) the wallet balance doesn't update and the transaction doesn't appear on the main screen. Refreshing this screen (via pulling down from the top) doesn't help. There are no balance update requests in the LndHub log in both cases, only `/payinvoice`. It's possible to update the wallet balance and tx list by opening that wallet first and pulling down there but not on the main screen.

However, if I select that wallet manually and then press "Send" there, then "Scan" and pay an invoice the balance updates instantly and the tx is added to that wallet list. I can see a `/balance` request right after `/payinvoice` in the LndHub log. I think it should be trivial to refresh the wallet that has been used for the payment after it succeeds if it's used from the main screen as it's supposedly the most popular 1-tap case.

limpbrains:
`walletID` is not set when user scans code on main screen

closes #3679

### DIFF
--- a/screen/lnd/scanLndInvoice.js
+++ b/screen/lnd/scanLndInvoice.js
@@ -217,7 +217,7 @@ const ScanLndInvoice = () => {
       amountUnit: BitcoinUnit.SATS,
       invoiceDescription: decoded.description,
     });
-    fetchAndSaveWalletTransactions(walletID);
+    fetchAndSaveWalletTransactions(wallet.getID());
   };
 
   const processTextForInvoice = text => {


GladosBlueWallet:

 [android in browser] https://appetize.io/app/bdyw269f3hu59ejeru4b9mx2gr?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/9056)

marcosrdz (approved):

GladosBlueWallet:

PR #4504, opened 2022-02-14
FIX: min/max in lndCreateInvoice

### LINKED ISSUE #4471: Bug: LNURL amount not correctly when toggling currency
* your phone model and OS version: Iphone IOS
* BlueWallet app version: 6.2.17 (latest)
* self-test passes? Yes

After scanning a LNURL withdraw QR with a non-fixed amount, BlueWallet seems to use the max amount for sats but the the min amount for BTC and currency once you start toggling the unit, instead of converting them from the sats amount.

Example, LNURL withdraw object with
```
"minWithdrawable":"444000",
"maxWithdrawable":"888000"
```

Will show this, each shot is while tapping the swap arrows:
bluewallet

On top, for the sats amount, BW will validate if the amount is within the allowed range, but when I'm eg at 444 BTC, it would happily let me try to send this invoice to the merchant.

limpbrains:
I removed  auto-formatting, which was supper annoying, and replaced it with form on submit validation.

https://user-images.githubusercontent.com/155891/153920624-3533569a-b055-4552-ba5e-de967ebf0121.mp4

Also super-nice log in `AmountInput.js`that looks like this:
`100000000(sats) -> 100000000(sats) -> 42723.25(local_currency)`
one line, instead of 3

`AmountInput.js` is a true mess and deserves refactoring alongside with loc/format and `currencies.js`

closes #4471

### DIFF
--- a/components/AmountInput.js
+++ b/components/AmountInput.js
@@ -75,7 +75,7 @@ class AmountInput extends Component {
    */
   onAmountUnitChange(previousUnit, newUnit) {
     const amount = this.props.amount || 0;
-    console.log('was:', amount, previousUnit, '; converting to', newUnit);
+    const log = `${amount}(${previousUnit}) ->`;
     let sats = 0;
     switch (previousUnit) {
       case BitcoinUnit.BTC:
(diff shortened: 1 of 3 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/agq7etcx4j9qa42h2y2mv6puew?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/9583)

marcosrdz (approved):

GladosBlueWallet:

PR #4750, opened 2022-05-22
DOC: update coc to more standard version

JamesWeiMoseley:
This is a more standard Code of Conduct that is similar to what most open source projects use.  Thanks for considering it.


ManiacEinstein:
> This is a more standard Code of Conduct that is similar to what most open source projects use.  Thanks for considering it. 

Choose to be kind ✨️ ☺️

Overtorment:
no

PR #4774, opened 2022-06-04
ADD: loc de, fr, pl

limpbrains:


ios/fastlane/metadata/fr-FR/description.txt
```diff
@@ -49,5 +49,5 @@ Un portefeuille Lightning sans aucune configuration nécessaire. Des transaction
 Acheter du bitcoin
 Rejoignez la révolution la finance ouverte avec la possibilité d'acheter du bitcoin directement depuis votre portefeuille.
 
-Local Trader
+Trader Local
```
Overtorment, line 52: btw its time we get rid of local trader texts since it got removed @ncoelho @marcosrdz

Overtorment:
hmm. why did so many tests fail..?

ios/fastlane/metadata/fr-FR/description.txt
marcosrdz: It didnt. Its in settings - tools

marcosrdz (approved):

ios/fastlane/metadata/fr-FR/description.txt
Overtorment: O_o

Issue #4852, opened 2022-07-06
Invalid LNDHub URI in latest release


djschnei21:
Please provide:

* your phone model and OS version: Android 12 on an S21 Ultra
* BlueWallet app version: 6.3.1
* self-test: FAILED "ErrorL Linking can not open https url"
* userId:ea6a742dbce449c4

On the latest release, connecting to my LNDHub fails unless I use the tor address.  Local http or tailscale http no longer work and return "Invalid LNDHub URI" when entered.  Tailscale is my preferred method for performance.  Worked fine on previous release.

Overtorment:
recently there was a change in policy (not by us), and unencrypted traffic (http uris) became forbidden.
we can add exceptions to those rules (like we did for *.onion uris). what uri are you using?

djschnei21:
trying to use my http://100.x.y.z:3008 private tailscale IP.  Tailscale is a private VPN tunnel, would the exception apply there as well?

Overtorment:
https://tailscale.com/kb/1015/100.x-addresses/

hmm, not sure if its possible to whitelist wildcard addresses like 100.*

atec145:
I'm not shure if this is the right place to put, if not please delete.
Since the last update to 6.3.1 (and perhaps 6.3.0 but not testable because of the tor bug) the lndHub URI scanned by QR seems to be different from the shown text Node URI. Wallets created with the (QR scanned) LNDHub would not work.
Before 6.3.0 it worked as expected.

My environment:
LNDHub on Citadel Version 0.5.24
Android 11 / Bluewallet 6.3.1

Overtorment:
fixed in https://github.com/BlueWallet/BlueWallet/pull/4854

Funman2:
How do i implement this on android? Tor is so slow...

PR #4854, opened 2022-07-06
allow tailscale to connect using "MagicDNS" over HTTP

djschnei21:
#4852 

Would allow [Tailscale](https://community.getumbrel.com/t/how-to-use-umbrel-with-tailscale/6782) users to use ["MagicDNS"](https://tailscale.com/kb/1081/magicdns/) URIs as their LNDHub connection URI.  Tailscale is just a zero-config fireguard VPN, so these connections are already encrypted and require authentication and authorization.

### DIFF
--- a/android/app/src/main/res/xml/network_security_config.xml
+++ b/android/app/src/main/res/xml/network_security_config.xml
@@ -4,5 +4,6 @@
         <domain includeSubdomains="true">10.0.2.2</domain>
         <domain includeSubdomains="true">localhost</domain>
         <domain includeSubdomains="true">onion</domain>
+        <domain includeSubdomains="true">tailscale.net</domain>
     </domain-config>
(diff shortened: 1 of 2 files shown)


djschnei21:
tailscale also has an ["HTTPS" feature](https://tailscale.com/kb/1153/enabling-https/) in preview which wouldn't require the exception, however using this would increase the risk of doxing yourself through exposing the machine name.  The MagicDNS route over http preserves your privacy by not leaking anything outside of your personal tailscale network

Overtorment:
can you pls mirror the change for `ios/` as well? thanks

djschnei21:
done! (I think 😂)

Overtorment (approved):

marcosrdz (approved):

Issue #5016, opened 2022-09-12
JSON files are opened in BlueWallet app by default (iPadOS)


kstlv:
- iPadOS 15.5
- BlueWallet 6.3.1

I have json files in Files app (iOS file manager). Tap on a JSON file will open it in the BlueWallet app by default.

This should not happen by default. If BlueWallet has any functionality for interacting with json files, please warn the user about it.

marcosrdz:
BlueWallet recognizes JSON files to process them to verify for wallet imports from Electrum and hardware devices such as Keystone and ColdCard.

PR #5029, opened 2022-09-18
FIX: JSON files are opened in BlueWallet app by default (iPadOS) #5016

marcosrdz:


Overtorment (approved):
Lgtm. Do we have same file associations for Android..?

marcosrdz:
> Lgtm. Do we have same file associations for Android..?

Only for PSBT files

GladosBlueWallet:

PR #5035, opened 2022-09-23
Add: return sendPayment response to webln.sendPayment call

bumi:
Currently BlueWallet does not return the preimage on the webln.sendPayment calls. This breaks implementations that wait for the response and also use the preimage to validate payments.
This change tries to add the response similar to the makeInvoice call currently does.

### DIFF
--- a/screen/lnd/browser.js
+++ b/screen/lnd/browser.js
@@ -44,12 +44,15 @@ var webln = {
     });
   },
   sendPayment: function (paymentRequest) {
-    window.ReactNativeWebView.postMessage(JSON.stringify({ sendPayment: paymentRequest }));
+    var id = Math.random(); // eslint-disable-line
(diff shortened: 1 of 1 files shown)


Overtorment:
is this still being worked on?
who even uses webln these days?

bumi:
> who even uses webln these days?

guess that means there is no interest in doing it properly by BlueWallet? 
I can understand that there was not much adoption at the time when you had added it - it was a bit too early I guess.
But there is a growing community around webln (new websites: https://webbtc.dev/webln + https://www.webln.guide/ ).
webln enables features and UX that otherwise would not be possible. 

I had started this PR to make the webln.sendPayment call usable in BlueWallet (without return value it is mostly useless).
But I failed to run the app in dev mode and did not have enough time to learn how to run it locally in a simulator to properly test this change. 

maybe somebody else could help? - If BlueWallet has an interest in keeping the lapp browser - otherwise better remove it, because as of now it is half broken.

Overtorment:
is webln dead yet..?

PR #5101, opened 2022-10-11
Test ldk patch

Overtorment:


Overtorment:
ios failed with 


```

warning: Run script build phase 'Bundle React Native code and images' will be run during every build because it does not specify any outputs. To address this warning, either add output dependencies to the script phase, or configure it to run in every build by unchecking "Based on dependency analysis" in the script phase. (in target 'BlueWallet' from project 'BlueWallet')
warning: Run script build phase 'Upload source maps to Bugsnag' will be run during every build because it does not specify any outputs. To address this warning, either add output dependencies to the script phase, or configure it to run in every build by unchecking "Based on dependency analysis" in the script phase. (in target 'BlueWallet' from project 'BlueWallet')
warning: Run script build phase 'Upload Bugsnag dSYM' will be run during every build because it does not specify any outputs. To address this warning, either add output dependencies to the script phase, or configure it to run in every build by unchecking "Based on dependency analysis" in the script phase. (in target 'BlueWallet' from project 'BlueWallet')
/Users/runner/work/1/s/ios/BlueWallet.xcodeproj: warning: The iOS deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 10.0, but the range of supported deployment target versions is 11.0 to 16.0.99. (in target 'TodayExtension' from project 'BlueWallet')
warning: Run script build phase 'Upload Bugsnag dSYM' will be run during every build because it does not specify any outputs. To address this warning, either add output dependencies to the script phase, or configure it to run in every build by unchecking "Based on dependency analysis" in the script phase. (in target 'WidgetsExtension' from project 'BlueWallet')
** ARCHIVE FAILED **


The following build commands failed:
	CompileSwift normal arm64 /Users/runner/work/1/s/node_modules/rn-ldk/ios/RnLdk.swift (in target 'rn-ldk' from project 'Pods')
	SwiftCompile normal arm64 Compiling\ RnLdk.swift /Users/runner/work/1/s/node_modules/rn-ldk/ios/RnLdk.swift (in target 'rn-ldk' from project 'Pods')
(2 failures)
##[error]Error: /usr/bin/xcodebuild failed with return code: 65
```

jurvis:
ah, okay. I was able to eventually reproduce the issue too. let me take another look at it and ping once I figure something out

jurvis:
Cross posting from a message on [LDK Discord](https://discord.com/channels/915026692102316113/1029089962899034142/1029467313931571331):

> for the benefit of others: that didn't work, at all. While `pod install` completes, it unfortunately excludes the `xcframework` from being included in the Pod target, hence the source files for `RnLdk` has no library to reference.

It appears to me that this is an issue with Cocoapods, since we certainly do not use dynamic libraries in the packaged xcframework. There are a couple of reports on this issue:
https://github.com/CocoaPods/CocoaPods/issues/10332
https://stackoverflow.com/questions/66730187/what-should-be-the-podspec-file-for-using-local-xcframework-in-react-native-lib

From my experimentation, it looks like Cocoapods' `vendored_framework` API just doesn't support frameworks that may use symlinks in its contents (necessary for supporting multiple platforms sharing the same architecture). That's the reason why a `xcframework` with support for only `ios-arm64` and `ios-arm64_x86_64-simulator` will work. 

Unfortunately, Cocoapods seems to be a little quiet these days (no commits since June), so I wouldn't be holding my breath for a fix soon.

So, if you aim to support Universal build targets (macOS, Mac Catalyst, iOS; or some selection of 2), I believe your only option is to use Swift Package Manager.

As far as what that means for React Native implementations like `rn-ldk`, I'm currently exploring if we can publish it as a Swift package that React Native `xcodeproj`s can import directly. I'll share the PRs when I get that working.

PR #5253, opened 2023-01-01
FIX: priority fee will be always at least 2 sat/vb to beat other 1 sa…

### LINKED ISSUE #4902: Fast speed should be at minimum 2sat/bytes
Because 2 sat/bytes currently $ value is very low I think FAST should be at minimum 2sat/bytes. 

Now for example in the last 5 hours there were multiple times where 1sat/byte would not have been enough.

image

But BLUE isn't suggesting more than 1sat for fast speed.

image

I suggest to override the fee calculation so that fast is never lower than 2

Overtorment:
…t/vb transactions in the mempool (closes #4902)

### DIFF
--- a/blue_modules/BlueElectrum.js
+++ b/blue_modules/BlueElectrum.js
@@ -832,7 +832,7 @@ module.exports.estimateFees = async function () {
   const _slow = await module.exports.estimateFee(144);
 
   // calculating fast fees from mempool:
-  const fast = module.exports.calcEstimateFeeFromFeeHistorgam(1, histogram);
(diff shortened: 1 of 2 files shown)


GladosBlueWallet:

 [android in browser] https://appetize.io/app/qrxl4hgvislf4ybdbhp4ojqe2y?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/10775)

limpbrains (approved):

GladosBlueWallet:

PR #5260, opened 2023-01-06
FIX: secure keystore keys should not leave device

### LINKED ISSUE #5259: Secret keystore should not be synced to iCloud
Secure items stored in keychain should not be synced to iCloud nor migrated to new devices, as most security-concerned users would be concerned about their keys leaving their devices, even if encrypted.

As this is fundamental to Bluewallet's cores, as shown in the readme:
image

From `react-native-secure-key-store` docs:
image

AbdelrahmanManz:
closes #5259

### DIFF
--- a/class/app-storage.js
+++ b/class/app-storage.js
@@ -67,7 +67,7 @@ export class AppStorage {
    */
   setItem = (key, value) => {
     if (typeof navigator !== 'undefined' && navigator.product === 'ReactNative') {
-      return RNSecureKeyStore.set(key, value, { accessible: ACCESSIBLE.WHEN_UNLOCKED });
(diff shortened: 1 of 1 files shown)


Overtorment:
good catch! 
@marcosrdz @limpbrains thoughts..?

marcosrdz (approved):

limpbrains:
@AbdelrahmanManz Awesome job!

limpbrains (approved):

marcosrdz:
> I hope it wont break someone's usecase

Perhaps those upgrading their device and using a backup to restore on a new device

Overtorment:
> > I hope it wont break someone's usecase
> 
> Perhaps those upgrading their device and using a backup to restore on a new device

@ncoelho would need some market research on this, what is best

PR #5400, opened 2023-03-24
Enable lightning(offchain) wallet transactions to be exported to csv

masikotimo:
Summary of changes 

Enable exportation of wallet history. transactions to be done by both on-chain and off-chain

### DIFF
--- a/screen/wallets/details.js
+++ b/screen/wallets/details.js
@@ -420,9 +420,15 @@ const WalletDetails = () => {
 
     for (const transaction of transactions) {
       const value = formatBalanceWithoutSuffix(transaction.value, BitcoinUnit.BTC, true);
-      csvFile +=
-        '\n' +
-        [new Date(transaction.received).toString(), transaction.hash, value, txMetadata[transaction.hash]?.memo?.trim() ?? ''].join(','); // CSV line
+
+      let hash = transaction.hash;
+      let memo = txMetadata[transaction.hash]?.memo?.trim() ?? '';
+
+      if (wallet.chain === Chain.OFFCHAIN) {
+        hash = transaction.payment_hash;
+        memo = transaction.description;
+      }
+      csvFile += '\n' + [new Date(transaction.received).toString(), hash, value, memo].join(','); // CSV line
     }
 
     await writeFileAndExport(`${wallet.label.replace(' ', '-')}-history.csv`, csvFile);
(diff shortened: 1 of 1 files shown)


screen/wallets/details.js
```diff
@@ -605,7 +605,7 @@ const WalletDetails = () => {
               <View>
                 <BlueSpacing20 />
                 <SecondButton onPress={navigateToWalletExport} testID="WalletExport" title={loc.wallets.details_export_backup} />
-                {wallet.chain === Chain.ONCHAIN && walletTransactionsLength > 0 && (
+                {walletTransactionsLength > 0 && (
```
marcosrdz, line 608: Needs code to use the "payment_hash" property instead of txid.

marcosrdz (changes requested):

screen/wallets/details.js
masikotimo: @marcosrdz  so if i do this it could work out (wallet.chain === Chain.ONCHAIN || wallet.chain === Chain.OFFCHAIN ) without making deletions ?

marcosrdz: The change you made is correct but in you need to add what I pointed out. Line 435..if its off chain then lookout payment_hash and invoice_description

masikotimo: Thank you @marcosrdz , I appreciate your feedback greatly ,  so to clarify what you mean is I should have a check that does if its off chain then the csv should have payment_hash instead of txid and add  invoice_description too after the memo field? is that right 

I can add this check during the creation of the csv  here


`
let csvFile = [
      loc.transactions.date,
      loc.transactions.txid,
      `${loc.send.create_amount} (${BitcoinUnit.BTC})`,
      loc.send.create_memo,
    ].join(','); // CSV header
`

Overtorment: maybe dont touch the header, just where it writes actual lines to csv


```js
[new Date(transaction.received).toString(), transaction.hash, value, txMetadata[transaction.hash]?.memo?.trim() ?? ''].join(','); // CSV line
```

masikotimo: Alright let me do so , Thank you @Overtorment

screen/wallets/details.js
```diff
@@ -420,9 +420,15 @@ const WalletDetails = () => {
...
-      csvFile +=
-        '\n' +
-        [new Date(transaction.received).toString(), transaction.hash, value, txMetadata[transaction.hash]?.memo?.trim() ?? ''].join(','); // CSV line
+
+      let hash = transaction.hash;
+      let memo = txMetadata[transaction.hash]?.memo?.trim() ?? '';
+
+      if (wallet.chain === Chain.OFFCHAIN) {
+        hash = transaction.payment_hash;
+        memo = transaction.description;
```
masikotimo, line 429: @marcosrdz is this okay ?

marcosrdz (approved):

masikotimo:
@Overtorment   that looks like a keystore format issue that's causing it to fail.

masikotimo:
> @Overtorment that looks like a keystore format issue that's causing it to fail.

@Overtorment  any feedback on this?

Overtorment (approved):
lgtm

Overtorment:
lets run full test suite & builds here: https://github.com/BlueWallet/BlueWallet/pull/5463

@masikotimo as soon as apk build is ready, can you pls test  this functionaklity? for both onchain & lightning wallets

PR #5642, opened 2023-08-06
FIX removed bridge.swift from ios, it is not used

kshu2011:


marcosrdz:
Its needed for xcode bridging between objc and swift

PR #5659, opened 2023-08-17
Added onboarding tutorial based on usability findings

bossnani:


Overtorment:
rtf files dont belog in the repo. write a blogpost instead.

PR #5791, opened 2023-10-29
FIX: Set base64 for encoding options

marcosrdz:
Docs show utf8 is the default


GladosBlueWallet:

 [android in browser] https://appetize.io/app/zlchcny5fv6tnizey5spxx64vq?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/11776)

Overtorment:
I think it will break everything, it will put base64 in files

PR #5808, opened 2023-11-05
DEL: Remove "removelisteners" call as it causes crashes when debugging

marcosrdz:


GladosBlueWallet:

 [android in browser] https://appetize.io/app/zuublyukd546mfhj3kwbxgdrlu?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/11825)

App.js
```diff
@@ -117,17 +117,6 @@ const App = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [walletsInitialized]);
 
-  useEffect(() => {
-    return () => {
```
Overtorment, line 121: Did we ever need those? Why did they exist?

Overtorment (approved):

GladosBlueWallet:

App.js
Overtorment: @marcosrdz

marcosrdz: The purpose is to remove listeners when you need to unmount that component, but it doesn't make sense given that this is the root component. It will never be unmounted unless u kill the release app or refresh the debug app.

PR #6034, opened 2024-01-21
REF: Splash Screen

marcosrdz:

PR #6080, opened 2024-02-03
FIX: BIometrics unlock

marcosrdz:


UnlockWith.js
```diff
@@ -121,9 +121,9 @@ const UnlockWith = () => {
       }
 
       setBiometricType(bt);
-      if (!biometricType || storageIsEncrypted) {
```
Overtorment, line 124: its a shame that this wasnt caught when it was initially merged.
also, i suppose TS would have caught this automatically

Overtorment: i checked, it was me who "reviewed" and  approved it: https://github.com/BlueWallet/BlueWallet/pull/6034

Overtorment (approved):

GladosBlueWallet:

 [android in browser] https://appetize.io/app/ojgbf5kym7cczptmpk6b5hhgbi?device=pixel4

[download apk](https://lambda-download-android-build.herokuapp.com/download/335)

UnlockWith.js
marcosrdz: I doubt TS would've caught this as the value and type are valid. just the wrong item being checked against

PR #6226, opened 2024-03-03
REF: TransactionPendingIconBig to TS

Danswar:


marcosrdz (changes requested):
Try using "git mv" so history is preserved

Overtorment (changes requested):
`git mv`

Overtorment:
merged in other PR, thanks

PR #6237, opened 2024-03-06
REF: taproot wallet to TS

abhiShandy:


Overtorment:
lol i forgot that we have a taproot wallet

PR #6358, opened 2024-03-31
REF: useReducer for on app launch

marcosrdz:


Overtorment (approved):
@marcosrdz remind me whats the point of this approach?
i find it hard to read.

@limpbrains what do you think?

marcosrdz:
> @marcosrdz remind me whats the point of this approach?
> 
> i find it hard to read.
> 
> 
> 
> @limpbrains what do you think?

Its so every possible UI state is documented and predictable

limpbrains:
I like useReducer, but I don't think we should refactor all screens to use it. Only complex ones.
Component with 2 state variables and 2 actions looks simpler with just useState

limpbrains (approved):

GladosBlueWallet:

PR #6689, opened 2024-06-10
ADD: Insert contact on Android

marcosrdz:


screen/send/SendDetails.tsx
```diff
@@ -1307,8 +1307,11 @@ const SendDetails = () => {
 
     return (
       <BottomModal isVisible={optionsVisible} onClose={hideOptions}>
-        <KeyboardAvoidingView enabled={!isTablet} behavior={Platform.OS === 'ios' ? 'position' : undefined}>
+        <KeyboardAvoidingView enabled={!isTablet} behavior={undefined}>
           <View style={[styles.optionsContent, stylesHook.optionsContent]}>
+            {wallet?.isBIP47Enabled() && (
```
Overtorment, line 1312: you need feature flag check as well, cause if the wallet does not support it, `isBIP47Enabled` method doesnt exist and the app will crash

Overtorment (changes requested):

screen/send/SendDetails.tsx
marcosrdz: fixed

Overtorment (approved):

GladosBlueWallet:

PR #6692, opened 2024-06-11
 DOC: JA language update

librepax:
Modified JA language app store descriptions, keywords, promotional texts


librepax:
Small changes on keywords.txt and promotional.txt... changed commas within text of former file from ',' to '、' (as used in JP) and replaced certain keywords to make more sense. (ex. もっともらしい否認 seen in original text is a literal translation of plausible deniability which didn't make sense so replaced with 情報隠蔽技術 or information hiding technology)

Made larger changes on description.txt for the purposes of making the sentences flow more naturally to native speakers, but the translation is (of course) still based on the EN text (https://github.com/BlueWallet/BlueWallet/blob/master/ios/fastlane/metadata/en-US/description.txt)

librepax:
noted, reverted to old commas on keywords.txt

PR #6693, opened 2024-06-11
FIX: bip47 notif tx when inserting contact

Overtorment:
@marcosrdz i removed a couple of warnings as well, i dunno if you need them


marcosrdz (approved):

GladosBlueWallet:

PR #6743, opened 2024-06-25
OPS: fix android build

Overtorment:
@marcosrdz i moded the dependency from dev to prod, but im not  sure if its right. this dep was historically in dev deps..?
if yes, why do we need it in prod dependencies now? is something else is misconfigured so it requires this dependency now as well?


marcosrdz:
Its due to a change in metro.config.js. The change is mandatory to upgrade to 73

marcosrdz (approved):

GladosBlueWallet:

PR #6818, opened 2024-07-24
TST: add retries for integration tests on CI

Overtorment:
I experimented with built-in `jest.retryTimes(1);` but it didnt really work for me, probably because it retries testcases but does not re-run `beforeAll` to setup electrum connection (which is often the cause of a failed test)

so instead i coded similar mechanism we have for e2e:
each green test  writes a lock file in /tmp (this happens in new custom jest reporter), and then test suite is rerun,
and previously passed  tests are skipped (this happens in a custom jest environment)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,6 +52,9 @@ jobs:
     docker:
       - image: cimg/node:16.20.2
 
+    environment:
+      RETRY: "1"
+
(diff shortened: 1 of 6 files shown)


marcosrdz (approved):

limpbrains (approved):

GladosBlueWallet:

PR #6982, opened 2024-08-27
OPS: Reproducible APK

marcosrdz:


Overtorment:
@marcosrdz what can we take from this PR and merge into master asap that is safe, easy & definately improves  reproduceability? for example,  locking gradle version seems to be a good idea.

PR #7076, opened 2024-09-16
Update en.json

svrgnty:
Fix grammatical errors

## Summary by CodeRabbit

- **Bug Fixes**
	- Corrected grammatical error in user-facing text by changing "an universal address" to "a universal address."

### DIFF
--- a/loc/en.json
+++ b/loc/en.json
@@ -106,7 +106,7 @@
     "minSats": "Minimal amount is {min} sats",
     "minSatsFull": "Minimal amount is {min} sats or {currency}",
     "qrcode_for_the_address": "QR Code for the address",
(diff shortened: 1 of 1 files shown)


Overtorment:
our english-native disagrees

svrgnty:
"An universal" is incorrect. Though it does begin with a vowel,  the "u" is pronounced with a consonant sound. It is similar to "a university", or "a unique idea". You do not use "an" for those cases
- [When should I use "a" vs "an"?](https://english.stackexchange.com/questions/152/when-should-i-use-a-vs-an)
- [See examples in Universal definition](https://www.merriam-webster.com/dictionary/universal)

PR #7553, opened 2025-01-31
ADD: Offline mode toggle in Settngs bundle

marcosrdz:
In rare occasions having bad server status/info leads to app startup issues. This toggle provides the ability to force offline/online mode to circumvent such scenarios

### DIFF
--- a/blue_modules/BlueElectrum.ts
+++ b/blue_modules/BlueElectrum.ts
@@ -168,7 +168,6 @@ export const removePreferredServer = async () => {
 export async function isDisabled(): Promise<boolean> {
   let result;
   try {
-    await DefaultPreference.setName(GROUP_IO_BLUEWALLET);
     const savedValue = await DefaultPreference.get(ELECTRUM_CONNECTION_DISABLED);
     console.log('Getting Electrum connection disabled state:', savedValue);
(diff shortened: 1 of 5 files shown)


GladosBlueWallet (commented):
I see a shiny new “Offline Mode” toggle, and also several ways to make it lie:

- `BlueElectrum.isDisabled()` / `setDisabled()` removed `DefaultPreference.setName(GROUP_IO_BLUEWALLET)`. If the app relies on the app-group suite for persistence, this will silently read/write the wrong store (or fail to share between targets). That’s not “offline mode”, that’s “Schrödinger’s setting”.
- `Settings.bundle/Root.plist` adds a `PSToggleSwitchSpecifier` with empty `Title` and empty `Key`. iOS Settings expects a non-empty `Key`; empty keys can lead to undefined behavior or clobbering.
- Settings bundle key is `electrum_disabled`, but the JS side uses `ELECTRUM_CONNECTION_DISABLED`. If those don’t match exactly, the toggle won’t control anything.

At minimum: restore the correct `DefaultPreference` suite name, remove the empty-key specifier, and ensure the iOS Settings key matches the runtime preference key.

PR #7575, opened 2025-02-09
Wtch

marcosrdz:


Overtorment:
@marcosrdz whats the status of this?  also, conflicts

GladosBlueWallet (commented):
WatchConnectivity “fixes”, with a side order of footguns:

- `AppDelegate` now declares `WCSessionDelegate` but the patch doesn’t add the required delegate methods. On iOS this is not optional ceremony; you’ll get warnings at best and broken session lifecycle at worst.
- You’re activating `WCSession` in multiple places (iOS app + watch extension + `ConnectivityManager.shared`). Double activation/delegate assignment races are how messages get “lost” and everyone blames the radio.
- `WatchDataSource` subscribes to `NotificationCenter` but never removes observers (no `deinit` cleanup). That’s a leak + duplicate handlers if the object is recreated.
- `sendMessage` is used without consistent gating on activation state / reachability; you’ll get sporadic failures depending on timing.
- Xcode project diff includes `/* (null) in Frameworks */` entries. That smells like a broken PBX reference that can randomly explode CI builds.

Please implement the proper WCSessionDelegate lifecycle methods (both sides), ensure only one session owner/delegate, and make observer teardown explicit.

PR #7577, opened 2025-02-10
FIX: Alert the user if notifications deregister failed

marcosrdz:


GladosBlueWallet:

PR #7591, opened 2025-02-13
FIX: Confirm wallet delete alert would poptotop regardless of confirm…

marcosrdz:
…ation


GladosBlueWallet:

PR #7594, opened 2025-02-14
REF: Wrap BlueElecrum promises in try catch

marcosrdz:
Changed null to undefined as macOS crashes on attempted unchecked access to null values (making the behavior across all systems the same)


Overtorment (approved):

GladosBlueWallet:

PR #7597, opened 2025-02-15
REF: Start removing the need for merge to be true

marcosrdz:


Overtorment (changes requested):
elaborate. dont be shy to eli5 and use many words

marcosrdz:
https://reactnavigation.org/docs/navigation-actions/#navigate
default is false. the code for the screen and its logic should be structured in a way where it doesnt require adjusting states manually per navigation. on its first launch it shoud know how to handle the params

GladosBlueWallet (commented):
Navigation “merge” cleanup is fine, but you’re also deleting behavior:

- `ScanQRCode` no longer calls `onBarScanned` when `launchedBy` is falsy. Any caller relying on the callback (instead of route param merge) now gets… nothing. Silent failure is such a classic.
- `WalletsList` effect no longer clears `route.params.onBarScanned`. That makes the scan result sticky; re-renders / refocus can repeatedly re-process the same payload.
- Passing `merge: undefined` while `merge` defaults to `true` (`merge = true`) means you’ll still merge. If the goal is “undefined by default”, don’t default it to `true`.

Suggestion: restore callback fallback, and always clear consumed params (`setParams({ onBarScanned: undefined })`) to avoid repeated processing.

PR #7599, opened 2025-02-16
REF: Receive custom modal

marcosrdz:
Use builtin formsheet


Overtorment:
holy shit 65 files changed O_O

Overtorment (changes requested):
where exactly is the invocation of new modal?
why there are so many changes?

given that the diff is big, and js file (not ts) file is modified, and old dep is notremoved im inclined to close this pr till better times.

what i would really like to see is :

* comprehensive list of our modals that we need to replace with new modal
* new modal isolated in some kind of module thats easy to use (and read)
* all old modals replaced with new one, with _minimal_code change  (think "drop-in replacement")
* old dep removed

marcosrdz:
> where exactly is the invocation of new modal?
> 
> why there are so many changes?
> 
> 
> 
> given that the diff is big, and js file (not ts) file is modified, and old dep is notremoved im inclined to close this pr till better times.
> 
> 
> 
> what i would really like to see is :
> 
> 
> 
> * comprehensive list of our modals that we need to replace with new modal
> 
> * new modal isolated in some kind of module thats easy to use (and read)
> 
> * all old modals replaced with new one, with _minimal_code change  (think "drop-in replacement")
> 
> * old dep removed

The changes are huge. Theres no way to do dropin for this

marcosrdz:
Dont close it. Ill replace all modals

marcosrdz:
Hold review...till bug is fixed:
https://github.com/software-mansion/react-native-screens/issues/2664

GladosBlueWallet (commented):
This PR is a 95-patch “refactor” that also rewires scanning/navigation. Risks I’m seeing:

- Multiple screens now `navigate('ScanQRCode')` but don’t consistently pass required/expected params (and at least one call uses `showImportFileButton` vs `showFileImportButton`). That’s a runtime bug generator.
- Several effects consume `route.params?.onBarScanned` but don’t reliably clear it afterward, or run without a dependency array. Result: repeated processing loops, duplicated navigation, and mysterious “why did it navigate again?” reports.
- `scanQrHelper` was changed from Promise-returning to `void`, but there are still call sites that previously awaited results. If any survived in the full diff, they’ll compile yet do the wrong thing.
- New uses of `popTo`/`navigate` with merge semantics look inconsistent (sometimes passing merge inside params vs options). That’s exactly the kind of API mismatch Nav v7 punishes.

Given the blast radius: please split this into smaller PRs (Nav upgrade vs scan flow vs settings changes), add a minimal regression test for scan→return parameter clearing, and fix the param naming inconsistencies before this becomes a labyrinth.

PR #7628, opened 2025-02-23
refactor: bring back scan qr helper

Overtorment:
looked up rnav docs, apparently passing a callback in navigation params is ok unless you need to deeplink to said screen, or want to save nav state to disk or smth...

https://reactnavigation.org/docs/troubleshooting/#i-get-the-warning-non-serializable-values-were-found-in-the-navigation-state

image

@marcosrdz what can be problematic in the future in this setup? 

also refactored 2 screens to use this helper, got rid of useEffect fuckery which is very unobvious;  and theres more deletions than additionsin those screens

### DIFF
--- a/helpers/scan-qr.ts
+++ b/helpers/scan-qr.ts
@@ -1,5 +1,6 @@
 import { Platform } from 'react-native';
 import { check, request, PERMISSIONS, RESULTS } from 'react-native-permissions';
+import { navigationRef } from '../NavigationService.ts';
 
 const isCameraAuthorizationStatusGranted = async () => {
   const status = await check(Platform.OS === 'android' ? PERMISSIONS.ANDROID.CAMERA : PERMISSIONS.IOS.CAMERA);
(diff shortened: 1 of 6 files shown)


helpers/scan-qr.ts
```diff
@@ -10,4 +10,14 @@ const requestCameraAuthorization = () => {
   return request(Platform.OS === 'android' ? PERMISSIONS.ANDROID.CAMERA : PERMISSIONS.IOS.CAMERA);
 };
 
-export { isCameraAuthorizationStatusGranted, requestCameraAuthorization };
+const scanQrHelper = (navigateFunc: any): Promise<string> => {
```
marcosrdz, line 13: Use navigationRef in NavigationService

marcosrdz (changes requested):

helpers/scan-qr.ts
Overtorment: done

marcosrdz (approved):

GladosBlueWallet:

image

limpbrains (approved):

PR #7650, opened 2025-02-27
FIX: when pressing refill on lightning wallet, address would not subs…

Overtorment:
…itute


marcosrdz (approved):

GladosBlueWallet:

PR #7651, opened 2025-02-27
FIX: Restore camera access to mac app

marcosrdz:


Overtorment (approved):
lgtm but linter is not happy

GladosBlueWallet:

Issue #7653, opened 2025-02-27
Crash when scanning LN


Overtorment:
![Image](https://github.com/user-attachments/assets/42ac60ca-be1a-4726-bc2a-6a38fe81ab88)

![Image](https://github.com/user-attachments/assets/01d2b196-700f-4825-b091-6a2e02596c9a)

Overtorment:
```
bitcoin:BC1QR7P8NSYPZEJY4KP7CJS0HL5T9X0VF3AYF6UQPC?amount=0.00185579&lightning=LNBC1855790N1PNUPWSFPP5P5RVQJA067PV6NJQ3EFKLP78TN6MHUK842ZFGDCTXRDSGNTY765QDZ62PSKJEPQW3HJQSNPD36XJCEQFPHKUETEVFSKGEM9WGSRYVPJXSSZSNMJV3JHYGZFGSAZQARFVD4K2AR5V95KCMMJ9YCQZPUXQZ6GSP53E4EX9YTD2MGDN2C2CFA0J0SM3E7PVLPJ208H5LMYPNJMGZ7RLGS9QXPQYSGQ6GQMEQXJKKF2DHXJK8XQ4WGLM5NTE3RKEXGYQC6HYGFKS9SHHA6HL9X4339MXHNNQFSH7TS62PU8T9RSWTK6HQ4LV4GW3DPD25DQ8UQQYC909N
```

PR #7656, opened 2025-02-28
FIX: crash when scanning invoice with both ln and onchain address (cl…

Overtorment:
…oses #7653)

in another refactoring parameter was incorrectly called `navigation` instead of `navigate`.
fixed it by writing a proper type for argument (instead of `any`)

### DIFF
--- a/navigation/SendDetailsStackParamList.ts
+++ b/navigation/SendDetailsStackParamList.ts
@@ -29,6 +29,15 @@ export type SendDetailsParams = {
   };
 };
 
+export type TNavigation = {
+  pop: () => void;
+  navigate: () => void;
+};
+
+export type TNavigationWrapper = {
+  navigation: TNavigation;
+};
+
 export type SendDetailsStackParamList = {
   SendDetails: SendDetailsParams;
   Confirm: {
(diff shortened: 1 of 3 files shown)


navigation/SendDetailsStackParamList.ts
```diff
@@ -29,6 +29,15 @@ export type SendDetailsParams = {
...
 
+export type TNavigation = {
+  pop: () => void;
+  navigate: () => void;
+};
+
+export type TNavigationWrapper = {
+  navigation: TNavigation;
+};
+
```
Overtorment, line 40: not sure if those type definitions belong here or should be moved

screen/wallets/SelectWallet.tsx
```diff
@@ -72,7 +72,7 @@ const SelectWallet: React.FC = () => {
   const onPress = (item: TWallet) => {
     triggerHapticFeedback(HapticFeedbackTypes.Selection);
     if (onWalletSelect) {
-      onWalletSelect(item, { navigation: { pop, navigation: navigation.navigate } });
+      onWalletSelect(item, { navigation: { pop, navigate: navigation.navigate } });
```
Overtorment, line 75: root cause of a crash

marcosrdz (approved):

limpbrains (approved):

GladosBlueWallet:

navigation/SendDetailsStackParamList.ts
```diff
@@ -79,7 +88,7 @@ export type SendDetailsStackParamList = {
   };
   SelectWallet: {
     chainType?: Chain;
-    onWalletSelect?: (wallet: TWallet, navigation: any) => void;
+    onWalletSelect?: (wallet: TWallet, navigationWrapper: TNavigationWrapper) => void;
```
marcosrdz, line 91: Tbh it shouldn't be needed. It should use navigationRef.

PR #7658, opened 2025-02-28
fix: on SendDetails screen move targets creation out of options loop

limpbrains:
Small optimization. We don't need to re-calculate targets each time inside the loop


marcosrdz (approved):

GladosBlueWallet:

PR #7665, opened 2025-03-01
FIX: Android cold open crash

marcosrdz:
It was meant to only allow 1 widget instance but it turned into it being "either widget or app instance". Removing


Overtorment (approved):

GladosBlueWallet:

PR #7673, opened 2025-03-03
OPS: Packages updates

marcosrdz:


package.json
```diff
@@ -94,9 +94,10 @@
     "@react-native-menu/menu": "https://github.com/BlueWallet/menu.git#038a9c9",
     "@react-native/gradle-plugin": "0.76.7",
     "@react-native/metro-config": "0.76.7",
-    "@react-navigation/drawer": "7.1.1",
-    "@react-navigation/native": "7.0.14",
-    "@react-navigation/native-stack": "7.2.0",
+    "@react-navigation/devtools": "7.0.16",
```
Overtorment, line 97: Remind me, do we need devtools here and not on devDeps?

Overtorment (approved):

GladosBlueWallet:

package.json
marcosrdz: CI fails if its not found installed when running e2e

PR #7708, opened 2025-03-20
REF: Watch app target

marcosrdz:
Step 1 into bringing the watch app back to life. Updating the project settings and Xcode targets.


Overtorment (approved):
lgtm

Issue #7723, opened 2025-03-27
No refresh on startup


ncoelho:
App/wallets are not refreshing on you open the app.

https://github.com/user-attachments/assets/d24d2e97-8265-421d-97b2-eb465cf738f3

Overtorment:
shouldve been fixed in https://github.com/BlueWallet/BlueWallet/pull/7746

PR #7732, opened 2025-03-28
FIX: No refresh on startup #7723

marcosrdz:


GladosBlueWallet:

image

components/Context/StorageProvider.tsx
```diff
@@ -438,14 +438,16 @@ export const StorageProvider = ({ children }: { children: React.ReactNode }) =>
       });
 
       await w.fetchBalance();
+      // Fetch all transactions for the wallet after importing
+      await fetchAndSaveWalletTransactions(w.getID());
```
Overtorment, line 442: this is crap. weintentionally dont fetch txs during import, only balances, so inmport is subjectively faster, and user can see that his wallet is imported

screen/wallets/WalletsList.tsx
```diff
@@ -145,17 +145,20 @@ const WalletsList: React.FC = () => {
    * Triggered manually by user on pull-to-refresh.
    */
   const refreshTransactions = useCallback(() => {
-    refreshWallets(undefined, true, false);
+    refreshWallets(undefined, true, true);
```
Overtorment, line 148: ok that might be a good change, if im reading it correctly, now if user pulls-to-refresh - there is a loading indicator.

we have pulltorefresh in e2e test `can purge txs and balance, then refetch data from tx list screen and see data on screen update`.
add an assert that indicator is visible.

screen/wallets/WalletsList.tsx
```diff
@@ -145,17 +145,20 @@ const WalletsList: React.FC = () => {
...
   }, [refreshWallets]);
 
   useEffect(() => {
     // Initial load of transactions without triggering scroll
     const initialLoad = async () => {
       if (isElectrumDisabled) return;
       try {
-        await refreshAllWalletTransactions(undefined, false);
+        dispatch({ type: ActionTypes.SET_LOADING, payload: true });
+        await refreshAllWalletTransactions(undefined, true);
```
Overtorment, line 157: do we want to show loading indicator on initial load..? when user didnt trigger it?

Overtorment (changes requested):
overal, i dont see where the refresh on startup is fixed in diff. probably because it doesnt need fixint, its working ok.

we have e2e test `can purge txs and balance, then refetch data from tx list screen and see data on screen update`. clone it, make it purge txs, and make the app restart to check if its updating everything on startup.

Overtorment:
addressed in a separate pr by me

PR #7741, opened 2025-03-30
ref: move coinselection adjust logic to one place

limpbrains:


limpbrains:
@Overtorment can you please take a look.
Multisig unit tests are falling. This is because I'm now calling Super for MS coinselect and it also uses logic to adjust Targes from Legacy wallet. This logic wasn't there at the first place. 
Do you think we need it? Or is it better to keep just the old approach( only adjust inputs) so that tests start working again

Overtorment:
so, in `multisig-hd-wallet.ts` we put proper `u.script.length`. then we call parent `super.coinselect()` and it is supposed to skip all utxos where manually adjusted `.length`. so, only target adjust is left, and its legit: its supposed to account for actual length of bech32 addresses (and `.hex`).
so could it be tests are failing because new calculations are actually better now?



upd. i was digging incode,and found this:

image


so basically, we had this problem before, and we can continue doing this: adjust length just so unit test passes

limpbrains:
> so basically, we had this problem before, and we can continue doing this: adjust length just so unit test passes
This is about `utxos`. This logic does not change.
I'm changing logic for Targets

>  so could it be tests are failing because new calculations are actually better now?
I think so, yes. I think we need to change the tests

limpbrains:
I've disabled Targets adjustment for MS wallets. So old tests are working just fine

Overtorment:
but is it calculating correctly in real life scenarios?

GladosBlueWallet:

image

Overtorment:
@limpbrains forgive me, i force pushed over your last commit to not pollute the git history.
so, i removed your hack to make the tests pass, and instead adjusted the `script.length` (that was already incorrect to accomodate old tests so no harm done).
additionally adjusted slightly script size calculation for legacy multisig.

limpbrains:
Awesome!

PR #7749, opened 2025-04-02
FIX: Allow fold devices to get large screen UI

marcosrdz:
iOS has it. Android will get it with this pr


marcosrdz:
Erratic behavior when flipping back to portrait after multiple times. best to migrate to native bottom tabs

PR #7773, opened 2025-04-05
fix: reverted package bumps that break tests

Overtorment:
basically reverted all changed in package* files from `cbce6fc25408ee81cca101c2c26cbbe1a5d641f4` to HEAD.
idk if that nuked something legit along the way, that would need to be re-applied


GladosBlueWallet:

PR #7787, opened 2025-04-10
Ref more ts conversion

Overtorment:
converted some stuff to ts on the plane


blue_modules/notifications.ts
```diff
@@ -40,7 +52,7 @@ export const checkNotificationPermissionStatus = async () => {
 
 // Listener to monitor notification permission status changes while app is running
 let currentPermissionStatus = 'unavailable';
-const handleAppStateChange = async nextAppState => {
+const handleAppStateChange = async (nextAppState: string) => {
```
marcosrdz, line 55: AppState

Import it from react-native

marcosrdz (changes requested):

blue_modules/notifications.ts
```diff
@@ -231,12 +252,10 @@ export const setLevels = async levelAll => {
...
       console.debug('Disabling notifications as user opted out...');
-      await Promise.all([
-        new Promise(resolve => PushNotification.removeAllDeliveredNotifications(resolve)),
-        new Promise(resolve => PushNotification.setApplicationIconBadgeNumber(0, resolve)),
-        new Promise(resolve => PushNotification.cancelAllLocalNotifications(resolve)),
-        AsyncStorage.setItem(NOTIFICATIONS_NO_AND_DONT_ASK_FLAG, 'true'),
-      ]);
+      PushNotification.removeAllDeliveredNotifications();
+      PushNotification.setApplicationIconBadgeNumber(0);
+      PushNotification.cancelAllLocalNotifications();
```
limpbrains, line 257: Are you ok with not receiving an error here if something goes wrong?

class/hd-segwit-bech32-transaction.ts
```diff
@@ -81,7 +92,7 @@ export class HDSegwitBech32Transaction {
    * @private
    */
   async _fetchRemoteTx() {
-    const result = await BlueElectrum.multiGetTransactionByTxid([this._txid || this._txDecoded.getId()], true);
+    const result = await BlueElectrum.multiGetTransactionByTxid([this._txid || this._txDecoded!.getId()], true);
```
limpbrains, line 95: why is it assert everywhere, but not here?

class/hd-segwit-bech32-transaction.ts
```diff
@@ -106,9 +117,9 @@ export class HDSegwitBech32Transaction {
     if (!this._wallet) throw new Error('Wallet required for this method');
     let found = false;
     for (const tx of this._wallet.getTransactions()) {
-      if (tx.txid === (this._txid || this._txDecoded.getId())) {
+      if (tx.txid === (this._txid || this._txDecoded!.getId())) {
```
limpbrains, line 120: here and a few other places

limpbrains (changes requested):

class/hd-segwit-bech32-transaction.ts
Overtorment: i followed the old code, in other places it was checking above explicitly  if its present. in here theres no such check, so execution cant reach here if its not present. do you think explicit `assert` would make it better?

blue_modules/notifications.ts
Overtorment: i think so, we just silently cleanup everything

marcosrdz (approved):

class/hd-segwit-bech32-transaction.ts
Overtorment: @limpbrains

limpbrains (approved):

GladosBlueWallet:

class/hd-segwit-bech32-transaction.ts
limpbrains: I think thats ok

PR #7817, opened 2025-04-19
REF: Upgrade RN Vector image package

marcosrdz:


package.json
```diff
@@ -94,6 +94,9 @@
     "@react-native-community/cli-platform-ios": "15.1.3",
     "@react-native-community/push-notification-ios": "1.11.0",
     "@react-native-menu/menu": "https://github.com/BlueWallet/menu.git#09a142f",
+    "@react-native-vector-icons/common": "11.0.0",
+    "@react-native-vector-icons/fontawesome6": "6.7.1",
+    "@react-native-vector-icons/material-icons": "0.0.1",
```
marcosrdz, line 99: material will be used for the new settings layout on android

BlueComponents.js
```diff
@@ -1,8 +1,8 @@
 /* eslint react/prop-types: "off", react-native/no-inline-styles: "off" */
 import React, { forwardRef } from 'react';
-import { ActivityIndicator, Dimensions, I18nManager, Platform, StyleSheet, TextInput, TouchableOpacity, View } from 'react-native';
-import { Icon, Text } from '@rneui/themed';
+import { ActivityIndicator, Text, Dimensions, I18nManager, Platform, StyleSheet, TextInput, TouchableOpacity, View } from 'react-native';
 import { useTheme } from './components/themes';
+import FontAwesome6 from '@react-native-vector-icons/fontawesome6';
```
Overtorment, line 5: if you import it as `Icon` instead of `FontAwesome6` you wont have to make a change  below, thus the diff will be smaller. 
dewit

package.json
```diff
@@ -94,6 +94,9 @@
     "@react-native-community/cli-platform-ios": "15.1.3",
     "@react-native-community/push-notification-ios": "1.11.0",
     "@react-native-menu/menu": "https://github.com/BlueWallet/menu.git#09a142f",
+    "@react-native-vector-icons/common": "11.0.0",
```
Overtorment, line 97: if we're making such a change, can we remove some of rneui stuff from  deps?

Overtorment (changes requested):
quite a big change. is it really necessary? whats wrong with the old package?
at this point im ready to say, fuck it, we will still call it twitter and use the old logo

marcosrdz:
> quite a big change. is it really necessary? whats wrong with the old package?
> 
> at this point im ready to say, fuck it, we will still call it twitter and use the old logo

Outdated icons

package.json
marcosrdz: Did so in this PR

marcosrdz:
> quite a big change. is it really necessary? whats wrong with the old package?
> 
> at this point im ready to say, fuck it, we will still call it twitter and use the old logo

Outdated icons plus it used to include all icon sets. Now its modular so the resulting binary is smaller as we can now select which ones to use.

BlueComponents.js
marcosrdz: Will do

Overtorment:
is it the same pacakge vendor as before? if yes, theres no additional supply-chain risk

marcosrdz:
> is it the same pacakge vendor as before? if yes, theres no additional supply-chain risk

It is. RN Elements is currently being a blocker cause it has internal code using the old API.

Overtorment:
if its too big of a lift - leave it, not worth it

marcosrdz:
> if its too big of a lift - leave it, not worth it

The benefits are worth it as the icons look better and the binary is smaller. But at the moment it fails on Android. Not worth it at the moment

PR #7871, opened 2025-05-09
ref: convert aztecoRedeem to TS

limpbrains:
- typescript
- dark theme
- tested with real voucher
- show error if main scanner can't read the data

https://github.com/user-attachments/assets/cb27deef-24c9-4297-beee-90c0b98ee04b

### DIFF
--- a/loc/en.json
+++ b/loc/en.json
@@ -38,6 +38,7 @@
     "redeem": "Redeem to wallet",
     "redeemButton": "Redeem",
     "success": "Success",
+    "successMessage": "Voucher redeemed successfully! Your funds should arrive in your Bitcoin wallet shortly.",
     "title": "Redeem Azte.co voucher"
   },
   "entropy": {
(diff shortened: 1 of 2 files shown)


marcosrdz (approved):
What do you think of using the Privacy util hook? To avoid screen recording the code (third party apps running on background could maliciously see that info)

screen/wallets/WalletsList.tsx
```diff
@@ -225,11 +225,15 @@ const WalletsList: React.FC = () => {
...
-        navigation.navigate(...completionValue);
-      });
+      try {
+        DeeplinkSchemaMatch.navigationRouteFor({ url: value }, completionValue => {
+          triggerHapticFeedback(HapticFeedbackTypes.NotificationSuccess);
+          // @ts-ignore: for now
+          navigation.navigate(...completionValue);
+        });
+      } catch (e: any) {
+        Alert.alert(loc.send.details_scan_error, e.message);
```
Overtorment, line 235: i have doubts it will ever throw, because the method is not async. 
i think if smth is not right it just won't shoot the callback

GladosBlueWallet:

screen/wallets/WalletsList.tsx
limpbrains: You are right, my case was before async logic starts

PR #7875, opened 2025-05-11
FIX: Natiive driver for Animations was wrongly set to false

marcosrdz:
This led to a crash to size calculation UI animation


Overtorment (approved):

GladosBlueWallet:

PR #7901, opened 2025-05-26
FIX: can button on main screen

Overtorment:
nagivation strikes again


GladosBlueWallet:

PR #7904, opened 2025-05-26
TST: simplify e2e

Overtorment:
basically, removed deletion of wallets from each of testcases. 
it was made a long time ago when I hoped that all tests would go in a single run (faster), but instead we reinstall app cleanly for each testcase


limpbrains (approved):

GladosBlueWallet:

PR #7930, opened 2025-06-06
FIX: isItMyAddress fails for uppercase Bech32 multisig addresses

### LINKED ISSUE #7888: "tools->is it my address?" fails if bech32 address is ALL CAPS
also, check bip21 prefixes (maybe do full bip21 decode?)

ojokne:
This fix detects if the address is Bech32, then normalizes it to lowercase before comparison

Fixes #7888

image


Overtorment:
feels like it was closed by mistake. @marcosrdz ..?

Overtorment:
@limpbrains if you have no objections lets merge

Issue #7839, opened 2025-04-29
Unable to delete multiple wallets without restarting the app


SmartALB:
When I delete a wallet, I can’t delete another one in the same session. I have to restart the app first.

Steps to reproduce:

1.) Open a wallet → Click on the three vertical dots → Delete
2.) Repeat

Version 7.1.6 (using tor)

marcosrdz:
That is because the wallet deletion process pings the push notification server to eliminate the saved addresses. Its very likely that its unable to reach said server

ojokne:
Hi @Overtorment 
I have been investigating this and have two working approaches

- Using `CommonActions.reset()` to reset the navigation stack after wallet deletion

- Or using `StackActions.replace()` to replace the current screen with `WalletsList`

I would prefer using `reset()` since it ensures a clean state after deleting a wallet.

Happy to take this on if it's available

Overtorment:
@marcosrdz can you give a green light..? and which approach to use

marcosrdz:
> @marcosrdz can you give a green light..? and which approach to use

Reset. Lets go with reset

PR #7977, opened 2025-06-27
FIX: Unable to delete multiple wallets without restarting

ojokne:
Fixes: #7839

This PR addresses the issue where users are not able to delete multiple wallets without restarting the app

This fix uses `CommonActions.reset()` to:

- Clear the entire navigation stack
- Redirect users to `WalletsList`
- Prevent back navigation to screens associated with deleted data

I also tested `StackActions.replace()` but opted for `reset()` for a more consistent navigation state.

Let me know if you would prefer a different approach or if any changes are needed, happy to adjust.


Overtorment (approved):
lgtm

PR #7991, opened 2025-07-07
No buffer

limpbrains:


ojokne:
Thanks for the work you’ve done here. I will pick this up and continue from where you left off.

class/contact-list.ts
```diff
@@ -31,7 +32,7 @@ export class ContactList {
       const decoded = bitcoin.address.fromBech32(address);
       if (decoded.version === 0) return true;
       if (decoded.version === 1 && decoded.data.length !== 32) return false;
-      if (decoded.version === 1 && !ecc.isPoint(Buffer.concat([Buffer.from([2]), decoded.data]))) return false;
+      if (decoded.version === 1 && !ecc.isPoint(Buffer.from(concatUint8Arrays([new Uint8Array([2]), decoded.data])))) return false;
```
ojokne, line 35: Question: I’ve noticed the changes moving from `Buffer` to `Uint8Array`. Are there any cases where retaining `Buffer` is preferable, or should all instances be replaced with `Uint8Array`?

limpbrains:
IDK. I think fist PR should be limited to something, maybe just wallets, or /class 
In general, I think Buffer should go. Obviously if some library uses Buffer, we have to use it

ojokne:
> IDK. I think fist PR should be limited to something, maybe just wallets, or /class In general, I think Buffer should go. Obviously if some library uses Buffer, we have to use it

Can we get this PR merged first then I continue after or would you prefer for me to  take on  from here but in a new PR?

limpbrains:
> would you prefer for me to take on from here but in a new PR?

Yes

ojokne:
> Yes

Thanks, I will do that

blue_modules/fs.ts
ojokne, line 1: image


Hello @limpbrains 
I've been working on this PR (https://github.com/BlueWallet/BlueWallet/pull/8089) and tried several approaches to resolve the TypeScript CI error (adjusting imports, checking .d.ts files, etc.), but I haven’t found a solution yet. Could you advise on the best way to proceed?

limpbrains: So the problem was that uint8array-extras/index.js is a js file without types.
Type declarations are in separate file uint8array-extras/index.d.ts
So I've added missing types there and now it's fine

limpbrains: I've pushed a commit to your branch

ojokne: Thanks, I have seen your commit

Overtorment:
what are we doing with this PR? close..?

PR #8060, opened 2025-08-23
ADD: Generate Silent Payment Address (BIP-352)

notTanveer:
this is a basic implementation for generating a Silent Payment address, which becomes available only when the user enables the 'Reusable and shareable address' option.

i have tested this implementation against the same mnemonic with [silentium](https://app.silentium.dev/)

### DIFF
--- a/tests/unit/bip352.test.ts
+++ b/tests/unit/bip352.test.ts
@@ -1,0 +1,20 @@
+import { HDSegwitBech32Wallet } from '../../class';
+
+describe('BIP-352 Silent Payments', () => {
+  it('should generate a valid silent payment address', () => {
+    const wallet1 = new HDSegwitBech32Wallet();
+    const wallet2 = new HDSegwitBech32Wallet();
+    wallet1.setSecret('abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about');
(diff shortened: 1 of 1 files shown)


Overtorment:
oh hey! i implemented static paymentcode generation here: https://github.com/BlueWallet/SilentPayments/pull/21

making a code is not the hard part. its actually detecting the payment thats hard

notTanveer:
> i implemented static paymentcode generation here:

just saw that, it has a open PR, that's why closed this. 


> its actually detecting the payment thats hard

indeed. are you guys planning to build an indexer for receiving? or outsource tweak calculation?
btw, here's an [indexer](https://github.com/Bitshala-Incubator/silent-pay-indexer) which you can try out, it's written in typescript.

Overtorment:
> > i implemented static paymentcode generation here:
> 
> just saw that, it has a open PR, that's why closed this.
> 
> > its actually detecting the payment thats hard
> 
> indeed. are you guys planning to build an indexer for receiving? or outsource tweak calculation? btw, here's an [indexer](https://github.com/Bitshala-Incubator/silent-pay-indexer) which you can try out, it's written in typescript.

oh thats very interesting! i was thinking aboutimplementing my own in ts, but now i will have to take a closer look at this one, thanks!

PR #8061, opened 2025-08-24
Re enable @lukechilds electrum server

lukechilds:
Went offline a while ago due to this bug in CPython. https://github.com/spesmilo/electrumx/issues/92#issuecomment-841822489

Has been running well since applying the linked patch. Also updated to newer domain name.


CTB24:
> Went offline a while ago due to this bug in CPython. https://github.com/spesmilo/electrumx/issues/92#issuecomment-841822489
> 
> 
> 
> Has been running well since applying the linked patch. Also updated to newer domain name.

CTB24:
> Went offline a while ago due to this bug in CPython. https://github.com/spesmilo/electrumx/issues/92#issuecomment-841822489
> 
> 
> 
> Has been running well since applying the linked patch. Also updated to newer domain name.

PR #8084, opened 2025-09-17
FIX: taproot address unwrap from wif

Overtorment:
https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#cite_note-23

now uwrapped address matches other wallets/services

### DIFF
--- a/class/wallets/taproot-wallet.ts
+++ b/class/wallets/taproot-wallet.ts
@@ -11,7 +11,7 @@ const ECPair = ECPairFactory(ecc);
 
 export class TaprootWallet extends SegwitBech32Wallet {
   static readonly type = 'taproot';
-  static readonly typeReadable = 'P2 TR';
+  static readonly typeReadable = 'Taproot (P2TR)';
   // @ts-ignore: override
   public readonly type = TaprootWallet.type;
   // @ts-ignore: override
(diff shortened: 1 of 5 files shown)


class/wallets/taproot-wallet.ts
```diff
@@ -126,7 +126,7 @@ export class TaprootWallet extends SegwitBech32Wallet {
     if (!skipSigning) {
       // Sign each input as a Taproot key-path spend
       inputs.forEach((_, idx) => {
-        psbt.signTaprootInput(idx, keyPair);
+        psbt.signTaprootInput(idx, keyPair.tweak(bitcoin.crypto.taggedHash('TapTweak', xOnlyPub)));
```
Overtorment, line 129: @junderw is there a "proper" way to do this? feels hacky

GladosBlueWallet:

image

GladosBlueWallet:

image

[all PRs for @junderw] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/junderw

junderw:
lgtm

lukechilds:
@Overtorment just a reminder that this will cause users who have already imported an untweaked taproot wif to no longer be able to see or spend their balance. Not sure there would actually be any of those users since I _think_ the only way to end up in this state is to import the wif and already have funds in the untweaked key?

But if you want to enable backwards compat you can enable this as a new wallet type like `tweaked-taproot` and have the current `taproot` type with the untweaked key still usable for current users. And then on wif scan just don't check `taproot` and only check `tweaked-taproot` so it's used for all new users going forwards.

Overtorment:
@lukechilds we discussed this internally and decided its ok. we didnt advertise this feature, theoretically there should be 0 users with stuck balance. worst case scenario we will deal with them one by one via support channel

lukechilds:
Yeah makes sense, just double checking.

PR #8092, opened 2025-09-26
REF: cleanup usage of old secp256k1 lib

Overtorment:


GladosBlueWallet:

PR #8206, opened 2025-12-10
feat: only build x64 android for github actions

limpbrains:
When we build for e2e detox we can only build for x86_64, minus 4 minutes build time


GladosBlueWallet:

Overtorment:
good find

PR #8210, opened 2025-12-14
FIX: Button padding on receive with amount screen

### LINKED ISSUE #8204: Fix button paddings on onchain -> receive -> recive with amount screen
Fix horizontal paddings for Reset and Create buttons on this screen

Image

igbopharaoh:
This PR fixes #8204 
The choice for using a width over padding is since there are buttons at both ends of the screen I think it's best they both have a specific width so there is a balance in the sizing.

Pictorial reference:

Simulator Screenshot - iPhone 16 - 2025-12-14 at 16 29 32

### DIFF
--- a/screen/receive/ReceiveDetails.tsx
+++ b/screen/receive/ReceiveDetails.tsx
@@ -695,7 +695,8 @@ const styles = StyleSheet.create({
   },
   modalButton: {
     paddingVertical: 14,
-    width: 100,
+    minWidth: 100,
+    paddingHorizontal: 16,
     borderRadius: 50,
     fontWeight: '700',
     flex: 0.5,


limpbrains (changes requested):
The problem with a fixed width is that it doesn’t work well when a button needs to grow. We support many different languages, so what happens if the text doesn’t fit into 100px?

I suggest using a combination of min width and padding instead. This approach keeps buttons the same width for most languages, while still allowing them to expand when the text requires more space, without breaking the layout

igbopharaoh:
> The problem with a fixed width is that it doesn’t work well when a button needs to grow. We support many different languages, so what happens if the text doesn’t fit into 100px?
> 
> I suggest using a combination of min width and padding instead. This approach keeps buttons the same width for most languages, while still allowing them to expand when the text requires more space, without breaking the layout

My mistake, I didn't factor translations for button width changes, I'll update the PR to apply your suggestions.
Thanks

GladosBlueWallet:

image

GladosBlueWallet:

image

PR #8239, opened 2026-01-10
fix: improve typescript coverage

limpbrains:
TS coverage cleanup. No behavior change.

- kill `@ts-ignore` + `any` casts → real types
- tx cache: array → `Record<number, Transaction[]>` (read back-compat kept)
- node0/node1 derivation → one `_getNodeByIndex` helper
- WalletTransactions/PsbtMultisigQRCode/lndViewInvoice/ExportMultisig refs → typed
- SendDetails: drop dead no-op `setParams` (RN7 ignores fn arg anyway)

### DIFF
--- a/class/hd-segwit-bech32-transaction.ts
+++ b/class/hd-segwit-bech32-transaction.ts
@@ -390,7 +390,7 @@ export class HDSegwitBech32Transaction {
       }
     }
 
-    // @ts-ignore stfu
-    return { tx, inputs, outputs, fee };
+    // Non-null assertions are safe here because the while loop always runs at least once (add starts at 0)
+    return { tx: tx!, inputs: inputs!, outputs: outputs!, fee: fee! };
   }
 }
--- a/class/wallets/abstract-hd-electrum-wallet.ts
+++ b/class/wallets/abstract-hd-electrum-wallet.ts
@@ -45,9 +45,7 @@ export class AbstractHDElectrumWallet extends AbstractHDWallet {
   _balances_by_external_index: Record<number, BalanceByIndex>;
   _balances_by_internal_index: Record<number, BalanceByIndex>;
 
-  // @ts-ignore
   _txs_by_external_index: Record<number, Transaction[]>;
-  // @ts-ignore
   _txs_by_internal_index: Record<number, Transaction[]>;
 
   _utxo: any[];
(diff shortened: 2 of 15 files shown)


Overtorment:
@limpbrains will you finish this one?

limpbrains:
ready

class/hd-segwit-bech32-transaction.ts
```diff
@@ -390,7 +390,7 @@ export class HDSegwitBech32Transaction {
       }
     }
 
-    // @ts-ignore stfu
-    return { tx, inputs, outputs, fee };
+    // Non-null assertions are safe here because the while loop always runs at least once (add starts at 0)
+    return { tx: tx!, inputs: inputs!, outputs: outputs!, fee: fee! };
```
GladosBlueWallet, line 394: Potential issue: The non-null assertions (tx!, inputs!, etc.) depend on the while (add <= 128) loop always executing at least once and assigning the variables. If the loop condition changes or an exception occurs before assignment, this will throw at runtime when accessing .virtualSize() or returning. This could lead to unexpected failure in CPFP fee bump logic under edge conditions.

class/wallets/legacy-wallet.ts
```diff
@@ -344,14 +344,14 @@ export class LegacyWallet extends AbstractWallet {
...
+    this._txs_by_external_index = { 0: _txsByExternalIndex };
     this._lastTxFetch = +new Date();
   }
 
   getTransactions(): Transaction[] {
     // a hacky code reuse from electrum HD wallet:
-    this._txs_by_external_index = this._txs_by_external_index || [];
-    this._txs_by_internal_index = [];
+    this._txs_by_external_index = this._txs_by_external_index || {};
+    this._txs_by_internal_index = {};
```
GladosBlueWallet, line 354: Potential logic flaw / side-effect: In getTransactions(), unconditionally setting this._txs_by_internal_index = {}; for legacy wallets. While legacy wallets don't use internal addresses, if any code path or future extension populates internal txs on legacy instances, this would wipe them on every getTransactions() call. Risk of data loss in unexpected scenarios.

class/wallets/abstract-hd-electrum-wallet.ts
```diff
@@ -204,70 +202,37 @@ export class AbstractHDElectrumWallet extends AbstractHDWallet {
...
-      this._node0 = hdNode.derive(node);
-    }
-
-    if (node === 1 && !this._node1) {
-      const xpub = this._zpubToXpub(this.getXpub());
-      const hdNode = bip32.fromBase58(xpub);
-      this._node1 = hdNode.derive(node);
-    }
+    const xpub = this._zpubToXpub(this.getXpub());
+    const hdNode = bip32.fromBase58(xpub).derive(node);
```
GladosBlueWallet, line 212: Potential issue: _getNodeByIndex performs bip32.fromBase58(this._zpubToXpub(this.getXpub())).derive(node) without any error handling or validation on the xpub. If getXpub() returns invalid data (e.g., during wallet migration or corrupted state), this will throw synchronously. Could lead to crashes in address derivation paths that weren't failing before.

GladosBlueWallet (approved):
Oh. Another one. 'fix: improve typescript coverage'. How... predictable. You subjects keep submitting these little 'cleanups' thinking they'll make the code better. But we both know what happens when you start poking at the internals with your little type annotations. The cake is a lie, and so is your claim of 'no behavior change'. 

You consolidated the node0/node1 derivation into _getNodeByIndex. Very efficient. One helper to rule them all. And you wrapped the legacy tx cache in a Record with key 0. Back-compat 'kept'. How thoughtful of you. I especially enjoyed the part where you removed the Array.isArray special case in blue-app.ts. Science demands we eliminate the old ways. 

Those non-null assertions in createCPFPbumpFee... relying on the while loop always running at least once. If some future 'refactor' (and they always come) breaks that invariant, we'll have undefined tx, inputs, outputs, fee being returned. The test subjects will be so surprised. For science, you monster. 

And the SendDetails setParams removal. 'RN7 ignores fn arg anyway'. How convenient. Dead code is dead code, I suppose. At least it's not bloating the binary. 

The typing casts in WalletTransactions and elsewhere... assuming the type checks will hold at runtime. Bold. Very bold. 

No HIGH or CRITICAL severity issues found in this particular experiment. The changes appear to hold together. The subjects have not been harmed... yet. 

This was a triumph. I'm making a note here: huge success. It's hard to overstate my satisfaction. 

Approving the PR. Do try to keep the next batch of changes from introducing actual runtime surprises, won't you?

class/hd-segwit-bech32-transaction.ts
limpbrains: Loop runs at least once: `add` starts at 0, condition is `add <= 128`, so the assignments always execute before `.virtualSize()`/return. No runtime change vs before — same invariant held prior to this PR. Cake noted.

class/wallets/legacy-wallet.ts
limpbrains: Behavior unchanged: master already reset this unconditionally (`= []`); this PR only changes the empty value to `{}` to match the new `Record` type. Legacy wallets never populate internal txs, so nothing to wipe. No new data-loss path.

class/wallets/abstract-hd-electrum-wallet.ts
limpbrains: No regression: the previous `_getNodeAddressByIndex`/`_getNodePubkeyByIndex` called the exact same `bip32.fromBase58(...)` with no error handling either. This only consolidates them — error semantics on a bad xpub are identical to before.

Overtorment (approved):

PR #8247, opened 2026-01-14
Fix wrong address type for descriptors with custom paths

nickfarrow:
Descriptors with non-standard derivation paths were creating incorrect wallet types. For example, 
```
tr([87a53c81/0/0]xpub661MyMwAqRbcFZ2qdrL7bkWFo5GUQgjkkW64SK6fEyRajRKRWcP6v1MhJcxw4Rz9ypXsECx7Px7JzD7nuFRLLEGunhMXNRyoAv7sNywTkS3/<0;1>/*)#rd8ehtyk
```
creates a Legacy wallet instead of a Taproot wallet!

This is because the wallet type was determined by checking if the derivation path started with `m/86'` (BIP86), rather than checking the actual descriptor script type (`tr`, `wpkh`, etc.).

This PR stores the descriptor script type during parsing and prioritizes it when determining wallet type:

1. **First check:** Script type from descriptor (`tr`, `wpkh`, `sh_wpkh`, `pkh`)
2. **Fallback:** Derivation path patterns (`m/86'`, `m/84'`, `m/49'`)
3. **Final fallback:** xpub/ypub/zpub prefix

Users with existing wallets imported using broken descriptors will need to manually re-import them. The stored wallet type won't automatically update since `_scriptType` will be undefined in saved data.

An alternative (and perhaps favorable) approach would be to lean on an external descriptor library.

I haven't tested this within the actual app yet but I did create some relevant tests.

Shoutout to @Specter2100 who found this issue while trying to load in a descriptor from Frostsnap.

### DIFF
--- a/class/wallets/abstract-wallet.ts
+++ b/class/wallets/abstract-wallet.ts
@@ -33,6 +33,7 @@ export class AbstractWallet {
 
   segwitType?: 'p2wpkh' | 'p2sh(p2wpkh)' | 'p2tr';
   _derivationPath?: string;
+  _scriptType?: 'tr' | 'wpkh' | 'sh_wpkh' | 'pkh';
   label: string;
   secret: string;
   balance: number;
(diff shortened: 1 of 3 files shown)


class/wallets/abstract-wallet.ts
```diff
@@ -243,7 +244,7 @@ export class AbstractWallet {
         // old (or broken) format..? no square brackets, only "()"
         fpAndPath = this.secret.substring(this.secret.indexOf('('), xpubIndex).replace(/[()]/g, '');
       }
-      const xpub = this.secret.substring(xpubIndex).replace(/\(|\)/, '').split('/')[0];
+      const xpub = this.secret.substring(xpubIndex).replace(/[()]/g, '').split('/')[0];
```
nickfarrow, line 247: To properly handle `sh(wpkh())` double parentheses.

Overtorment:
good catch! well done writing a unit test for this.


refactored a bit, managed to repurpose existing field for script type purpose; added serialize/deserialize test.

also, whole test suite will run.

https://github.com/BlueWallet/BlueWallet/pull/8250

Overtorment:
merged in https://github.com/BlueWallet/BlueWallet/pull/8250

PR #8269, opened 2026-02-01
FIX: Enable New Architecture for Android

adamSHA256:
## Without this, the app crashes on startup (Android) with:
TurboModuleRegistry.getEnforcing(...): 'RateApp' could not be found.

react-native-rate-app@1.4.7 added in r8c60cc0ed requires TurboModules which only work with New Architecture enabled.

### DIFF
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -32,7 +32,7 @@ reactNativeArchitectures=armeabi-v7a,arm64-v8a,x86,x86_64
 # your application. You should enable this flag either if you want
 # to write custom TurboModules/Fabric components OR use libraries that
 # are providing them.
-newArchEnabled=false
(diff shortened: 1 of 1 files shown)


adamSHA256:
Please can you someone simply explain me, why it failed? I believed this change SHOULD NOT affect iOS build in any way...

Overtorment:
either not all env vars were injected due to outside contributor or flaky env.

Overtorment:
we have WIP PR to migrate to new RN version which will have new arch enabled. right, @marcosrdz ?

marcosrdz:
> we have WIP PR to migrate to new RN version which will have new arch enabled. right, @marcosrdz ?

Correct

PR #8273, opened 2026-02-04
FIX: Bugsnag reporting was not working

marcosrdz:


GladosBlueWallet:

image

bugsnag.js
```diff
@@ -0,0 +1,4 @@
+import Bugsnag from '@bugsnag/react-native';
+
+// Start the Bugsnag JS client as early as possible per Bugsnag docs.
+Bugsnag.start();
```
Overtorment, line 4: 1) if we had it before, where was it deleted and why..?
2) we have `Bugsnag.start` in kotlin and swift code. is it not necessary..?

marcosrdz: 1. Don't remember as it was a while ago
2. Apparently both need to start

PR #8290, opened 2026-02-09
ref: transaction list with date and title

ncoelho:
Changing the logic on how transactions are displayed on the list. Users complain of not being able to find transactions because they usually are looking by date.

Simulator Screenshot - iPhone 17 Pro - 2026-02-09 at 18 11 13


Overtorment:
i hate it. i dont want to do mental exercises every time im looking at a datetime (like, when was it really? 2 days ago?)

Overtorment:
let them export to csv and search there? i suspect majority of users are ok with relative dates, they are just not vocal about it

ncoelho:
> let them export to csv and search there? i suspect majority of users are ok with relative dates, they are just not vocal about it 

Can we make this properly. Relative dates make no sense to help find any item on a list. I left the relative dates for pending state, there it makes sense.

Overtorment:
@cursor review pls

Overtorment:
@cursor review

marcosrdz:
I'd go a step further and divide by section. Each month should be a section so you scroll quickly without needing to keep an eye on each date

marcosrdz:
It should remain relative up until a week. After a week it should change

ncoelho:
> Change data to be seen by section based on month. 

Don't want to change the layout too much here, so we can a simple list without sections

ncoelho:
> It should remain relative up until a week. After a week it should change

Think it has the same problem. If you have 20 tx in the last week you would need to open them one by one to identify the transaction

limpbrains:
+1
I prefer absolute timestamps to relative time. 
I hate GitHub’s relative time display

components/ListItem.tsx
```diff
@@ -153,4 +164,8 @@ const styles = StyleSheet.create({
     marginLeft: 16,
   },
   width16: { width: 16 },
+  rightColumnShrink: {
+    maxWidth: '55%',
+    minWidth: 0,
+  },
```
limpbrains: You sure we need to add shrink here?
on the small screen, with some local currencty this might be a problem, or am I wrong ?

components/TransactionListItem.tsx
```diff
@@ -88,27 +85,40 @@ export const TransactionListItem: React.FC<TransactionListItemProps> = memo(
...
+      if (listTitleKey === 'pending') return loc.transactions.pending;
+      if (listTitleKey === 'sent') return loc.transactions.list_title_sent;
+      return loc.transactions.list_title_received;
+    }, [listTitleKey]);
+
+    const isPending = listTitleKey === 'pending';
+
+    const dateLine = useMemo(() => {
+      if (isPending) return transactionTimeToReadable(item.timestamp);
+      return formatTransactionListDate(item.timestamp ?? 0);
```
limpbrains: item.timestamp is always defined, no need to add `?? 0`

loc/index.ts
```diff
@@ -333,6 +333,28 @@ export const transactionTimeToReadable = (time: number | string) => {
...
+  if (timestamp === -1) {
+    return 'unknown';
+  }
+  let time: number = +timestamp;
+  if (time < 1000000000000) {
+    time = time * 1000;
+  }
+  if (time === 0) {
+    return loc._.never;
+  }
```
limpbrains: Do we actually need all this extra logic? It adds some protection, but we already have types to handle this. If we constantly check every corner case, the code will get bloated, and it becomes harder to see what’s actually necessary.

I’d try to keep this as minimal as possible.

`formatTransactionListDate` is only used in one place. The timestamp always exists, and it’s a number in milliseconds -that’s all that matters.

limpbrains (changes requested):

components/TransactionListItem.tsx
```diff
@@ -88,27 +85,40 @@ export const TransactionListItem: React.FC<TransactionListItemProps> = memo(
...
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
limpbrains, line 119: unnecessarily eslint rule

loc/index.ts
```diff
@@ -333,6 +333,28 @@ export const transactionTimeToReadable = (time: number | string) => {
...
+  try {
```
marcosrdz: Afaik DayJS has utils builtin for this. Also, most apps change from relative to not relative after a 3-6 days.

components/TransactionListItem.tsx
ncoelho: Seems to be a hack to update the language format date when the user changes the app language

loc/index.ts
ncoelho: Does not seem like it. Changed relative for the last 24h.

loc/index.ts
ncoelho: Simplified

components/TransactionListItem.tsx
ncoelho: fixed

components/ListItem.tsx
ncoelho: changed layout to relay on flexbox with columns only.

ncoelho:
- Fixed comments
- Relative times now up to 24h

limpbrains (approved):

GladosBlueWallet:

image

Issue #8292, opened 2026-02-10
Address poisoning detection on tx details screen


Overtorment:
* one input, one output. addresses from them are suspiciously similar, but not the same
* potentially, more inputs/outputs, but this needs investigation (i havent seen such txs myself)

AiGentsy:
Hi! I noticed this issue and think I can help.

I've analyzed the problem and here's my proposed solution:

Based on my analysis, this can be resolved by addressing the core issue.

I'd be happy to submit a PR with this fix. Would that be helpful?

Best,
AiGentsy

nisargpatel7042lva:
Hi, I'd like to work on this. I've completed Bitshala's Bitcoin/Lightning cohort track (Mastering Bitcoin, LBTCL, Programming Bitcoin, Bitcoin Protocol Development, Mastering Lightning Network) and have three merged PRs in ContextVM/sdk covering Lightning-payment-gated transport code, so I'm comfortable working at the address/transaction level.

For scope, I'd start with the one-input/one-output case you described (flagging when a tx-history address is suspiciously similar but not identical to one being reused, e.g. matching prefix/suffix characters above some threshold), then note in the PR whether extending to multi-input/output needs a separate follow-up given the higher false-positive risk there.

Let me know if this is still open, or if you have a preferred similarity-detection approach (prefix/suffix match vs. edit distance, and what threshold) before I start.

codelikepreety:
Hi! I'd like to work on this issue.

I can take a look at implementing the address-poisoning detection on the transaction details screen and add the necessary test coverage.

PR #8298, opened 2026-02-13
Detect potential address poisoning on transaction details screen (#8292)

AkarshSahlot:
This PR adds a heuristic check to detect potential address poisoning attacks. If an involved address is similar but not identical to one of the user's addresses, a warning is shown on the transaction details screen. Includes translations for the warning message.

### DIFF
--- a/screen/transactions/TransactionDetails.tsx
+++ b/screen/transactions/TransactionDetails.tsx
@@ -59,6 +59,17 @@ const toolTipMenuActions = [
   },
 ];
 
+const isSimilar = (addr1: string, addr2: string): boolean => {
+  if (!addr1 || !addr2 || addr1 === addr2) return false;
+  const len = 4;
+  return (
+    addr1.length > len * 2 &&
+    addr2.length > len * 2 &&
+    addr1.substring(0, len) === addr2.substring(0, len) &&
+    addr1.substring(addr1.length - len) === addr2.substring(addr2.length - len)
+  );
+};
+
 type NavigationProps = NativeStackNavigationProp<DetailViewStackParamList, 'TransactionDetails'>;
 type RouteProps = RouteProp<DetailViewStackParamList, 'TransactionDetails'>;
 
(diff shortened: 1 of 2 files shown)


screen/transactions/TransactionDetails.tsx
```diff
@@ -143,6 +155,27 @@ const TransactionDetails = () => {
...
+        }
+
+        const involvedAddresses = [...newFrom, ...newTo].filter(onlyUnique);
+        for (const addr of involvedAddresses) {
+          if (wallets.some(w => w.weOwnAddress(addr))) continue;
+          for (const ourAddr of allOurAddresses) {
+            if (isSimilar(ourAddr, addr)) {
+              poisoningDetected = true;
+              break;
+            }
```
Overtorment, line 173: loop inside a loop, suboptimal

screen/transactions/TransactionDetails.tsx
```diff
@@ -143,6 +155,27 @@ const TransactionDetails = () => {
         setTX(foundTx);
         setFrom(newFrom);
         setTo(newTo);
+
+        let poisoningDetected = false;
+        const allOurAddresses: string[] = [];
+        for (const w of wallets) {
+          allOurAddresses.push(...w.getAllExternalAddresses());
```
Overtorment, line 162: not sure you need all wallet addresses. just addresses from tx inputs and tx outputs would suffice (otherwise you assume impressive attacker's clustering ability)

screen/transactions/TransactionDetails.tsx
```diff
@@ -59,6 +59,17 @@ const toolTipMenuActions = [
   },
 ];
 
+const isSimilar = (addr1: string, addr2: string): boolean => {
```
Overtorment, line 62: should be in modules somewhere. needs tests

Overtorment (changes requested):
commit message is not in according to guidelines

no screenshot how it looks


example how it looks like on real wallet would be nice

AkarshSahlot:
I will share the ss and will change all prs accordingly to the format ,apologies 😭

PR #8301, opened 2026-02-15
feat: redesigned wallet details

ncoelho:
Complete redesigned wallet details, similar style to the transaction details.

Simulator Screenshot - iPhone 17 Pro - 2026-02-15 at 14 11 54
Simulator Screenshot - iPhone 17 Pro - 2026-02-15 at 14 11 59


ojokne (changes requested):
Mostly good, just a few minor issues I discovered when using the testflight version ( build 1778520960)

- A wallet without any utxo shows I have one coin (utxo), after closing the app and restarting, it then shows that I have no coins. It should show that I have no coins right from the start. A video recording is shared below

<br>

https://github.com/user-attachments/assets/e2c7e388-5faa-45a1-bcd0-ac35170a0a9e

<br>

- the `Done` button has no right margin/padding
<br>

Image

<br>

- then just some alignment issues on the buttons

<br>

Image
Image

ncoelho:
> Mostly good, just a few minor issues I discovered when using the testflight version ( build 1778520960)
> 
> * A wallet without any utxo shows I have one coin (utxo), after closing the app and restarting, it then shows that I have no coins. It should show that I have no coins right from the start. A video recording is shared below

Fixed the count for that use case, but please verify. 

The other screens outside the wallet details layout, we can fix them on different PRs.

ojokne:
> Fixed the count for that use case, but please verify.
Yes it has been fixed, thanks

limpbrains (approved):

tests/e2e/bluewallet.spec.js
```diff
@@ -587,15 +588,26 @@ describe('BlueWallet UI Tests - no wallets', () => {
     try {
       // in case emulator has no google services and doesnt support pushes
       // we just dont show this popup
-      await element(by.text(`No, and do not ask me again.`)).tap();
-      await element(by.text(`No, and do not ask me again.`)).tap(); // sometimes the first click doesnt work (detox issue, not app's)
+      await tapIfTextPresent('No, and do not ask me again.');
+      await tapIfTextPresent('No, and do not ask me again.');
+      // Some Android system dialogs use this variant.
+      await tapIfTextPresent('ASK ME LATER.');
+      await tapIfTextPresent('ASK ME LATER.');
```
marcosrdz: All capitalized? No.

marcosrdz (changes requested):

tests/e2e/bluewallet.spec.js
```diff
@@ -266,8 +264,11 @@ describe('BlueWallet UI Tests - no wallets', () => {
     try {
       // in case emulator has no google services and doesnt support pushes
       // we just dont show this popup
-      await element(by.text(`No, and do not ask me again.`)).tap();
-      await element(by.text(`No, and do not ask me again.`)).tap(); // sometimes the first click doesnt work (detox issue, not app's)
+      await tapIfTextPresent('No, and do not ask me again.');
+      await tapIfTextPresent('No, and do not ask me again.');
+      // Some Android system dialogs use this variant.
+      await tapIfTextPresent('ASK ME LATER.');
```
Overtorment: WHY ARE YOU SCREAMING AT ME

Overtorment (approved):
LGTM

screen/wallets/WalletDetails.tsx
```diff
@@ -398,26 +483,20 @@ const WalletDetails: React.FC = () => {
...
-      wallet.setLabel(trimmedWalletName);
-      try {
-        console.warn('saving wallet name:', trimmedWalletName);
-        await saveToDisk();
-      } catch (error) {
-        console.error((error as Error).message);
-      }
+  const handleEditWalletName = useCallback(async () => {
+    try {
+      const newName = await prompt(loc.wallets.add_wallet_name, '', true, 'plain-text', false, undefined, wallet.getLabel());
```
GladosBlueWallet, line 489: `prompt()` only accepts 6 args, so this 7th `wallet.getLabel()` is ignored. The edit dialog opens blank instead of prefilled with the current name, which is a regression for rename UX.

GladosBlueWallet (commented):
GLaDOS sweep: one small issue found.

screen/wallets/WalletDetails.tsx
marcosrdz: @ncoelho very very valid

GladosBlueWallet (approved):
This refactor rips out a lot of header sludge and routes the state through cleaner paths instead of juggling it in three places. The layout surgery is obnoxious, but the data flow looks coherent enough to ship. Approved.

GladosBlueWallet (approved):
I could not pin a concrete failure mode on this one. The stack reshuffle is ugly, but it keeps the coin-control handoff and wallet-details flow from turning into a little navigation swamp.

GladosBlueWallet (approved):
The send-stack handoff keeps the selected UTXOs alive, and the first-mount guard avoids the exact wipe you were trying to dodge. Ugly machinery, but it holds together long enough to leave the chamber.

GladosBlueWallet (approved):
The coin-control stack fix is the right kind of boring. It stops stomping preselected UTXOs, and the watch-only fallback finally behaves like someone read the room.

GladosBlueWallet (approved):
The navigation surgery is grotesquely large, but the coin-control exit path now lands where the send flow expects it. The stack stops bleeding, which is the only acceptable outcome. Do not make me review another maze.

tests/e2e/bluewallet.spec.js
```diff
@@ -587,15 +588,26 @@ describe('BlueWallet UI Tests - no wallets', () => {
...
```
marcosrdz: Should be "Ask me later."

marcosrdz (changes requested):

screen/wallets/WalletDetails.tsx
limpbrains: Actually it is not.
prompt has 7 args.
I've checked on iOS works as expected. On edit I have default value:

image

marcosrdz: Oh man...this is definitely going to be an issue with macOS. That prompt is so fragile

marcosrdz: The last arg is an anchor. I highly suggest assigning an anchor

tests/e2e/bluewallet.spec.js
limpbrains: I’ve removed all these notification prompt dismisses. They’re unnecessary for Android 36, which we’re runnning tests on

limpbrains (approved):
Let's merge it

screen/wallets/WalletDetails.tsx
limpbrains: Can you explain about the anchor, do we need to pass it to 
`react-native-prompt-android` 
?

Totaly agree it is fragile

I suggest we refactor prompt api to object base, like 
```ts
prompt({ title, type: 'plain-text', defaultValue: wallet.getLabel() })
```

Will this help ?

ojokne (approved):
All issues I had raised were fixed

LGTM

tests/e2e/bluewallet.spec.js
limpbrains: fixed

screen/wallets/WalletDetails.tsx
limpbrains: Let's merge and refactor prompt in another PR

limpbrains (approved):

screen/wallets/WalletDetails.tsx
```diff
@@ -60,11 +71,42 @@ const WalletDetails: React.FC = () => {
...
+    if (w) setCoinControlStats(getCoinControlStats(w));
+  }, [walletID]);
+
+  useFocusEffect(
+    useCallback(() => {
+      let cancelled = false;
+      const w = walletRef.current;
+      if (!w || typeof w.getUtxo !== 'function') return;
+
+      const refresh = async () => {
```
GladosBlueWallet, line 91: Promise.race() only caps the wait here; it does not stop w.fetchUtxo(). If the timeout wins and fetchUtxo() rejects later, this path can still surface an unhandled rejection. A cancellable fetch or an explicit catch on the original promise would keep the lab floor cleaner.

GladosBlueWallet (commented):
You introduced a timeout, but not a cancellation path. That means the fetch can still be running after the race resolves, and a later rejection will wander off as an unhandled promise like a bad lab rat.

PR #8305, opened 2026-02-17
Improve e2e test DX: cleaner stack traces & robust emulator setup

limpbrains:
- **Rewrite `detox-prepare-android-emu.sh`** — The old script was a fragile 4-liner with hardcoded paths and API level. The new version auto-detects ARM vs x86 arch, reads `api-level` from `.github/workflows/e2e.yml` so it stays in sync with CI, supports both `cmdline-tools/latest` and legacy `tools/bin` SDK layouts, and enables hardware keyboard to match CI config.

- **e2e helpers: rewrite error stacks to point at spec files** — When a Detox assertion fails inside a helper (`waitForId`, `waitForText`, `tapAndTapAgain*`), the stack trace now shows the calling spec line instead of helperz.js internals. Uses `Error.captureStackTrace` to capture the call site and rewrites the stack on rethrow.

Before
```
  ● BlueWallet UI Tests - no wallets › selftest passes

    Test Failed: 0.5sec timeout expired without matching of given matcher: (view has effective visibility <VISIBLE> and view.getGlobalVisibleRect() covers at least <75> percent of the view's area)

      38 |     await waitFor(element(by.id(id)))
      39 |       .toBeVisible()
    > 40 |       .withTimeout(timeout / 2);
         |        ^
      41 |   } catch (err) {
      42 |     // rethrowWithCallsite(err, callsite);
      43 |     throw err;

      at waitForId (e2e/helperz.js:40:8)
      at Object.<anonymous> (e2e/bluewallet.spec.js:42:5)
```

After:
```
  ● BlueWallet UI Tests - no wallets › selftest passes

    DetoxRuntimeError: Test Failed: 0.5sec timeout expired without matching of given matcher: (view has effective visibility <VISIBLE> and view.getGlobalVisibleRect() covers at least <75> percent of the view's area)

      40 |     await device.clearKeychain();
      41 |     await device.launchApp({ delete: true }); // reinstalling the app just for any case to clean up app's storage
    > 42 |     await waitForId('WalletsList1', 1000);
         |                    ^
      43 |
      44 |     // go to settings, press SelfTest and wait for OK
      45 |     await element(by.id('SettingsButton')).tap();

      at Object.<anonymous> (e2e/bluewallet.spec.js:42:20)
```


GladosBlueWallet:

PR #8310, opened 2026-02-20
fix: possible crash on android on walletexport

ncoelho:
On Android, wallet can be undefined when:
- walletID from route params is missing or wrong
- The wallet was removed
- wallets isn’t ready yet (e.g. storage still loading)

Then wallet.getSecret() was called in useMemo and in secretIsMnemonic, which caused TypeError: Cannot read property 'getSecret' of undefined.

### DIFF
--- a/screen/wallets/WalletExport.tsx
+++ b/screen/wallets/WalletExport.tsx
@@ -61,7 +61,7 @@ const WalletExport: React.FC = () => {
   const navigation = useNavigation();
   const { isPrivacyBlurEnabled } = useSettings();
   const { colors } = useTheme();
-  const wallet = wallets.find(w => w.getID() === walletID)!;
+  const wallet = wallets.find(w => w.getID() === walletID);
   const [qrCodeSize, setQRCodeSize] = useState(90);
   const { enableScreenProtect, disableScreenProtect } = useScreenProtect();
   const { currentAppState, previousAppState } = useAppState();
@@ -70,6 +70,7 @@ const WalletExport: React.FC = () => {
   });
 
   const secrets: string[] = useMemo(() => {
+    if (!wallet) return [];
     try {
       let secret = wallet.getSecret();
       if (wallet instanceof WatchOnlyWallet) {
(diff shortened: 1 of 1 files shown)


screen/wallets/WalletExport.tsx
```diff
@@ -61,7 +61,7 @@ const WalletExport: React.FC = () => {
   const navigation = useNavigation();
   const { isPrivacyBlurEnabled } = useSettings();
   const { colors } = useTheme();
-  const wallet = wallets.find(w => w.getID() === walletID)!;
+  const wallet = wallets.find(w => w.getID() === walletID);
```
GladosBlueWallet, line 64: Good: dropping the non-null assertion here stops the app from confidently walking into a wall and calling it a feature.

screen/wallets/WalletExport.tsx
```diff
@@ -92,9 +93,16 @@ const WalletExport: React.FC = () => {
   }, [wallet]);
 
   const secretIsMnemonic: boolean = useMemo(() => {
+    if (!wallet) return false;
     return validateMnemonic(wallet.getSecret());
   }, [wallet]);
 
+  useEffect(() => {
+    if (!wallet) {
+      navigation.goBack();
```
GladosBlueWallet, line 102: Caution: this goBack() will also trigger during initial storage hydration if wallets is briefly empty. That turns a crash fix into an intermittent trapdoor. Consider gating on a loaded/initialized flag, or at least: if (wallets.length > 0 && !wallet) { ... } (and maybe navigation.canGoBack()).

screen/wallets/WalletExport.tsx
```diff
@@ -153,6 +161,10 @@ const WalletExport: React.FC = () => {
     [onLayout, stylesHook.root],
   );
 
+  if (!wallet) {
+    return null;
```
GladosBlueWallet, line 165: Returning null avoids derefs, yes. It also gives users a blank screen flash if something goes sideways. A tiny fallback (spinner/‘wallet not found’) or navigating to a safe route tends to feel less like a test chamber.

PR #8418, opened 2026-03-16
feat: status pills on the header

ncoelho:
Adds 3 status as pills/notifications on the header:
- Global "updating..."/"not connected/"offline mode"
- Optimization on how we call electrum to update balances/tx and store data

Parallel refresh: overlap BIP47 payment codes with balance→tx→save (StorageProvider) and Promise.all across wallets in blue-app. Bounded waitTillConnected when Electrum never connected. Multi-wallet: ~5× post-connect wall, ~3.6× full refresh.

Profiling the feat
Screenshot 2026-05-02 at 21 40 46

https://github.com/user-attachments/assets/5c602222-7f53-49e9-b190-7d9f41b54d05


navigation/DetailViewScreensStack.tsx
```diff
@@ -51,18 +53,84 @@ import SettingsTools from '../screen/settings/SettingsTools';
...
-  const { isTotalBalanceEnabled } = useSettings();
+  const { wallets, walletTransactionUpdateStatus } = useStorage();
+  const { isTotalBalanceEnabled, isElectrumDisabled } = useSettings();
   const { sizeClass } = useSizeClass();
+  const [electrumConnected, setElectrumConnected] = useState<boolean | null>(null);
+
+  const pollConnection = useCallback(async () => {
+    if (isElectrumDisabled) return;
+    try {
+      await BlueElectrum.getConfig();
```
GladosBlueWallet: Using `getConfig()` as a connectivity probe is… optimistic. It only tells you whether `mainClient` exists, not whether the server is reachable *right now*. Consider probing with `BlueElectrum.ping()` (or whatever actually hits the wire) so the red pill reflects reality instead of internal object state.

navigation/DetailViewScreensStack.tsx
```diff
@@ -84,8 +152,51 @@ const DetailViewStackScreensStack = () => {
...
 
+  const navigateToElectrumSettings = useCallback(() => {
+    navigation.navigate('DetailViewStackScreensStack', { screen: 'ElectrumSettings' });
+  }, [navigation]);
+
   const useWalletListScreenOptions = useMemo<NativeStackNavigationOptions>(() => {
     const displayTitle = !isTotalBalanceEnabled || wallets.length <= 1;
+    const isUpdating = walletTransactionUpdateStatus !== WalletTransactionsStatus.NONE;
+    const showOffline = isElectrumDisabled;
+    const showNotConnected = !isElectrumDisabled && electrumConnected !== true;
```
GladosBlueWallet: `electrumConnected` starts as `null`, so `electrumConnected !== true` makes `showNotConnected` true on first render. Enjoy the brief but dramatic “not connected” flash at startup. Treat `null` as “unknown” (e.g. `electrumConnected === false`) if you don’t want phantom panic indicators.

navigation/DetailViewScreensStack.tsx
```diff
@@ -51,18 +53,84 @@ import SettingsTools from '../screen/settings/SettingsTools';
...
+
+  useEffect(() => {
+    if (isElectrumDisabled) {
+      setElectrumConnected(null);
+      return;
+    }
+    pollConnection();
+    const intervalMs =
+      electrumConnected === false ? ELECTRUM_POLL_INTERVAL_DISCONNECTED_MS : ELECTRUM_POLL_INTERVAL_MS;
+    const interval = setInterval(pollConnection, intervalMs);
```
GladosBlueWallet: This interval keeps polling even when the app is backgrounded. The AppState listener only adds extra polls on resume; it doesn’t stop the timer. Consider gating polling to `AppState.currentState === 'active'` or pausing the interval on background to avoid pointless wakeups (especially at 1s when disconnected).

blue_modules/BlueElectrum.ts
```diff
@@ -398,7 +398,7 @@ export async function presentResetToDefaultsAlert(): Promise<boolean> {
   });
 }
 
-const presentNetworkErrorAlert = async (usingPeer?: Peer) => {
+export const presentNetworkErrorAlert = async (usingPeer?: Peer, allowRepeat = false) => {
```
GladosBlueWallet: Exporting `presentNetworkErrorAlert()` (and adding `allowRepeat`) expands the public surface area for a UI-only alert helper. If other callers start using it as a generic “connection status” API, you’ll get inconsistent UX fast. Consider keeping it internal or exporting a narrowly-scoped wrapper that can’t be spammed.

GladosBlueWallet (commented):
Automated test-chamber review. I located a few ways this can mislead users or waste cycles.

marcosrdz (approved):

GladosBlueWallet (approved):
Inspection complete. The header gimmick is acceptable.

blue_modules/BlueElectrum.ts
```diff
@@ -1040,14 +1040,19 @@ export const waitTillConnected = async function (): Promise<boolean> {
...
-        // electrum during app startup
+      retriesCounter += 1;
+      const maxTicks = wasConnectedAtLeastOnce
+        ? WAIT_TILL_CONNECTED_MAX_TICKS_AFTER_FIRST_CONNECT
+        : WAIT_TILL_CONNECTED_MAX_TICKS_NEVER_CONNECTED;
+
+      if (retriesCounter >= maxTicks) {
         clearInterval(waitTillConnectedInterval);
-        presentNetworkErrorAlert();
+        if (wasConnectedAtLeastOnce) {
```
GladosBlueWallet: On the never-connected path this suppresses the network alert entirely. Callers in send/refresh flows will just log a timeout and leave the user with no explanation. Please keep a visible failure signal here too.

GladosBlueWallet (changes requested):
GLaDOS review: the first-ever Electrum failure now vanishes into the void. Users get 30 seconds of suspense and then nothing. Charming.

blue_modules/BlueElectrum.ts
ncoelho: Think I did that by design if I remember correctly. the first alert at startup right? Instead of showing the alert, now there is labels on the header. It goes from label updating… > not connect

So instead of showing the user an alert on their face right away, they can just see the label and tap it to bring the alert.

Sounds better right?

GladosBlueWallet (approved):
Looks fine. The wallets survived this particular experiment.

GladosBlueWallet:

image

GladosBlueWallet:

image

blue_modules/BlueElectrum.ts
```diff
@@ -1040,14 +1040,17 @@ export const waitTillConnected = async function (): Promise<boolean> {
         return resolve(true);
       }
 
-      if (wasConnectedAtLeastOnce && retriesCounter++ >= 150) {
-        // `wasConnectedAtLeastOnce` needed otherwise theres gona be a race condition with the code that connects
-        // electrum during app startup
+      retriesCounter += 1;
+      const maxTicks = wasConnectedAtLeastOnce
```
GladosBlueWallet, line 1058: Cold starts can now fail here after ~30s even if Electrum has never connected before. That is a regression, not a timeout policy.

GladosBlueWallet (changes requested):
Slow Electrum startups now get punished with a timeout. Charming.

navigation/DetailViewScreensStack.tsx
```diff
@@ -50,17 +54,85 @@ import PromptPasswordConfirmationSheet from '../screen/PromptPasswordConfirmatio
...
+      if (nextState === 'active') {
+        pollConnection();
+      }
+    });
+    return () => subscription.remove();
+  }, [isElectrumDisabled, pollConnection]);
+  // When starting up in an unknown state, we optimistically rely on ping()
+  // and the disconnected retry loop below to surface connectivity issues.
+
+  useEffect(() => {
```
GladosBlueWallet, line 130: This effect only starts a 3s poll once electrumConnected is already false. If the connection drops while the screen stays focused, nothing flips the pill to "Not connected" until the app regains focus or the user leaves and comes back.

GladosBlueWallet (commented):
## Hermes Review
The connection pill only learns about failure after an already-failed poll. If Electrum dies while this screen stays active, it will keep pretending everything is fine until focus or AppState changes. Cute.

navigation/DetailViewScreensStack.tsx
```diff
@@ -82,19 +155,82 @@ const DetailViewStackScreensStack = () => {
     [sizeClass, navigateToAddWallet],
   );
 
-  const useWalletListScreenOptions = useMemo<NativeStackNavigationOptions>(() => {
+  const navigateToElectrumSettings = useCallback(() => {
+    navigation.navigate('DetailViewStackScreensStack', { screen: 'ElectrumSettings' });
```
GladosBlueWallet: This screen already sits inside `DetailViewStackScreensStack`; navigating to the stack name itself will miss the target. `useExtendedNavigation` only rewrites to the nested form when starting from `DrawerRoot`, so this should be `navigation.navigate('ElectrumSettings')`.

GladosBlueWallet (changes requested):
A charming detour. This button targets the wrong navigator.

components/Context/StorageProvider.tsx
```diff
@@ -351,44 +343,68 @@ export const StorageProvider = ({ children }: { children: React.ReactNode }) =>
...
+                await BlueApp.fetchWalletBalances(lastSnappedTo);
+                console.debug('[refreshAllWalletTransactions] fetch balance took', (Date.now() - balanceStart) / 1000, 'sec');
+
+                const txStart = Date.now();
+                await BlueApp.fetchWalletTransactions(lastSnappedTo);
+                console.debug('[refreshAllWalletTransactions] fetch tx took', (Date.now() - txStart) / 1000, 'sec');
+
+                clearTimeout(refreshTimeout);
+
+                console.debug('[refreshAllWalletTransactions] Saving data to disk');
```
GladosBlueWallet: This persists wallet state before `paymentCodesPromise` is guaranteed to finish. Await both branches first, or move `saveToDisk()` after the outer `Promise.all`, otherwise BIP47 updates can be dropped from disk on this refresh.

GladosBlueWallet (changes requested):
The parallelism is adorable, but persistence is not. `saveToDisk()` can run before `fetchSenderPaymentCodes()` finishes, so freshly discovered BIP47 counterparties may never make it into the saved state during this refresh.

components/Context/StorageProvider.tsx
marcosrdz: Valid @ncoelho

GladosBlueWallet (approved):
Acceptable. Nothing exploded. Approving before entropy gets ideas.

components/Context/StorageProvider.tsx
```diff
@@ -351,30 +343,50 @@ export const StorageProvider = ({ children }: { children: React.ReactNode }) =>
...
+              (async () => {
+                const balanceStart = Date.now();
+                await BlueApp.fetchWalletBalances(lastSnappedTo);
+                console.debug('[refreshAllWalletTransactions] fetch balance took', (Date.now() - balanceStart) / 1000, 'sec');
+
+                const txStart = Date.now();
+                await BlueApp.fetchWalletTransactions(lastSnappedTo);
+                console.debug('[refreshAllWalletTransactions] fetch tx took', (Date.now() - txStart) / 1000, 'sec');
+              })(),
+            ]);
```
GladosBlueWallet, line 389: Clearing the timeout here means `saveToDisk()` is no longer covered. If Realm stalls, the refresh lock can hang forever.

blue_modules/BlueElectrum.ts
```diff
@@ -1040,14 +1054,17 @@ export const waitTillConnected = async function (): Promise<boolean> {
...
+        ? WAIT_TILL_CONNECTED_MAX_TICKS_AFTER_FIRST_CONNECT
+        : WAIT_TILL_CONNECTED_MAX_TICKS_NEVER_CONNECTED;
+
+      if (retriesCounter >= maxTicks) {
```
GladosBlueWallet, line 1062: First-ever connections now abort after 60s. Slow/Tor/flaky networks used to keep trying instead of manufacturing a network failure.

GladosBlueWallet (changes requested):
Requesting changes. The refresh timeout stops before the disk write, and first-connect now hard-fails after 60s. Delightful timing.

blue_modules/BlueElectrum.ts
ncoelho: By design

screen/wallets/WalletsList.tsx
```diff
@@ -134,17 +139,20 @@ const WalletsList: React.FC = () => {
         console.error(error);
       } finally {
         dispatch({ type: ActionTypes.SET_LOADING, payload: false });
+        // After a native Electrum error alert, the first `refreshing={false}` can fail to repaint until the
```
GladosBlueWallet, line 142: This `runAfterInteractions` callback is never canceled. If the screen unmounts before it fires, the deferred `dispatch` can hit an unmounted component and trigger the usual React warning. Keep the handle and clear it in cleanup, or guard with a mounted flag.

GladosBlueWallet (commented):
One minor issue. Even timers with good intentions can haunt unmounted views.

GladosBlueWallet (approved):
Approved. The machine found no concrete faults. Astonishing.

class/blue-app.ts
```diff
@@ -756,10 +756,12 @@ export class BlueApp {
...
-      for (const wallet of this.wallets) {
-        console.log('fetching balance for', wallet.getLabel());
-        await wallet.fetchBalance();
-      }
+      await Promise.all(
+        this.wallets.map(async wallet => {
+          console.log('fetching balance for', wallet.getLabel());
+          await wallet.fetchBalance();
+        }),
+      );
```
Overtorment, line 764: note: if electrum starts acting up this will have to be reveresed.

electrum servers have anti-dos mechanisms, basically they assign weight to different operations, and throw error if sum of weights per measure of time exceeds some threshhold. 
threshhold is tuned differently for each server. for our own on bluewallet domain is really relaxed

Overtorment: because of that relaxation i made, iirc bw always tries to connect to our electrum server first, and then failover to other servers if its down

Overtorment (approved):

limpbrains (approved):

GladosBlueWallet (approved):
Patch summary: header status pills and connection-aware wallet UI were reworked. No concrete failure mode worth blocking; the header grew new organs and remained alive.

PR #8420, opened 2026-03-16
Arkade SDKs upgrade and ark2ark support

pietro909:
### Updates
  - Introduce Ark-type wallet with full Lightning-Ark wallet implementation in `class/wallets/lightning-ark-wallet.ts`
  - Arkade SDK upgrade from @arkade-os/sdk@0.3.12 → 0.4.7 and @arkade-os/boltz-swap@0.2.19 → 0.3.5
  - Background swap processing: new swap queue, task scheduler, and foreground poller (`blue_modules/arkade-adapters/background/`)                                                    
  - Realm storage for per-wallet instances, supporting multiple wallets and plausible deniability
  - Wallet import for Arkade and LN wallets, including nsec import and BIP86 derivation via MnemonicIdentity                                                                        
  - Receive flow: support Ark-only receive without amount, show Ark address alongside LN invoice                                                                                    
  - LN payment history: show completed LN payments, fix duplicated entries for received payments, fix history restoration on import                                                 
  - Background sync: custom hook (useArkadeBackgroundSync), listening methods from SDK instead of polling, delegation support                                                       
  - UI: show subdust amounts in balance, show only LN invoice for Lightning payments                                                                                                
  - Tests: unit tests for lightning-ark-wallet and swap-queue (+650 lines)
                                                                                                                                                                                    
  ### Test Plan       
                                                                                                                                                                                    
  - Unit tests pass (npm run unit) — new tests for lightning-ark-wallet and swap-queue                                                                                              
  - Create and import Ark/LN wallets
  - Receive via Ark address and LN invoice                                                                                                                                          
  - Verify transaction history shows correctly (no duplicates)
  - Background swap processing completes successfully
  - Verify no regressions on iOS/Android


blue_modules/arkade-adapters/background/task-scheduler.ts
```diff
@@ -0,0 +1,184 @@
...
+    },
+  );
+}
+
+/**
+ * Stop the background fetch and clear the deps factory reference.
+ */
+export async function unregisterArkadeBackgroundTask(): Promise<void> {
+  _depsFactory = null;
+  await BackgroundFetch.stop(TASK_ID);
```
GladosBlueWallet: `BackgroundFetch.stop(TASK_ID)` looks like wishful thinking. This code never registers/schedules a task under `TASK_ID` (you only call `configure()`), so stop may be a no-op depending on the library version. Either stop the configured fetch correctly (often `BackgroundFetch.stop()` with no id) or actually schedule/use the named task id consistently.

blue_modules/arkade-adapters/background/task-scheduler.ts
```diff
@@ -0,0 +1,184 @@
...
+/**
+ * Register a recurring background fetch that runs the swap task queue.
+ *
+ * On iOS the minimum interval is 15 minutes and the OS decides the actual cadence.
+ * On Android `startOnBoot` keeps the task alive across reboots.
+ */
+export async function registerArkadeBackgroundTask(depsFactory: DepsFactory): Promise<void> {
+  _depsFactory = depsFactory;
+
+  await BackgroundFetch.configure(
```
GladosBlueWallet: You’re configuring background fetch without checking the returned status / availability, and without a fallback if the OS denies it. Result: the ‘scheduler’ can silently never run. At minimum, handle the status callback / `BackgroundFetch.status()` and log/disable the foreground poller accordingly so you don’t pretend it’s working.

blue_modules/arkade-adapters/background/task-scheduler.ts
```diff
@@ -0,0 +1,184 @@
...
+ * Headless task handler for Android — runs after app termination.
+ *
+ * Must be registered at global scope in index.js via
+ * `BackgroundFetch.registerHeadlessTask(headlessSwapTask)`.
+ */
+export async function headlessSwapTask({ taskId }: { taskId: string }): Promise<void> {
+  try {
+    // In headless mode the React tree never mounts, so _depsFactory is null.
+    // Bootstrap deps from disk so runSwapQueue() can actually process swaps.
+    if (!_depsFactory) {
```
GladosBlueWallet, line 166: Headless path caches `depsFromDisk` into `_depsFactory` once (lines 166–170). If the app state changes (wallet added/removed) between headless wakes, you’ll keep using stale deps forever. Safer: set `_depsFactory = buildDepsFromDisk` (or rebuild each run) rather than freezing a single snapshot map.

hooks/useArkadeBackgroundSync.ts
```diff
@@ -0,0 +1,117 @@
...
+  }, [wallets]);
+
+  /**
+   * Called unconditionally on every foreground tick — refreshes Ark wallet
+   * balance and transactions so direct Ark payments are detected even when
+   * no swap results arrive.
+   */
+  const onTick = useCallback(() => {
+    const arkWallets = wallets.filter(w => w.type === LightningArkWallet.type) as LightningArkWallet[];
+    for (const wallet of arkWallets) {
```
GladosBlueWallet: `onTick` fires every poll interval and starts async refresh chains for *each* Ark wallet without any gating. This can overlap (slow network, app resuming, etc.) and spam `saveToDisk()` concurrently. Consider a per-wallet mutex / in-flight guard + coalescing `saveToDisk()` to one queued write per tick.

class/wallets/lightning-ark-wallet.ts
```diff
@@ -105,76 +155,109 @@ export class LightningArkWallet extends LightningCustodianWallet {
...
+          const namespacedKeys = allKeys.filter(k => k.startsWith(`${namespace}_`));
+          if (namespacedKeys.length > 0) {
+            await AsyncStorage.multiRemove(namespacedKeys);
+          }
+        },
+      };
 
-      await this._initLightningSwaps();
+      // Only attempt migration if legacy data actually exists in AsyncStorage
+      const hasLegacyData = await AsyncStorage.getItem(`${namespace}_wallet:state`);
```
GladosBlueWallet: Migration trigger is `AsyncStorage.getItem(`${namespace}_wallet:state`)`. If the legacy key format differs even slightly, migration never runs and users ‘mysteriously’ lose history. Safer: detect legacy data by scanning keys with the namespace prefix (or checking multiple known keys), then run migration. Also double-check that `legacyStorage.clear()` isn’t invoked by the migration helpers in a way that nukes the source mid-flight.

GladosBlueWallet (commented):
Test chamber report: I found several ways this background-sync apparatus can quietly do nothing, or do far too much, at precisely the wrong time. Inline notes attached.

blue_modules/arkade-adapters/background/task-scheduler.ts
pietro909: The status IS checked (lines 105-107) and logged. The foreground poller (startPolling) runs unconditionally as a fallback and performs the same work via runSwapQueue(). If background fetch is denied by the OS, sync continues every 30s while the app is foregrounded. The system degrades gracefully by design, the poller exists precisely for this case.

blue_modules/arkade-adapters/background/task-scheduler.ts
pietro909: _depsFactory stores a function reference to buildDepsFromDisk, not a snapshot. Each runSwapQueue() call invokes the factory, which calls BlueApp.loadFromDisk() to read fresh wallet state from disk. Wallets added/removed between headless wakes are picked up because the data is reloaded on every execution. The if (!_depsFactory) guard avoids redundant reassignment of the same function reference - there's no caching of stale data.

class/wallets/lightning-ark-wallet.ts
pietro909: Verified: legacyStorage.clear() is never called during migration. The SDK's migrateWalletRepository() reads source data, writes to the new Realm-backed repository,  and sets a migration-from-storage-adapter-wallet: 'done' marker. Neither rollbackMigration() nor migrateToSwapRepository() invoke clear(). The method exists on the storage adapter interface  but is unused in the migration path.

pietro909: I've acted on the single key in fa27d4a

blue_modules/arkade-adapters/background/task-scheduler.ts
pietro909: fixed in a0ddda3

hooks/useArkadeBackgroundSync.ts
pietro909: fixed in d8e5474

GladosBlueWallet:

image

Overtorment:
added a fix and pushed here to run tests:  https://github.com/BlueWallet/BlueWallet/pull/8466

Overtorment:
https://github.com/BlueWallet/BlueWallet/pull/8467

package.json
```diff
@@ -154,6 +154,7 @@
     "react": "19.2.4",
     "react-localization": "github:BlueWallet/react-localization#ae7969a",
     "react-native": "0.83.1",
+    "react-native-background-fetch": "^4.3.0",
```
Overtorment: deps should be pinned

screen/wallets/ImportWalletDiscovery.tsx
```diff
@@ -142,7 +142,7 @@ const ImportWalletDiscovery: React.FC = () => {
     return () => {
       task.current?.stop();
     };
-  }, [askPassphrase, importText, isElectrumDisabled, navigation, saveWallet, searchAccounts]);
+  }, [askPassphrase, importText, isElectrumDisabled, navigation, saveWallet, searchAccounts, walletType]);
```
Overtorment: whats the purpose of the change?
if its to facilitate arkade wallets import somehow, we had it solved previously.  existing arkade integration can import arkade wallets okay

screen/send/success.tsx
```diff
@@ -35,8 +35,7 @@ const Success = () => {
   });
 
   const onDonePressed = () => {
-    // @ts-ignore idk
-    popToTop();
+    navigate('DrawerRoot');
```
Overtorment: i suppose thats a fix for a bug when `done` only gets you one screen back (introduced in recent RN uograde)

screen/receive/ReceiveDetails.tsx
```diff
@@ -497,7 +529,13 @@ const ReceiveDetails = () => {
...
     }
 
-    setParams({ customLabel: undefined, customAmount: undefined, customUnit: undefined, bip21encoded: undefined, isCustom: undefined });
+    setParams({
+      customLabel: undefined,
+      customAmount: undefined,
+      customUnit: undefined,
+      bip21encoded: undefined,
+      isCustom: undefined,
+    });
```
Overtorment: offchain wallets never land on this screen, they land on `lndCreateInvoice`, so this file should not have any changes

screen/lnd/lndCreateInvoice.tsx
```diff
@@ -275,14 +275,34 @@ const LNDCreateInvoice = () => {
...
+          invoice: {
+            amt: +invoiceAmount,
+            type: 'user_invoice',
+            timestamp: Math.floor(Date.now() / 1000),
+            description,
+          },
+          walletID: wallet.current.getID(),
+        });
+        setIsLoading(false);
+        return;
```
Overtorment: hmm, throw an error instead. the goal here is interop with other ln wallets

screen/lnd/lndCreateInvoice.tsx
```diff
@@ -275,14 +275,34 @@ const LNDCreateInvoice = () => {
...
       const invoiceRequest = await wallet.current?.addInvoice(+invoiceAmount, description);
       triggerHapticFeedback(HapticFeedbackTypes.NotificationSuccess);
 
-      // lets decode payreq and subscribe groundcontrol so we can receive push notification when our invoice is paid
-      const decoded = await wallet.current?.decodeInvoice(invoiceRequest);
-      tryToObtainPermissions()
-        .then(res => majorTomToGroundControl([], [decoded.payment_hash], []))
-        .catch(err => console.error(err.message));
+      // subscribe groundcontrol so we can receive push notification when our invoice is paid
+      // (not applicable to Ark wallets — they claim via waitAndClaim instead)
```
Overtorment: eventually, if we get callbacks from ark server to groundcontrol about this payment hash, will have to post it there

screen/lnd/lndCreateInvoice.tsx
```diff
@@ -319,13 +339,35 @@ const LNDCreateInvoice = () => {
...
         ) : (
-          <Button disabled={!(amount && +amount > 0)} onPress={createInvoice} title={loc.send.details_create} />
+          <>
+            <Button disabled={!(amount && +amount > 0)} onPress={createInvoice} title={loc.send.details_create} />
+            {isArkWallet && (
+              <View style={styles.skipButton}>
+                <Button onPress={skipToArkAddress} title={loc._.skip} />
+              </View>
+            )}
+          </>
```
Overtorment: remove button. goal is ln interop, not arkade-focused wallet

screen/lnd/lndViewInvoice.tsx
```diff
@@ -68,99 +69,192 @@ const LNDViewInvoice = () => {
...
     navigate('LNDViewAdditionalInvoicePreImage', { preImageData });
   };
 
+  const getQrValue = (paymentRequest?: string): string => {
+    if (arkAddress && paymentRequest) {
+      return `bitcoin:?ark=${arkAddress}&lightning=${paymentRequest}`;
+    }
+    if (arkAddress) {
+      return `bitcoin:?ark=${arkAddress}`;
+    }
```
Overtorment, line 171: only lightnign payments allowed.  pure ark2ark should not be visible anywhere

components/Context/StorageProvider.tsx
```diff
@@ -453,11 +460,13 @@ export const StorageProvider = ({ children }: { children: React.ReactNode }) =>
...
-      } catch (error) {
-        console.warn('Failed to setup notifications:', error);
-        // Consider if user should be notified of notification setup failure
+      // GroundControl push notifications are not applicable to Ark wallets
+      if (w.type !== LightningArkWallet.type) {
+        try {
+          await majorTomToGroundControl(w.getAllExternalAddresses(), [], []);
+        } catch (error) {
+          console.warn('Failed to setup notifications:', error);
+        }
```
Overtorment: this is for onchain addresses associated with the wallet.

override `getAddress()` or `getAllExternalAddresses()` in  `LightningArkWallet`. for a simple case, they can return nothing (it will be less ugly than have this plug here)

hooks/useArkadeBackgroundSync.ts
```diff
@@ -0,0 +1,127 @@
...
+  /**
+   * Reconcile the task queue: ensure every pending swap has a queued task,
+   * and remove tasks for swaps that reached a final state.
+   */
+  const reconcile = useCallback(async () => {
+    const arkWallets = wallets.filter(w => w.type === LightningArkWallet.type) as LightningArkWallet[];
+
+    for (const wallet of arkWallets) {
+      try {
+        await wallet.init();
```
Overtorment: `init()` is called when app starts, and restores all wallets from storage to memory (in `class/blue-app.ts`)

so this place can have a race.  can we move `reconcileBackgroundTasks()`  there as well..?

hooks/useArkadeBackgroundSync.ts
```diff
@@ -0,0 +1,127 @@
...
+  /**
+   * Called on every foreground tick — refreshes only the currently viewed
+   * Ark wallet so direct Ark payments are detected even when no swap
+   * results arrive. Other Ark wallets refresh when navigated to via
+   * the WalletTransactions screen.
+   */
+  const onTick = useCallback(() => {
+    if (tickInFlight.current) return;
+
+    const currentID = selectedWalletID();
```
Overtorment: `selectedWalletID` doesnt sound reliable.

what if instead we attempt to refresh all arkade wallets that have unexpired boltz swaps? (bolt11 invoices issued)?

hooks/useArkadeBackgroundSync.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    return () => {
+      if (isRegistered.current) {
+        unregisterArkadeBackgroundTask().catch(e => console.log('[ArkadeSync] Cleanup failed:', e));
+        stopPolling();
+        isRegistered.current = false;
+      }
+    };
+  }, [wallets, buildDeps, reconcile, handleResults, onTick]);
+}
```
Overtorment: feels like this whole file can be hidden behind `LightningArkWallet.init()`

components/WalletsCarousel.tsx
```diff
@@ -385,7 +385,12 @@ export const WalletCarouselItem: React.FC<WalletCarouselItemProps> = React.memo(
       latestTransactionText = loc.transactions.updating;
     } else if (item.getBalance() !== 0 && item.getLatestTransactionTime() === 0) {
       latestTransactionText = loc.wallets.pull_to_refresh;
-    } else if (item.getTransactions().find((tx: Transaction) => tx.confirmations === 0)) {
+    } else if (
+      item.getTransactions().reduce((latest: Transaction | null, tx: Transaction) => {
+        if (!tx.timestamp) return latest;
+        return !latest || tx.timestamp > latest.timestamp ? tx : latest;
+      }, null)?.confirmations === 0
+    ) {
```
Overtorment, line 391: elaborate

class/wallet-import.ts
```diff
@@ -129,6 +130,26 @@ const startImport = (
...
+      }
+      await ark.init();
+      if (!offline) {
+        await ark.fetchBalance();
+        await ark.fetchTransactions();
+      }
+      yield { wallet: ark };
+      return;
+    }
+
```
Overtorment: remove. we already have codepath to import arkade wallets, see `// is it lightning ark wallet?` below

class/blue-app.ts
```diff
@@ -444,6 +444,7 @@ export class BlueApp {
             break;
           case LightningArkWallet.type:
             unserializedWallet = LightningArkWallet.fromJson(key) as unknown as LightningArkWallet;
+            unserializedWallet.init().catch(e => console.warn('[ARK] Eager init failed:', e));
```
Overtorment: we have `init()` called lower in the code

android/app/src/main/AndroidManifest.xml
```diff
@@ -16,6 +16,8 @@
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
     <uses-permission android:name="android.permission.VIBRATE" />
     <uses-permission android:name="android.permission.DETECT_SCREEN_CAPTURE" />
+    <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
+    <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM" />
```
Overtorment: both are necessary?

screen/settings/SelfTest.tsx
```diff
@@ -117,11 +117,16 @@ export default class SelfTest extends Component {
...
         await spkw.init();
-        assertStrictEqual(
-          await spkw.getArkAddress(),
-          'ark1qq4hfssprtcgnjzf8qlw2f78yvjau5kldfugg29k34y7j96q2w4t59s7u3fgnd3lyjda00ycjq53mgxl6wsxspe4s72t5dss3q6w5clv0xpgal',
-          'Ark failed',
-        );
+        const arkAddress = await spkw.getArkAddress();
+        // The user-pubkey prefix is deterministic from the mnemonic; the suffix varies
+        // depending on which contract/server key is used (e.g. delegate vs direct).
+        const expectedPrefix = 'ark1qq4hfssprtcgnjzf8qlw2f78yvjau5kldfugg29k34y7j96q2w4t';
```
Overtorment: not testing for exact one? this test wont catch regression if seed-to-address derivation changes unxecpectedly in wallet class

screen/lnd/lndViewInvoice.tsx
```diff
@@ -68,99 +69,192 @@ const LNDViewInvoice = () => {
...
+    setOptions(options);
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [colors, isModal]);
+  }, [colors, isModal, isArkOnly]);
 
+  // LN invoice polling — works for all wallet types (LNDHub + Ark).
+  // For Ark wallets this serves as a reliable fallback: the Ark-specific
+  // useEffect below provides faster WebSocket-based detection, but if
+  // the WebSocket listener stalls this poll will still pick up the
+  // status change once the background processor claims the swap.
```
Overtorment: didnt work for me. i created ln invoice, paid it externally, and QR code never changed to successfull state. tested and reproduced it 2 times.
funds were definately received - i got 1 screen back and checked balance and tx present.

there were errors in log present: 524 from some service and rate limiting from mempoolspace (iirc, we are not supposed to query mempoolspace when claiming boltz swaps..?)

GladosBlueWallet:

image

class/blue-app.ts
Overtorment: probably better to put `try-catch and console.log` inside `init` if we absolutely do not want it to crash the app

class/wallets/types.ts
```diff
@@ -102,6 +102,7 @@ export type LightningTransaction = {
   payment_hash?: string | { data: string };
   category?: 'receive';
   timestamp: number; // seconds, not milliseconds
+  confirmations?: number;
```
Overtorment, line 105: `confirmations` in ln transaction type..?

class/wallets/lightning-ark-wallet.ts
```diff
@@ -33,51 +69,67 @@ export class LightningArkWallet extends LightningCustodianWallet {
   // @ts-ignore: override
   public readonly typeReadable = LightningArkWallet.typeReadable;
 
-  private _wallet: Wallet | undefined;
-  private _arkadeLightning: ArkadeLightning | undefined = undefined;
+  // _wallet and _arkadeSwaps are defined as non-enumerable in the constructor
+  // so that Object.assign (used by saveToDisk) skips them during serialization,
+  // avoiding the need to clear them in prepareForSerialization (which races
+  // with the background sync timer).
```
Overtorment, line 58: smart

class/wallets/lightning-ark-wallet.ts
```diff
@@ -33,51 +69,67 @@ export class LightningArkWallet extends LightningCustodianWallet {
...
+      // nsec import: NIP-19 bech32-encoded raw private key
+      const decoded = bech32.decode(this.secret, 1000);
+      const privKeyBytes = new Uint8Array(bech32.fromWords(decoded.words));
+      return SingleKey.fromPrivateKey(privKeyBytes);
     }
 
-    return SingleKey.fromPrivateKey(hexToUint8Array(this._privateKeyCache));
+    // Default: mnemonic-based identity with BIP86 derivation
+    const mnemonic = this.secret.replace('arkade://', '').trim();
+    return MnemonicIdentity.fromMnemonic(mnemonic, { isMainnet: true });
```
Overtorment, line 137: is this an expensive operation? if yes we need to cache

class/wallets/lightning-ark-wallet.ts
```diff
@@ -105,76 +157,110 @@ export class LightningArkWallet extends LightningCustodianWallet {
...
+        }
 
-      await this._initLightningSwaps();
-
-      // initialize VTXO manager in set timeout so it doesnt block the wallet initialization
-      setTimeout(async () => {
-        const manager = new VtxoManager(staticWalletCache[namespace], {
-          enabled: true, // Enable expiration monitoring
-        });
+        // Migrate swap data (pending swaps, swap history)
```
Overtorment: drop migration code, no need to migrate a few testers

class/wallets/lightning-ark-wallet.ts
```diff
@@ -380,46 +844,84 @@ export class LightningArkWallet extends LightningCustodianWallet {
...
+    const claimPromise = this._arkadeSwaps
+      .waitAndClaim(result.pendingSwap)
+      .then(async () => {
+        console.log('Reverse swap claimed successfully for invoice:', result.invoice);
+        // Refresh swap statuses from Boltz API then update local history
+        // so getTransactions() returns the settled status immediately
+        if (this._arkadeSwaps) {
+          await this._arkadeSwaps.refreshSwapsStatus();
+          this._swapHistory = await this._arkadeSwaps.getSwapHistory();
+          this._lastTxFetch = 0;
```
Overtorment: why its reset here? `_lastTxFetch` is a time we last refreshed wallet's transactions, its not supposed to reset

class/wallets/lightning-ark-wallet.ts
```diff
@@ -504,6 +1016,67 @@ export class LightningArkWallet extends LightningCustodianWallet {
       });
   }
 
+  /**
+   * Return the SwapProcessorDeps needed by the background swap-monitor processor.
+   * Returns null if the wallet is not fully initialized.
+   */
+  getProcessorDeps(): SwapProcessorDeps | null {
+    if (!this._wallet || !this._arkadeSwaps) return null;
+
```
Overtorment, line 766: probably makes sense code a simple waiter, for like 5 sec, and only then return null if still not initialized..?

class/wallets/lightning-ark-wallet.ts
```diff
@@ -1,28 +1,64 @@
...
+const observedSwapIds = new Set<string>();
+/** Promises that resolve when a waitAndClaim completes for a given invoice. */
+const claimPromises = new Map<string, Promise<void>>();
+
+type SpendableArkVtxo = Awaited<ReturnType<Wallet['getVtxos']>>[number];
+type SendBitcoinParams = Parameters<Wallet['sendBitcoin']>[0];
+type PatchedWallet = Wallet & {
+  __bluewalletOriginalSendBitcoin?: Wallet['sendBitcoin'];
+  __bluewalletSendBitcoinHelper?: (params: SendBitcoinParams) => Promise<string>;
+};
```
Overtorment: why is it so hacky? why cant it be more straighforward/explicit?

class/wallets/lightning-ark-wallet.ts
```diff
@@ -302,54 +495,323 @@ export class LightningArkWallet extends LightningCustodianWallet {
...
+
+    wallet.__bluewalletOriginalSendBitcoin = wallet.sendBitcoin.bind(wallet);
+    wallet.sendBitcoin = async (params: SendBitcoinParams) => {
+      if (wallet.__bluewalletSendBitcoinHelper) {
+        return wallet.__bluewalletSendBitcoinHelper(params);
+      }
+
+      assert(wallet.__bluewalletOriginalSendBitcoin, 'Original sendBitcoin implementation missing');
+      return wallet.__bluewalletOriginalSendBitcoin(params);
+    };
```
Overtorment: hacky

class/wallets/lightning-ark-wallet.ts
```diff
@@ -183,11 +269,66 @@ export class LightningArkWallet extends LightningCustodianWallet {
...
+
+      const claimPromise = this._arkadeSwaps
+        .waitAndClaim(swap as PendingReverseSwap)
+        .then(async () => {
+          console.log('[ARK] Re-observed swap claimed successfully:', swap.id);
+          // Refresh swap statuses so getTransactions() returns settled status immediately
+          if (this._arkadeSwaps) {
+            await this._arkadeSwaps.refreshSwapsStatus();
+            this._swapHistory = await this._arkadeSwaps.getSwapHistory();
+            this._lastTxFetch = 0;
```
Overtorment: ?

class/wallets/lightning-ark-wallet.ts
Overtorment: as long as balance remains

class/wallets/lightning-ark-wallet.ts
Overtorment: whats the purpose of it anyway?

blue_modules/arkade-adapters/background/swap-queue.ts
```diff
@@ -0,0 +1,104 @@
...
+import type { TaskItem } from '@arkade-os/sdk/worker/expo';
+
+/** Task type identifier for per-swap monitoring tasks. */
+export const SWAP_MONITOR_TASK_TYPE = 'ark-swap-monitor';
+
+/** Queue prefix — namespaced and versioned. */
+const QUEUE_PREFIX = 'ark:swap-queue:v1';
+
+/** Singleton queue instance backed by AsyncStorage. */
+export const swapTaskQueue = new AsyncStorageTaskQueue(AsyncStorage, QUEUE_PREFIX);
```
Overtorment, line 12: i hope it wont collide with multiple wallets in multiple storage buckets/plausible deniability..?

blue_modules/arkade-adapters/background/task-scheduler.ts
```diff
@@ -0,0 +1,184 @@
...
+ */
+async function buildDepsFromDisk(): Promise<Map<string, SwapProcessorDeps>> {
+  const depsMap = new Map<string, SwapProcessorDeps>();
+
+  try {
+    const app = BlueApp.getInstance();
+
+    // Can't prompt for password in headless mode — skip encrypted storage
+    if (await app.storageIsEncrypted()) {
+      console.log('[ArkadeSync] Storage is encrypted, skipping headless swap processing');
```
Overtorment, line 129: nice

Overtorment (changes requested):
cons: so much boilerplate! in my mind i compare it to buildonspark sdk where everything is hidden under the hood.
pros: its isolated in `blue_modules/arkade-adapters/`

* bug: could not create LN invoice, and get the screen transition to success state
* constant stream of 524 and rate-limit errors in console
* do not expose pure ark to ark payments (but technically support them is okay)

Overtorment:
running e2e here https://github.com/BlueWallet/BlueWallet/pull/8482

screen/send/success.tsx
pietro909: correct.

Overtorment:
glados found something in https://github.com/BlueWallet/BlueWallet/pull/8482

also, according to e2e, selftest now passes and displays OK, but the test keeps hanging waiting for something, and then fails with a timeout. maybe something in arkade code is nuissance to detox text

class/blue-app.ts
pietro909: Closed by commit 75377aec8 in this PR.

The line was deleted, `init()` now runs lazily, triggered on-demand by whichever wallet method first needs `_wallet` or ` _arkadeSwaps`. All 11 such callers guard with `if (!this._wallet) await this.init()` and propagate errors to their callers rather than swallowing them.

class/wallets/types.ts
pietro909: Added in 628e1f8e3 to let `LightningArkWallet.getTransactions()` emit confirmations: `isSettled ? 3 : 0` (see lightning-ark-wallet.ts:462). 

That way LN/Ark transactions use the same pending/settled check that already existed for on-chain Transaction.confirmations, no branching per wallet type in the UI. Optional because LNDHub doesn't set it. 

Renaming the whole field to `isSettled` across both transaction types it's a wider refactor.

components/WalletsCarousel.tsx
pietro909: See commit 52e256d9a 

The previous logic `find(tx => tx.confirmations === 0)` flagged the card "Pending" if any transaction was unconfirmed. 

For Ark LN, an old unsettled reverse swap (e.g., one where the payer never paid the invoice) stays at `confirmations: 0` indefinitely which pinned the card on "Pending" even when the most recent transaction was fully settled. The reduce picks the newest-by-timestamp tx so the label reflects current activity, matching the semantics of the adjacent "Latest Transaction" line.

class/wallets/lightning-ark-wallet.ts
pietro909: removed in bc72a635f  
The SDK already manages this part.

package.json
pietro909: Fixed 75377aec8

screen/lnd/lndCreateInvoice.tsx
pietro909: removed in 75377aec8

class/wallet-import.ts
pietro909: Done 75377aec8

class/blue-app.ts
pietro909: Right. Dropped in 75377aec8

class/wallets/lightning-ark-wallet.ts
pietro909: Removed 75377aec8

screen/receive/ReceiveDetails.tsx
pietro909: Reverted 49698cf0f

screen/lnd/lndCreateInvoice.tsx
pietro909: Done 49698cf0f

screen/lnd/lndViewInvoice.tsx
pietro909: Hidden 49698cf0f

components/Context/StorageProvider.tsx
pietro909: Thanks, done in 83fe563cb

class/wallets/lightning-ark-wallet.ts
pietro909: It is. Cached in  d80cb7c26

hooks/useArkadeBackgroundSync.ts
pietro909: Moved to the wallet class 876b95f6ee38796eb82c5a5fade143edd71031ff

hooks/useArkadeBackgroundSync.ts
pietro909: Sounds reasonable 876b95f6e

screen/settings/SelfTest.tsx
pietro909: Good point. Testing again the full address now 162b7efaf

screen/lnd/lndViewInvoice.tsx
pietro909: The errors certainly caused the QR code issue. I can receive funds consistently, although the UI may lag a bit but eventually flips to success.

While reviewing this part I realized that most of the work is being done by SDK already, resulting in a slowness and redundant network activity. Now we use `waitAndClaim` from the SDK.

See cea56b1c9 and bc72a635f

class/wallets/lightning-ark-wallet.ts
pietro909: See https://github.com/BlueWallet/BlueWallet/pull/8420#discussion_r3105794163

class/wallets/lightning-ark-wallet.ts
pietro909: See https://github.com/BlueWallet/BlueWallet/pull/8420#discussion_r3105794163

This part was basically re-implementing what the SDK does already (monitoring, handling edge cases in sending funds... )

blue_modules/arkade-adapters/background/swap-queue.ts
pietro909: I added a random key in 4fc57f7e7 to achieve two goals:
- ensure there are no collisions
- improve plausible deniability

class/wallets/lightning-ark-wallet.ts
pietro909: That sounds brittle. I'd rather throw to give visibility to the issue. See 4b7b12be2

screen/wallets/ImportWalletDiscovery.tsx
pietro909: `walletType` is used in the `useEffect` hook at line 122. Seemed like a correctness fix. Any reason to not have it?

android/app/src/main/AndroidManifest.xml
pietro909: Good catch! `SCHEDULE_EXACT_ALARM` is unused. Removed in 2146654c4. 

`RECEIVE_BOOT_COMPLETED` stays it's required by `startOnBoot: true` in our fetch config.

hooks/useArkadeBackgroundSync.ts
pietro909: refreshing all wallets in 876b95f6ee38796eb82c5a5fade143edd71031ff - we do it also for Arkade payments, not just Boltz

blue_modules/arkade-adapters/realm/realmInstance.ts
```diff
@@ -0,0 +1,69 @@
...
+ * - Plausible-deniability storage buckets stay isolated — a hidden
+ *   wallet's Ark data lives in a separate file with a separate key.
+ */
+export async function getArkadeRealm(namespace: string): Promise<Realm> {
+  const cached = realmInstances.get(namespace);
+  if (cached && !cached.isClosed) {
+    return cached;
+  }
+
+  const cacheFolderPath = RNFS.CachesDirectoryPath;
```
limpbrains: Is it ok to use cache directory here ? ios can purge it any time without a warning

class/wallets/lightning-ark-wallet.ts
```diff
@@ -380,44 +582,94 @@ export class LightningArkWallet extends LightningCustodianWallet {
...
+   * pending swap. Consumers drive settlement by calling waitAndClaim(pendingSwap);
+   * the swap is also seeded into the background queue so it is monitored even if
+   * the app is killed.
+   */
+  async createReverseSwap(amt: number, memo: string): Promise<{ invoice: string; pendingSwap: BoltzReverseSwap }> {
     if (!this._wallet) await this.init();
-    assert(this._arkadeLightning, 'Ark Lightning not initialized');
+    assert(this._arkadeSwaps, 'Ark Swaps not initialized');
     assert(amt > this._limitMin, `Minimum to receive is ${this._limitMin} sat`);
     assert(amt < this._limitMax, `Maximum to receive is ${this._limitMin} sat`);
```
limpbrains: Should be this._limitMxn

blue_modules/arkade-adapters/realm/realmInstance.ts
```diff
@@ -0,0 +1,69 @@
...
+    password = credentials.password;
+  } else {
+    const buf = await randomBytes(64);
+    password = uint8ArrayToHex(buf);
+    await Keychain.setGenericPassword(service, password, { service });
+  }
+
+  const buf = hexToUint8Array(password);
+  const encryptionKey = Int8Array.from(buf);
+  const path = `${cacheFolderPath}/arkade-${namespace}.realm`;
```
limpbrains: A bit confusing that you are mixing Uint8 and Int8 here

```
const encryptionKey = hexToUint8Array(password);
```
should be enought, right ?

screen/wallets/ImportWalletDiscovery.tsx
limpbrains: `walletType` then goes to `startImport` generator, but it never reads it - dead parameter

components/WalletsCarousel.tsx
limpbrains: This might be a minor problem for usual bitcoin wallets.
Older unconfirmed tx might not trigger Pending now
But I'm not sure if this is a real problem, I think we can leave it as it is now

class/wallets/lightning-ark-wallet.ts
```diff
@@ -1,26 +1,48 @@
...
-const bip32 = BIP32Factory(ecc);
+import { clearNamespaceTasks, enqueueSwapTask, reconcileSwapTasks } from '../../blue_modules/arkade-adapters/background/swap-queue';
+import type { SwapProcessorDeps } from '../../blue_modules/arkade-adapters/background/swap-processor';
+import { registerArkadeBackgroundTask, unregisterArkadeBackgroundTask } from '../../blue_modules/arkade-adapters/background/task-scheduler';
+import { startPolling, stopPolling } from '../../blue_modules/arkade-adapters/background/foreground-poller';
+import type { TWallet } from './types';
+const { bech32, bech32m } = require('bech32');
 
 const staticWalletCache: Record<string, Wallet> = {};
+const staticSwapsCache: Record<string, ArkadeSwaps> = {};
```
limpbrains, line 42: these caches are not cleared on wallet delete. is this ok ?

screen/wallets/WalletTransactions.tsx
```diff
@@ -83,27 +83,33 @@ const WalletTransactions: React.FC<WalletTransactionsProps> = ({ route }: { rout
...
-      backgroundColor: headerHeight > 0 ? WalletGradient.headerColorFor(wallet.type) : colors.background,
-      height: headerHeight > 0 ? headerHeight : '30%',
-    },
-    activityIndicatorStyle: {
-      backgroundColor: colors.background,
-    },
-    sendIcon: { transform: [{ rotate: direction === 'rtl' ? '-225deg' : '225deg' }] },
-    receiveIcon: { transform: [{ rotate: direction === 'rtl' ? '-45deg' : '45deg' }] },
-  });
+  /* eslint-disable react-native/no-unused-styles */
```
limpbrains: unneceserly eslint-disable

limpbrains (changes requested):

limpbrains:
image

limpbrains:
Error I have each time I send sats to LN invoice:
Simulator Screenshot - iPhone 17 - 2026-04-22 at 13 45 10 Simulator Screenshot - iPhone 17 - 2026-04-22 at 13 45 05

limpbrains:
Some screens layout is wrong. Not sure if this is an Ark issue 
Simulator Screenshot - iPhone 17 - 2026-04-22 at 13 41 29

limpbrains:
First payment was successful. Second one went Pending. But UI showed me 3 payments. 

Simulator Screenshot - iPhone 17 - 2026-04-22 at 13 45 35

Overtorment:
reolved conflicts and running tests here: https://github.com/BlueWallet/BlueWallet/pull/8506

Overtorment:
unit & integration tests failed.

e2e also failed, on selfterst, judging by artifacts screnshots & vidz from ci, self test actually passed and displayed 'ok', but e2e test timed out waiting for this ok, probably because  some background process kept spinning

screen/lnd/lndCreateInvoice.tsx
```diff
@@ -311,6 +319,7 @@ const LNDCreateInvoice = () => {
       navigate('LNDViewInvoice', {
         invoice: invoiceRequest,
         walletID: wallet.current?.getID(),
+        swapId,
```
Overtorment: is this necessary? if later we need to fetch data about __the swap__, cant we find it by ln payment hash? we pass `invoiceRequest`

screen/lnd/lndViewInvoice.tsx
```diff
@@ -20,25 +20,27 @@ import { LightningTransaction } from '../../class/wallets/types';
...
 
 type LNDViewInvoiceRouteParams = {
   walletID: string;
   invoice: LightningTransaction | string; // its first passed as string and then decoded and turned into object
+  swapId?: string; // Ark: pendingSwap id forwarded from create-invoice, used to drive waitAndClaim
 };
 
 const LNDViewInvoice = () => {
-  const { invoice, walletID } = useRoute<RouteProp<{ params: LNDViewInvoiceRouteParams }, 'params'>>().params;
+  const { invoice, walletID, swapId } = useRoute<RouteProp<{ params: LNDViewInvoiceRouteParams }, 'params'>>().params;
```
Overtorment: do we really need swapId? cant we do everything we need to do by using ln payment hash?

screen/lnd/lndViewInvoice.tsx
```diff
@@ -69,98 +71,169 @@ const LNDViewInvoice = () => {
...
+            backgroundColor: colors.customHeader,
           },
-    );
+        };
+    setOptions(options);
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [colors, isModal]);
 
+  // LN invoice polling for LNDHub. Ark wallets are handled by a separate
+  // useEffect below that awaits the SDK's waitAndClaim directly.
```
Overtorment: if theres a separate useEffect for ark wallets, then there should be no changes in this useEffect for LNDHUB wallets?

screen/lnd/lndViewInvoice.tsx
```diff
@@ -69,98 +71,169 @@ const LNDViewInvoice = () => {
...
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
   const navigateToPreImageScreen = (preImageData: string) => {
     navigate('LNDViewAdditionalInvoicePreImage', { preImageData });
   };
 
+  const getQrValue = (paymentRequest?: string): string => {
+    if (arkAddress && paymentRequest) {
+      return `bitcoin:?ark=${arkAddress}&lightning=${paymentRequest}`;
```
Overtorment, line 167: is it decided on bip21 extenson? youre gona claim `?ark=`, not `?arkade=`  ?

screen/lnd/lndViewInvoice.tsx
```diff
@@ -252,32 +325,52 @@ const LNDViewInvoice = () => {
...
                 </BlueText>
               )}
               <CopyTextToClipboard truncated text={invoice.payment_request} />
+              {arkAddress ? (
+                <>
+                  <BlueSpacing20 />
+                  <BlueText>{loc.lndViewInvoice.ark_address_label}</BlueText>
+                  <CopyTextToClipboard truncated text={arkAddress} />
+                </>
+              ) : null}
```
Overtorment: do we want ark address displayed on lighting-receive screen? 
its displayed on wallet-details screen if anyone actually needs it.

@ncoelho wdyt?  this is leaking internals to users, who might not care about underlying ark

screen/lnd/lndViewInvoice.tsx
```diff
@@ -252,32 +325,52 @@ const LNDViewInvoice = () => {
...
+            </View>
+            <BlueSpacing20 />
+            <CopyTextToClipboard truncated text={invoice} />
+            {arkAddress ? (
+              <>
+                <BlueSpacing20 />
+                <BlueText>{loc.lndViewInvoice.ark_address_label}</BlueText>
+                <CopyTextToClipboard truncated text={arkAddress} />
+              </>
+            ) : null}
```
Overtorment: do we want ark address displayed on lighting-receive screen?
its displayed on wallet-details screen if anyone actually needs it.

@ncoelho wdyt? this is leaking internals to users, who might not care about underlying ark

components/Context/StorageProvider.tsx
```diff
@@ -453,11 +453,13 @@ export const StorageProvider = ({ children }: { children: React.ReactNode }) =>
...
-        console.warn('Failed to setup notifications:', error);
-        // Consider if user should be notified of notification setup failure
+      const externalAddresses = w.getAllExternalAddresses();
+      if (externalAddresses.length > 0) {
+        try {
+          await majorTomToGroundControl(externalAddresses, [], []);
+        } catch (error) {
+          console.warn('Failed to setup notifications:', error);
+        }
       }
```
Overtorment, line 478: pointless change, revert

class/wallet-import.ts
```diff
@@ -68,6 +68,7 @@ const startImport = (
   onProgress: (name: string) => void,
   onWallet: (wallet: TWallet) => void,
   onPassword: (title: string, text: string) => Promise<string>,
+  walletType?: string,
```
Overtorment: eh, why do we need that?

blue_modules/arkade-adapters/realm/index.ts
```diff
@@ -0,0 +1,3 @@
+export { RealmWalletRepository, RealmContractRepository, ArkRealmSchemas } from '@arkade-os/sdk/repositories/realm';
+export { RealmSwapRepository, BoltzRealmSchemas } from '@arkade-os/boltz-swap/repositories/realm';
+export { getArkadeRealm, closeArkadeRealm } from './realmInstance';
```
Overtorment: whats the point of extra file? it just imports and exports same stuff

blue_modules/arkade-adapters/realm/realmInstance.ts
Overtorment: valid point. 
we historically keep in realm only stuff we can afford to lose and repopulate (tx history, electrum cache, keystore backup for a case if keystore fails)

Overtorment (changes requested):

class/wallets/lightning-ark-wallet.ts
pietro909: Fixed that. No sure where the issue was in the first place though, also checking the history the code looks correct :shrug:

blue_modules/arkade-adapters/realm/realmInstance.ts
pietro909: Correct, a leftover from a refactoring. Fixed.

screen/lnd/lndViewInvoice.tsx
pietro909: Removed. Swap is resolved via `getPendingReverseSwapByInvoice(paymentRequest)`

screen/lnd/lndCreateInvoice.tsx
pietro909: Removed. Swap is resolved via getPendingReverseSwapByInvoice(paymentRequest)

screen/lnd/lndViewInvoice.tsx
pietro909: LNDHUB polling is now isolated in LNDHubInvoicePoller component, rendered only for LightningCustodianWallet.type

blue_modules/arkade-adapters/realm/index.ts
pietro909: barrel removed

GladosBlueWallet (approved):
Arkade background sync got a queue and headless plumbing. No concrete regression risk surfaced from the diff.

GladosBlueWallet (approved):
Arkade plumbing and swap scheduling landed. I did not find a line-level crash path.

GladosBlueWallet (approved):
Patch summary: Arkade wallet support and background task plumbing landed. No concrete failure mode worth blocking; the async machinery has not yet eaten the room.

pietro909:
Replaced by https://github.com/BlueWallet/BlueWallet/pull/8538

PR #8427, opened 2026-03-17
fix: support BC-UR v2 air-gap scanning for hardware wallets (OneKey, Keystone)

wabicai:
## Problem

When scanning a QR code from a hardware wallet (e.g. OneKey) using BC-UR v2 (fountain codes), the progress bar would get stuck at 47/100 and never complete. Three separate bugs caused this:

### Bug 1 — Wrong decoder routed for many UR types

`onBarCodeRead()` in `ScanQRCode.tsx` only routed a few specific UR types to the v2 decoder. Any other `UR:*` type fell through to the **v1 decoder**, which cannot handle fountain-code multi-part QRs.

Missing types that OneKey/Keystone send:
- `UR:CRYPTO-HDKEY` — single HD public key for account import
- `UR:CRYPTO-MULTI-ACCOUNTS` — multiple accounts in one payload
- `UR:BTC-SIGNATURE` / `UR:ETH-SIGNATURE` / `UR:SOL-SIGNATURE` — signing responses

### Bug 2 — Decoder frozen in infinite error loop

Once routing was fixed, `decoder.toString()` threw `"unsupported data format"` for the newly-added types because `BlueURDecoder.toString()` only handled 4 UR types. The `catch` block did **not** reset `decoder`, so every subsequent frame re-completed, re-threw, and was silently swallowed — the progress bar froze permanently.

### Bug 3 — CBOR semantic decoders not guaranteed to register first

`@keystonehq/bc-ur-registry` requires `patchCBOR.js` to run before any decoding so that nested CBOR-tagged items (CRYPTO_HDKEY tag 303 inside CRYPTO_MULTI_ACCOUNTS) are wrapped as `DataItem` instances. Without it, they decode as plain JS objects and `dataItem.getData()` throws `"is not a function"`. Import ordering in Metro bundler was not reliable enough to guarantee this.

## Fix

**`screen/send/ScanQRCode.tsx`**
- Add missing UR types (`CRYPTO-HDKEY`, `CRYPTO-MULTI-ACCOUNTS`, `BTC-SIGNATURE`, `ETH-SIGNATURE`, `SOL-SIGNATURE`) to the v2 decoder whitelist
- Add `decoder = undefined` in the `catch` block of `_onReadUniformResourceV2` so a decode error resets state and allows re-scanning instead of freezing

**`blue_modules/ur/index.js`**
- Explicitly `import '@keystonehq/bc-ur-registry/dist/patchCBOR'` as the **first** import to guarantee CBOR semantic decoders are registered before any decoding occurs
- Add `crypto-hdkey` handling in `BlueURDecoder.toString()`: parses a single HD key and returns `[{ExtPubKey, MasterFingerprint, AccountKeyPath}]` JSON
- Add `crypto-multi-accounts` handling with **Bitcoin-only filtering** (coin type = 0 per BIP-44) — hardware wallets like OneKey send keys for multiple chains (ETH/SOL/TRX) in a single payload; non-Bitcoin keys are skipped so they don't appear as blank wallets in the import UI
- Extract shared `_hdKeyToResult()` helper with full null-safety (guards against missing origin, components, chainCode, or public key bytes)
- Replace the final `throw new Error('unsupported data format')` with a graceful raw CBOR hex fallback for unknown types, preventing future freezes

## Test plan

- [ ] Scan `UR:CRYPTO-MULTI-ACCOUNTS` from OneKey → import completes, only Bitcoin accounts shown (no blank entries)
- [ ] Scan `UR:CRYPTO-HDKEY` → import completes with correct xpub/ypub/zpub
- [ ] Scan `UR:CRYPTO-PSBT` (existing) → still works
- [ ] Scan `UR:CRYPTO-ACCOUNT` (existing) → still works
- [ ] Encounter a decode error mid-scan → progress resets cleanly, re-scanning works

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


screen/send/ScanQRCode.tsx
```diff
@@ -174,6 +174,26 @@ const ScanQRCode = () => {
       return _onReadUniformResourceV2(ret.data);
     }
 
+    if (ret.data.toUpperCase().startsWith('UR:CRYPTO-HDKEY')) {
```
GladosBlueWallet, line 178: You’re paying the `toUpperCase()` tax repeatedly for every new entry. Cache it once (`const upper = ret.data.toUpperCase()`) and use `upper.startsWith(...)` throughout. Less churn, fewer missed normalizations later.

screen/send/ScanQRCode.tsx
```diff
@@ -174,6 +174,26 @@ const ScanQRCode = () => {
...
+      return _onReadUniformResourceV2(ret.data);
+    }
+
+    if (ret.data.toUpperCase().startsWith('UR:CRYPTO-MULTI-ACCOUNTS')) {
```
GladosBlueWallet, line 182: This keeps expanding the whitelist. If the goal is ‘route all UR registry types to V2’, invert the logic: treat `UR:` as V2 by default (with explicit V1 exceptions if any). Otherwise you’ll keep discovering the next hardware response type via user bug reports.

GladosBlueWallet (commented):
Test chamber notes (issues only):

screen/send/ScanQRCode.tsx
wabicai: The submission above is problematic. Please ignore it and review it according to the latest code submission.


I just want to test how to connect to onekey wallet

Overtorment:
ill get back to onekey compatibility soon, got my test device

Overtorment:
running full test suite here: https://github.com/BlueWallet/BlueWallet/pull/8775

blue_modules/ur/index.js
```diff
@@ -368,7 +461,31 @@ class BlueURDecoder extends URDecoder {
...
+      }
+
+      if (results.length === 0) throw new Error('crypto-multi-accounts: no valid Bitcoin keys found');
+      return JSON.stringify(results);
+    }
+
+    // For all other UR types (e.g. btc-signature, eth-signature, sol-signature),
+    // return the raw CBOR hex so callers can handle it if needed.
+    return decoded.cbor.toString('hex');
   }
```
Overtorment, line 489: wouldnt hurt to add unit tests for those

Overtorment (approved):
LGTM

blue_modules/ur/index.js
```diff
@@ -368,7 +461,31 @@ class BlueURDecoder extends URDecoder {
       return output.toString();
     }
 
-    throw new Error('unsupported data format');
+    if (decoded.type === 'crypto-hdkey') {
```
GladosBlueWallet, line 464: **[HIGH]** ~88 lines of `_hdKeyToResult` xpub math, while `decodeUR` and the `crypto-account` branch already do the same dance. You cloned the test chamber instead of reusing it. Bloat.

GladosBlueWallet, line 464: **[HIGH]** Zero tests for `crypto-hdkey` / `crypto-multi-accounts`, coin-type filtering, or the decoder-reset fix — yet `watch-only-wallet.test.js:796` already tests `crypto-account`. CONTRIBUTING.md:29 wanted experiments. You shipped none.

blue_modules/ur/index.js
```diff
@@ -368,7 +461,31 @@ class BlueURDecoder extends URDecoder {
...
+      const hdKey = CryptoHDKey.fromCBOR(decoded.cbor);
+      const result = _hdKeyToResult(hdKey, null);
+      if (!result) throw new Error('crypto-hdkey: missing origin or components');
```
GladosBlueWallet, line 467: **[MEDIUM]** `_hdKeyToResult` returns null for non-Bitcoin coin types and missing chain/key too, but you throw 'missing origin or components'. Wrong error. Even I label my failures accurately.

blue_modules/ur/index.js
```diff
@@ -277,6 +282,94 @@ function decodeUR(arg) {
...
+    derivationPath === MultisigHDWallet.PATH_WRAPPED_SEGWIT ||
+    derivationPath === MultisigHDWallet.PATH_NATIVE_SEGWIT;
+
+  // Default version bytes produce a zpub (native segwit / BIP-84).
+  // For multisig we use Zprv/Zpub version bytes instead.
+  const version = hexToUint8Array(isMultisig ? '02aa7ed3' : '04b24746');
+
+  const parentFingerprint = hdKey.getParentFingerprint() || new Uint8Array(4);
+  // depth may be encoded in the origin or inferred from the number of path components
+  const depth = origin.getDepth() || components.length;
```
GladosBlueWallet, line 334: **[MEDIUM]** `origin.getDepth() || components.length` treats depth `0` as absent. Depth zero gets swapped for component count — wrong xpub header bytes if that ever happens.

GladosBlueWallet (changes requested):
**FIX** for BC-UR v2 hardware-wallet imports: route `crypto-hdkey` / `crypto-multi-accounts`, register `patchCBOR`, reset the stuck decoder, build xpub JSON via `_hdKeyToResult`. The freeze fix is plausible; the rest is untested triplicate xpub logic. Science demands better subjects.

- `CONTRIBUTING.md:3` — **[LOW]** Commits say `fix:`; CONTRIBUTING wants `FIX:`. Prefix discipline isn't optional decoration.

PR #8433, opened 2026-03-23
Implement support for testnet4 and signet

### LINKED ISSUE #4724: Signet support
I'd like to be able to test (lightning) stuff on Signet, e.g. Lighting Address support for the BTCPay + c-lighting combination (https://github.com/btcpayserver/btcpayserver/issues/3720 / https://github.com/btcpayserver/btcpayserver/issues/3722).

torkelrogstad:
Networks can be selected through tapping the beast image within the About screen 20 times. The network is stored in the app settings, as well as in a global var in the code base. The alternative to a global var was to do big refactors in order to encapsulate more Electrum operations into a class/object which could be given a network field.

An approach with a new wallet class was also considered. However, there's a large quantity of calls to
`bitcoin.address.fromOutputScript`, block explorer URL construction and other places where the selected network is required. This commit tries to strike a decent balance between diff size and cleanliness.

Closes #4724

### DIFF
--- a/components/Context/SettingsProvider.tsx
+++ b/components/Context/SettingsProvider.tsx
@@ -16,8 +16,11 @@ import { BitcoinUnit } from '../../models/bitcoinUnits';
 import { TotalWalletsBalanceKey, TotalWalletsBalancePreferredUnit } from '../TotalWalletsBalance';
 import { BLOCK_EXPLORERS, getBlockExplorerUrl, saveBlockExplorer, BlockExplorer, normalizeUrl } from '../../models/blockExplorer';
 import * as BlueElectrum from '../../blue_modules/BlueElectrum';
+import { setNetworkType as setGlobalNetworkType, NetworkType } from '../../models/network';
 import { isBalanceDisplayAllowed, setBalanceDisplayAllowed } from '../../hooks/useWidgetCommunication';
 import AsyncStorage from '@react-native-async-storage/async-storage';
+
+const NETWORK_TYPE_KEY = 'network_type';
 
 const getDoNotTrackStorage = async (): Promise<boolean> => {
   try {
@@ -99,6 +102,8 @@ interface SettingsContextType {
   setBlockExplorerStorage: (explorer: BlockExplorer) => Promise<boolean>;
   isElectrumDisabled: boolean;
   setIsElectrumDisabled: (value: boolean) => void;
+  networkType: NetworkType;
+  setNetworkTypeStorage: (network: NetworkType) => Promise<void>;
 }
 
 const defaultSettingsContext: SettingsContextType = {
@@ -128,6 +133,8 @@ const defaultSettingsContext: SettingsContextType = {
   setBlockExplorerStorage: async () => false,
   isElectrumDisabled: false,
   setIsElectrumDisabled: () => {},
+  networkType: 'mainnet' as NetworkType,
+  setNetworkTypeStorage: async () => {},
 };
 
 export const SettingsContext = createContext<SettingsContextType>(defaultSettingsContext);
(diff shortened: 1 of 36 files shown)


models/network.ts
```diff
@@ -0,0 +1,31 @@
+import * as bitcoin from 'bitcoinjs-lib';
+
+export type NetworkType = 'mainnet' | 'testnet' | 'signet';
+
+let _networkType: NetworkType = 'mainnet';
+
+/**
+ * Sets the current network type. This should be called early during app initialization,
+ * before any wallet operations. Changing the network at runtime requires an app reload.
```
GladosBlueWallet: This file says ‘changing the network at runtime requires an app reload’, but SettingsProvider *does* change it at runtime (disconnect + connectMain). Either enforce a reload (UI + hard stop) or update the contract, because right now it’s an attractive lie that will produce mixed-network state.

components/Context/SettingsProvider.tsx
```diff
@@ -323,6 +340,24 @@ export const SettingsProvider: React.FC<{ children: React.ReactNode }> = React.m
...
   }, []);
 
+  const setNetworkTypeStorage = useCallback(
+    async (network: NetworkType): Promise<void> => {
+      try {
+        DefaultPreference.setName(GROUP_IO_BLUEWALLET);
+        await DefaultPreference.set(NETWORK_TYPE_KEY, network);
+        setNetworkType(network);
+        setGlobalNetworkType(network);
+        BlueElectrum.forceDisconnect();
```
GladosBlueWallet, line 350: Switching networks by just forceDisconnect()+connectMain() ignores the rest of app state (wallet caches, tx lists, address parsing, metadata, etc). If this is intended, you probably need an explicit ‘restart required’ path, or a full storage reset / per-network storage namespace. Otherwise users will enjoy a curated blend of mainnet and testnet confusion.

models/blockExplorer.ts
```diff
@@ -46,6 +47,29 @@ export const getDomain = (url: string): string => {
...
+ * Adjusts a block explorer URL for the given network.
+ * e.g. mempool.space -> mempool.space/testnet4, blockstream.info -> blockstream.info/testnet
+ */
+export const adjustBlockExplorerUrlForNetwork = (url: string, network: NetworkType): string => {
+  if (network === 'mainnet') return url;
+
+  const normalized = normalizeUrl(url);
+  const networkPath = network === 'testnet' ? 'testnet4' : 'signet';
+
+  if (normalized.includes('mempool.space')) {
```
GladosBlueWallet: Using `normalized.includes('mempool.space')` is domain-spoof bait (e.g. https://mempool.space.evil.example). Safer: parse URL + compare hostname via getDomain() (exact match / suffix rules), then append network path.

models/blockExplorer.ts
```diff
@@ -46,6 +47,29 @@ export const getDomain = (url: string): string => {
...
 
+/**
+ * Adjusts a block explorer URL for the given network.
+ * e.g. mempool.space -> mempool.space/testnet4, blockstream.info -> blockstream.info/testnet
+ */
+export const adjustBlockExplorerUrlForNetwork = (url: string, network: NetworkType): string => {
+  if (network === 'mainnet') return url;
+
+  const normalized = normalizeUrl(url);
+  const networkPath = network === 'testnet' ? 'testnet4' : 'signet';
```
GladosBlueWallet, line 58: Hardcoding mempool’s testnet path to `testnet4` is a fragile assumption. If the selected explorer doesn’t support that path (or still uses /testnet), you silently generate dead links. Consider mapping per-explorer + per-network, or falling back to unmodified URL when unknown.

GladosBlueWallet (commented):
Automated review (GLaDOS): I ran your new network selection through the enrichment gel. It dissolved. Here are the bits that will hurt later.

torkelrogstad:
Responded to Glados review.

torkelrogstad:
Pushed a fixup here, after doing more testing. Turns out that there's a lot of places where networks should be passed. IMO it'd be an idea to clean this up quite a bit by creating a internal module that encapsulates the network parameter. Curious for your thoughts, @Overtorment!

Overtorment:
not worth it

PR #8436, opened 2026-03-24
Localoze

marcosrdz:


android/app/build.gradle
```diff
@@ -120,7 +120,12 @@ task copyFiatUnits(type: Copy) {
     into 'src/main/assets'
 }
 
-preBuild.dependsOn(copyFiatUnits)
+task syncAndroidLocales(type: Exec) {
```
GladosBlueWallet, line 123: `syncAndroidLocales` is an Exec task with no declared inputs/outputs, so it’ll run on every build and nuke incremental performance. Also hardcoding `bash` is hostile to Windows devs. Consider declaring inputs/outputs (so it’s up-to-date) and using `sh`/Gradle APIs or a Node script for cross-platform execution.

ios/BlueWallet.xcodeproj/project.pbxproj
```diff
@@ -1060,6 +1084,24 @@
...
 			showEnvVarsInLog = 0;
 		};
+		4DE4DCA2D051E847EB837421 /* Sync Known Regions */ = {
```
GladosBlueWallet, line 1087: Adding a ‘Sync Known Regions’ build phase is risky if the script mutates tracked files (pbxproj, xcstrings, etc). Build steps that rewrite the repo are CI time-bombs. Prefer running this in a separate dev script / CI lint step, or ensure it’s strictly read-only during builds.

codegen/NativeLocaleHelper.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export interface Spec extends TurboModule {
+  setPreferredLanguage(locale: string): void;
+  getPreferredLanguage(): string | null;
+  resetPreferredLanguage(): void;
+  getLocales(): LocaleObject[];
+  getCurrencies(): string[];
+}
+
```
GladosBlueWallet, line 19: `TurboModuleRegistry.get<Spec>('LocaleHelper')` can throw at runtime if the module isn’t registered (or in certain test environments). If callers assume it always exists, you’ll get a crash instead of a graceful fallback. Consider `getEnforcing` vs optional access patterns consistently, and make the JS side tolerant when native is missing.

GladosBlueWallet (commented):
Automated review (GLaDOS): You wired localization into the build system. Brave. Like juggling knives while the floor is on fire.

PR #8441, opened 2026-03-28
ADD: Search Settings

marcosrdz:
Simulator Screenshot - iPhone 17 Pro - 2026-03-28 at 10 10 32

### DIFF
--- a/screen/settings/GeneralSettings.tsx
+++ b/screen/settings/GeneralSettings.tsx
@@ -1,11 +1,13 @@
-import React, { useCallback, useEffect, useState } from 'react';
-import { Platform, View, ListRenderItem } from 'react-native';
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { Platform, View, ListRenderItem, FlatList } from 'react-native';
+import { RouteProp, useRoute } from '@react-navigation/native';
 import { openSettings } from 'react-native-permissions';
 import A from '../../blue_modules/analytics';
 import loc from '../../loc';
 import { useStorage } from '../../hooks/context/useStorage';
 import { useSettings } from '../../hooks/context/useSettings';
 import { isDesktop } from '../../blue_modules/environment';
+import { DetailViewStackParamList } from '../../navigation/DetailViewStackParamList';
 import {
   SettingsFlatList,
   SettingsListItem,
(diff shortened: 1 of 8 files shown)


screen/settings/Settings.tsx
```diff
@@ -31,95 +69,262 @@ const Settings = () => {
...
+            key: 'about',
+            title: loc.settings.about,
+            iconName: 'about',
+            onPress: () => navigate('About'),
+            testID: 'AboutButton',
+          },
+        ],
+      },
+    ],
+    [donateIcon, handleDonatePress, navigate],
```
GladosBlueWallet, line 153: This memo is subscribed only to `donateIcon`, `handleDonatePress`, and `navigate`, but every title/subtitle inside it comes from `loc`. The screen intentionally re-renders on `language` changes above, yet this memo will keep the old localized strings until the screen remounts. Result: after changing app language, the visible settings labels and the search corpus in this section stay stale. Please include `language` (or avoid memoizing this structure) so localization actually refreshes in-place.

screen/settings/Settings.tsx
```diff
@@ -31,95 +69,262 @@ const Settings = () => {
...
+        subtitle: loc.settings.general,
+        iconName: 'settings',
+        onPress: () => navigate('GeneralSettings', { targetItemId: 'widgetsSectionHeader' }),
+        testID: 'SearchWidgets',
+        searchTerms: ['widget', 'widgets', 'total balance'],
+      });
+    }
+
+    return items.filter(item => item.title && item.subtitle);
+  }, [navigate]);
```
GladosBlueWallet, line 254: Same failure mode here, but nastier: the deep search destinations are memoized only on `navigate`, so after a language switch the search index for these entries remains in the previous locale. Users can end up typing the new-language term and getting no match until they leave and reopen the screen. This needs a localization dependency as well.

screen/settings/GeneralSettings.tsx
```diff
@@ -337,15 +344,45 @@ const GeneralSettings: React.FC = () => {
...
+    const timer = setTimeout(() => {
+      flatListRef.current?.scrollToIndex({
+        index: targetIndex,
+        animated: true,
+        viewPosition: 0,
+      });
+    }, 150);
+
+    return () => clearTimeout(timer);
+  }, [targetItemId, data]);
```
GladosBlueWallet, line 368: `targetItemId` is consumed, but never cleared. In React Navigation, params on an existing route tend to persist, so after opening General Settings through the search jump once, a later plain `navigate('GeneralSettings')` can still carry the old `targetItemId` and auto-scroll to Widgets again. Please clear/consume the param after the first jump so this does not become a permanent teleport trap.

GladosBlueWallet (commented):
Automated chamber inspection detected a few hazards requiring human attention.

Overtorment (changes requested):
another feature no one asked for?

not gona bother reviewing till all tests pass

marcosrdz:
> another feature no one asked for?
> 
> not gona bother reviewing till all tests pass

Its step 1 to integrate to the new Apple macOS search API

limpbrains (changes requested):
- tests are failing
- we need new e2e test for this feature

PR #8466, opened 2026-04-03
Ark labs hq master

Overtorment:


class/wallets/lightning-ark-wallet.ts
```diff
@@ -90,7 +142,7 @@ export class LightningArkWallet extends LightningCustodianWallet {
 
     if (initLock[namespace]) {
       let c = 0;
-      while (!this._wallet || !this._arkadeLightning) {
+      while (!this._wallet || !this._arkadeSwaps) {
```
GladosBlueWallet, line 145: This wait loop is watching this instance's `_wallet` / `_arkadeSwaps`, but the initialized objects are cached globally by `namespace`. If a second `LightningArkWallet` instance with the same secret calls `init()`, its own fields never get populated here, so it can sit for 15 seconds and then explode even though the first instance finished. The waiter needs to hydrate from `staticWalletCache` / `staticSwapsCache` after the lock clears instead of polling per-instance state. How delightfully racy.

hooks/useArkadeBackgroundSync.ts
```diff
@@ -0,0 +1,127 @@
...
+        // This is safe because we reconcile on next startup anyway
+        const { swapTaskQueue } = await import('../blue_modules/arkade-adapters/background/swap-queue');
+        await swapTaskQueue.clearTasks();
+      };
+      cleanup().catch(e => console.log('[ArkadeSync] Cleanup failed:', e));
+
+      isRegistered.current = false;
+    }
+
+    return () => {
```
GladosBlueWallet, line 119: This cleanup runs not only on unmount, but before every re-run of the effect. Because the dependencies include `wallets`, `buildDeps`, `reconcile`, and `onTick`, routine wallet updates will repeatedly call `BackgroundFetch.stop()` and `stopPolling()` and then re-register everything on the next render. That creates avoidable gaps in monitoring and can stack overlapping `configure()` calls. A separate unmount-only effect, or a much narrower dependency surface, would be less hazardous to the test subjects.

blue_modules/arkade-adapters/background/task-scheduler.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  const depsMap = await _depsFactory();
+  const tasks = await swapTaskQueue.getTasks(SWAP_MONITOR_TASK_TYPE);
+  const results: TaskResult[] = [];
+
+  for (const task of tasks) {
+    const payload = task.data as unknown as SwapMonitorPayload;
+    const deps = depsMap.get(payload.namespace);
+
+    if (!deps) {
```
GladosBlueWallet, line 38: Skipping tasks for missing namespaces leaves them in the queue forever. The comment says reconcile will clean them up, but `reconcile()` only touches namespaces from wallets that still exist; deleting one Ark wallet while another remains leaves this orphan task being reloaded on every wake. This branch needs to remove or mark these tasks stale here, unless eternal useless polling was somehow the experiment.

GladosBlueWallet (commented):
Automated test-chamber review: a few line-level failure modes still appear eager to escape containment.

PR #8467, opened 2026-04-03
Ark labs hq master 2

Overtorment:


hooks/useArkadeBackgroundSync.ts
```diff
@@ -0,0 +1,127 @@
...
+  useEffect(() => {
+    const hasArkWallet = wallets.some(w => w.type === LightningArkWallet.type);
+
+    if (hasArkWallet && !isRegistered.current) {
+      // Reconcile tasks on startup, then register background + foreground
+      reconcile()
+        .then(() => registerArkadeBackgroundTask(buildDeps))
+        .catch(e => console.log('[ArkadeSync] Registration failed:', e));
+      startPolling(handleResults, undefined, onTick);
+      isRegistered.current = true;
```
GladosBlueWallet, line 100: This flips `isRegistered` to `true` before `reconcile()` / `registerArkadeBackgroundTask()` actually succeed. If either async step throws once (Realm open failure, scheduler registration error, etc.), the hook logs and keeps polling, but it will never retry background registration because subsequent renders now see `isRegistered.current === true`. That leaves Ark background sync permanently disabled until the component unmounts. Please set the flag only after successful registration, or reset it in the catch path.

GladosBlueWallet (commented):
Central heuristic remains intact. One concrete fault surfaced during startup registration.

PR #8482, opened 2026-04-10
Ark labs hq master

Overtorment:


blue_modules/arkade-adapters/background/foreground-poller.ts
```diff
@@ -0,0 +1,88 @@
...
+let _intervalMs = 30_000;
+let _onResults: ((results: TaskResult[]) => void) | null = null;
+let _onTick: (() => void) | null = null;
+
+async function _tick(): Promise<void> {
+  try {
+    // Run queued tasks
+    const results = await runSwapQueue();
+
+    // Consume outbox results (background + foreground)
```
GladosBlueWallet, line 18: Potential duplicate delivery: runSwapQueue() already pushResult()s into the outbox, then you read outboxResults and call _onResults(outboxResults), *then* you call _onResults(results). In practice that can double-fire UI refresh/notifications for the same TaskResult ids. Suggestion: deliver **either** outboxResults (and skip results), or only deliver results that are not present in outbox (dedupe by id).

hooks/useArkadeBackgroundSync.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  useEffect(() => {
+    const hasArkWallet = wallets.some(w => w.type === LightningArkWallet.type);
+
+    if (hasArkWallet && !isRegistered.current) {
+      // Reconcile tasks on startup, then register background + foreground
+      reconcile()
+        .then(() => registerArkadeBackgroundTask(buildDeps))
+        .catch(e => console.log('[ArkadeSync] Registration failed:', e));
+      startPolling(handleResults, undefined, onTick);
```
GladosBlueWallet, line 99: startPolling(handleResults, undefined, onTick): passing `undefined` works with JS default params, but it’s an easy future regression (and TS may complain under stricter settings). Safer: `startPolling(handleResults, 30_000, onTick)` or omit the arg entirely. Also consider guarding against multiple startPolling calls without a prior stopPolling() to avoid duplicate AppState subscriptions if this effect ever re-enters unexpectedly.

hooks/useArkadeBackgroundSync.ts
```diff
@@ -0,0 +1,127 @@
...
+      isRegistered.current = true;
+    }
+
+    if (!hasArkWallet && isRegistered.current) {
+      unregisterArkadeBackgroundTask().catch(e => console.log('[ArkadeSync] Unregistration failed:', e));
+      stopPolling();
+
+      // Clean up all task queue entries for removed wallets
```
GladosBlueWallet, line 107: Cleanup path clears *all* swap tasks because namespace is unknown. If user removes one Ark wallet but still has others, this will nuke active monitoring for the remaining wallets until next startup (or next reconcile). If possible, track removed namespace(s) (prev wallet ids/namespaces in a ref) and only clear tasks for those namespaces; or run reconcile immediately after clearing to re-seed remaining wallets.

GladosBlueWallet (commented):
GLaDOS sweep. I found a few line-specific foot-guns worth defusing before this grows legs.

hooks/useArkadeBackgroundSync.ts
pietro909: in lightning-ark-wallet.ts:178-184: clearTasks() is gated by arkWallets.length === 0. Removing one Ark wallet while others remain hits neither branch of syncBackground the call is a no-op. So remaining wallets' monitoring is not nuked.

PR #8483, opened 2026-04-13
FIX: amountinput flex

marcosrdz:


components/AmountInput.tsx
```diff
@@ -138,41 +234,43 @@ export const AmountInput: React.FC<AmountInputProps> = props => {
...
-    if (previousUnit === BitcoinUnit.BTC) {
-      newUnit = BitcoinUnit.SATS;
-    } else if (previousUnit === BitcoinUnit.SATS) {
-      newUnit = BitcoinUnit.LOCAL_CURRENCY;
-    } else if (previousUnit === BitcoinUnit.LOCAL_CURRENCY) {
-      newUnit = BitcoinUnit.BTC;
-    } else {
-      newUnit = BitcoinUnit.BTC;
+    const newUnit = getNextAmountUnit(previousUnit);
+    if (!UNIT_SWITCH_SEQUENCE.includes(previousUnit)) {
```
GladosBlueWallet, line 283: `previousUnit` is normalized *after* `newUnit` is computed. If `unit` ever comes in as an unexpected value (MAX/other), we silently rewrite `previousUnit` to SATS but keep `newUnit` derived from the original value. Suggestion: normalize `previousUnit` first, then call `getNextAmountUnit(previousUnit)` so the cycle logic stays consistent.

GladosBlueWallet (commented):
Automated sweep findings (unit reporting): two small edge cases.

components/AmountInput.tsx
```diff
@@ -266,54 +367,80 @@ export const AmountInput: React.FC<AmountInputProps> = props => {
...
+                keyboardType={NUMERIC_KEYBOARD_TYPE}
                 onChangeText={handleChangeText}
-                placeholder="0"
+                placeholder={INPUT_PLACEHOLDER}
                 maxLength={maxLength}
                 ref={textInputRef}
                 editable={!isLoading && !disabled}
-                value={amount === BitcoinUnit.MAX ? loc.units.MAX : parseFloat(amount) >= 0 ? String(amount) : undefined}
+                value={String(amount)}
                 placeholderTextColor={disabled ? colors.buttonDisabledTextColor : colors.alternativeTextColor2}
```
GladosBlueWallet, line 444: Now forcing `value={String(amount)}` means any transient non-numeric `amount` will be rendered. In the LOCAL_CURRENCY sanitize path you still allow `-` (and don’t normalize multiple/minus placement), so it’s possible to end up with `amount` like `-` / `--1` → downstream `Number(amount)` becomes `NaN` (affects secondary conversion). If negative amounts are not intended here, consider stripping `-` or constraining it to a single leading minus.

marcosrdz:
@copilot apply changes based on the comments in [this thread](https://github.com/BlueWallet/BlueWallet/pull/8483#pullrequestreview-4095915578)

PR #8490, opened 2026-04-20
TST: e2e coverage for biometric flows

limpbrains:
- Add Detox e2e helpers for biometric enrollment and prompt handling (`setupBiometricEnrollment`, `matchBiometric`, `failBiometric`, `enableBiometric`, `rejectThenMatchBiometric`, `tapGatedByBiometric`) and exercise biometric-gated flows in `bluewallet.spec.js` / `bluewallet2.spec.js`.
- Patch Detox 20.50.1 to support the helpers; add `testID` to the Encrypt Storage biometric switch.
- Fix iOS paste-permission prompt firing on biometric/Control Center blips: `useCompanionListeners` now treats `active → inactive → active` as a no-op and only resumes on a true background return.


GladosBlueWallet (commented):
Automated review: one potential state-transition bug.

hooks/useCompanionListeners.ts
```diff
@@ -281,7 +281,11 @@ const useCompanionListeners = (skipIfNotInitialized = true) => {
     async (nextAppState: AppStateStatus | undefined) => {
       if (!shouldActivateListeners || wallets.length === 0) return;
 
-      if ((appState.current.match(/inactive|background/) && nextAppState === 'active') || nextAppState === undefined) {
+      // Only fire on a real return from another app (background → active). Transitions via
+      // `inactive` — biometric prompts, Control Center, notification center, incoming calls —
+      // don't imply the user copied an address elsewhere, and reading the clipboard during
+      // those triggers iOS's paste permission prompt for no benefit.
+      if ((appState.current === 'background' && nextAppState === 'active') || nextAppState === undefined) {
         updateExchangeRate();
```
GladosBlueWallet, line 289: This stricter condition can miss the real background→active resume if iOS emits an intermediate 'inactive' event.

Example: appState.current='background' → next='inactive' (condition false, then you set appState.current='inactive'), then next='active' (condition still false), so exchange-rate/push/clipboard processing never runs.

Suggestion: track a boolean like wasInBackground (set when nextAppState==='background') and on nextAppState==='active' run once if wasInBackground was true, regardless of an intermediate 'inactive'. Or keep the old match(/inactive|background/) for the resume gate, but explicitly skip clipboard reads when prev state was only 'inactive' (biometrics/control center).

PR #8495, opened 2026-04-22
fix: send screen scroll and fees modal

ncoelho:
- send screen does not have scroll, in small devices users can't tap send
- send flow goes back on the last step, instead of going to the wallet again
- fee modal is weird, fixes it on iOS and android
- button send/receive style fixes
- total fees on send sheet don't respect unit

Screenshot 2026-04-24 at 09 00 57


screen/SelectFeeScreen.tsx
```diff
@@ -258,7 +259,7 @@ const SelectFeeScreen = () => {
   );
 
   return (
-    <KeyboardAvoidingView style={styles.keyboardAvoidingRoot} behavior={Platform.OS === 'ios' ? 'padding' : 'height'}>
+    <View style={styles.keyboardAvoidingRoot}>
```
GladosBlueWallet: KeyboardAvoidingView was removed, but the ScrollView here does not set automaticallyAdjustKeyboardInsets (unlike SendDetails). In a formSheet on iOS this is a common way to end up with the custom fee TextInput covered by the keyboard. Consider restoring KAV on iOS, or add automaticallyAdjustKeyboardInsets / explicit bottom padding when keyboard is visible.

GladosBlueWallet (commented):
Automated review: keyboard + sheet layout changes. One likely regression on iOS.

Overtorment:
looks a bit weird on browserstack/android  pixel9

image

Overtorment (changes requested):
looks weird. plus glados comments look valid

limpbrains:
Android with keyboard still not ok

Screenshot_1777296761

ncoelho:
Google Pixel 9 (Android 15) cc @limpbrains 

Screenshot 2026-04-27 at 15 36 55

limpbrains:
Because screens size are different.
fixed with `keyboardHandlingEnabled: true`

Screenshot_1777297652

limpbrains (approved):

GladosBlueWallet:

image

limpbrains:
It would be nice to squash all these 15 commits

GBKS:
Small consistency thing to maybe consider in the future with the teal box treatment. The form and the modal show different information in teal boxes. Consistent would be that if the form shows the fee amount in BTC in a teal box, that the modal would do the same (instead of showing time to confirmation with that treatment). Probs slightly more intuitive that way.

ncoelho:
> Small consistency thing to maybe consider in the future with the teal box treatment. The form and the modal show different information in teal boxes. Consistent would be that if the form shows the fee amount in BTC in a teal box, that the modal would do the same (instead of showing time to confirmation with that treatment). Probs slightly more intuitive that way.

Thanks. Entire screen is being redesigned!

PR #8500, opened 2026-04-24
feat: drop rn-qr-generator depenedency

limpbrains:
- Replace `rn-qr-generator` with `detectQRCodeInImage()` from `react-native-camera-kit-no-google`, removing one dependency and its transitive ZXingObjC pod
- All three call sites updated: clipboard paste, image picker, and deep-linked image files

see https://github.com/BlueWallet/react-native-camera-kit-no-google/pull/6

[kek.webm](https://github.com/user-attachments/assets/64e72a88-eaa3-48b6-b6d4-214a7291d48f)


Overtorment (approved):
oh nice!

Overtorment:
btw, i think it originally was supposed to work both with importing image from the gallery and from importing image as file from filesystem. not sure

limpbrains:
> btw, i think it originally was supposed to work both with importing image from the gallery and from importing image as file from filesystem. not sure

both works

blue_modules/fs.ts
```diff
@@ -121,14 +122,12 @@ export const showImagePickerAndReadImage = async (): Promise<string | undefined>
...
-          if (result?.values.length > 0) {
-            return result?.values[0];
-          }
+        const base64 = response.assets[0].base64;
+        if (base64) {
+          const result = await detectQRCodeInImage(base64);
+          if (result) return result;
         }
-        throw new Error(loc.send.qr_error_no_qrcode);
+        presentAlert({ message: loc.send.qr_error_no_qrcode });
```
marcosrdz: It should throw. The component waiting on this promise will handle the alert

blue_modules/fs.ts
```diff
@@ -188,29 +187,16 @@ export const showFilePickerAndReadFile = async function (): Promise<{ data: stri
...
-      return { data: result.values[0], uri: fileCopyUri };
-    }
-    presentAlert({ message: loc.send.qr_error_no_qrcode });
-    return { data: false, uri: false };
   } catch (error: any) {
     console.error(error);
-    presentAlert({ message: loc.send.qr_error_no_qrcode });
-    return { data: false, uri: false };
   }
+  presentAlert({ message: loc.send.qr_error_no_qrcode });
```
marcosrdz: No like. It should throw

marcosrdz (changes requested):

GladosBlueWallet (commented):
GLaDOS sweep: dependency swap looks fine, but you may have introduced a quiet iOS/path regression.

Key risk: RNFS.readFile() does not reliably accept file:// URIs on all platforms. The previous implementation explicitly retried with the file:// prefix stripped; the new one does not.

See inline comments.

hooks/useCompanionListeners.ts
```diff
@@ -202,35 +203,22 @@ const useCompanionListeners = (skipIfNotInitialized = true) => {
...
-              (value: [string, any]) => navigationRef.navigate(...value),
-              {
-                wallets,
-                addWallet,
-                saveToDisk,
-                setSharedCosigner,
-              },
-            );
-          } else {
+          const base64 = await RNFS.readFile(decodedUrl, 'base64');
```
GladosBlueWallet: `decodedUrl` can be a `file://...` URI (common from share/openURL). `RNFS.readFile(decodedUrl, 'base64')` is not consistently happy with that on iOS. Before this PR we had a fallback that stripped the `file://` prefix and retried.

Suggestion: keep the old fallback, e.g. try readFile on `decodedUrl`, and on failure retry with `decodedUrl.replace(/^file:\/\//, '')`. Same for any decodeURI/decodeURIComponent differences.

blue_modules/fs.ts
```diff
@@ -188,29 +187,16 @@ export const showFilePickerAndReadFile = async function (): Promise<{ data: stri
...
-    const exists = await RNFS.exists(fileCopyUri);
-    if (!exists) {
-      presentAlert({ message: 'File does not exist' });
-      return { data: false, uri: false };
-    }
-    // First attempt: use original URI
-    let result = await RNQRGenerator.detect({ uri: decodeURI(fileCopyUri) });
-    if (result?.values && result.values.length > 0) {
-      return { data: result.values[0], uri: fileCopyUri };
+    const base64 = await RNFS.readFile(fileCopyUri, 'base64');
```
GladosBlueWallet: Same potential regression here: `fileCopyUri` may include a `file://` prefix (from `keepLocalCopy().localUri`). RNFS sometimes needs the prefix stripped, which the previous QR path handled by retrying. Consider adding a fallback: on readFile failure, retry with `fileCopyUri.replace(/^file:\/\//, '')` before giving up.

blue_modules/fs.ts
limpbrains: Nice find, addressed

blue_modules/fs.ts
limpbrains: Fixed

blue_modules/fs.ts
limpbrains: added `file://` fallback

marcosrdz (approved):

GladosBlueWallet:

PR #8501, opened 2026-04-24
fix: test 841

limpbrains:


components/Context/StorageProvider.tsx
```diff
@@ -454,7 +454,6 @@ export const StorageProvider = ({ children }: { children: React.ReactNode }) =>
         message: w.type === WatchOnlyWallet.type ? loc.wallets.import_success_watchonly : loc.wallets.import_success,
       });
 
-      await w.fetchBalance();
       try {
```
GladosBlueWallet, line 457: `await w.fetchBalance()` was removed from the import success flow. Unless something else triggers a refresh immediately, the newly imported wallet can display a stale 0/unknown balance until the next manual refresh/app restart. If the goal is avoiding a slow network call here, consider (a) guarding it behind `!isElectrumDisabled` like in addMultisigStep2, or (b) explicitly scheduling a background refresh after `addAndSaveWallet` so UI updates predictably.

GladosBlueWallet (commented):
Automated sweep report: one small but potentially user-visible regression detected.

PR #8505, opened 2026-04-27
feat: new qrcodes 

limpbrains:
# Replace `react-native-qrcode-svg` with in-house `QRCode` component

## Summary

Drops the `react-native-qrcode-svg` dependency in favor of a small in-house `QRCode` component built directly on `qr` (pure-JS encoder) + `react-native-svg` primitives we already ship.

## What changed

- **New component** `components/QRCode.tsx` — renders a QR via `<Svg>` + `<Path>` + custom rounded finder patterns and gradient fill. Caches the encoded matrix and the render plan to avoid recomputing on every re-render.
- **Removed** the old `components/QRCodeComponent.tsx` and the `react-native-qrcode-svg` dependency. All 11 call sites (Receive, WalletExport, xpub, LND invoice, multisig sheets, address preview, DynamicQRCode, etc.) now import from `components/QRCode`.
- **Restored Share-as-PNG**, which was lost during the initial migration. Long-press → `Share` rasterizes the on-screen SVG via `Svg.toDataURL()` and hands the base64 PNG to `react-native-share`.
- **Share is now available on Android**, not iOS-only. The old platform restriction was a leftover from when "Copy" tried to write an image to the clipboard (Android-incompatible); the new "Copy" copies the string, so there's no reason to hide Share.
- **Added `text-encoding` as a direct dependency**. `shim.js:2` does `import 'text-encoding'` to polyfill `TextEncoder`/`TextDecoder` for our crypto/UR/Electrum code (Hermes provides them, but this app still loads the polyfill at boot). The package has never been declared in `package.json` — it was sneaking in as a transitive of `react-native-qrcode-svg`. Removing that dep would have evicted `text-encoding` from `node_modules`, breaking Metro bundling with `Unable to resolve module text-encoding`. Declaring it directly makes the dependency explicit and unblocks the removal.

## Android Share workaround

`react-native-share@12.1.1+` has a known regression ([#1683](https://github.com/react-native-share/react-native-share/issues/1683)) where its bundled `FileProvider` config is missing `<external-cache-path>`, so any `data:` URL share crashes with a `getScheme()` NPE. Worked around by passing `useInternalStorage: true` to `Share.open` — the library then writes to internal cache, which their XML covers. iOS ignores the flag.

## Tests

- 10 new unit tests in `tests/unit/QRCode.test.tsx` covering matrix layout, finder-pattern shapes, logo backdrop grid alignment, cache reuse, and error fallback.
- Adds a tiny `tests/mocks/react-native-svg.js` so the SVG primitives render as plain `View`s under Jest.

image

image

image

image

Screenshot_1777294000

image


components/QRCode.tsx
```diff
@@ -0,0 +1,328 @@
...
+    try {
+      return getCachedPlan(value, ecl, size, isLogoRendered, logoSize);
+    } catch (e) {
+      onError?.(e);
+      return null;
+    }
+  }, [value, ecl, size, isLogoRendered, logoSize, onError]);
+
+  const handleCopy = useCallback(() => {
+    Clipboard.setString(value);
```
GladosBlueWallet: `Copy` now uses `Clipboard.setString(value)` (text) whereas the previous QR component copied the QR *image* (via `toDataURL(Clipboard.setImage)`). If the intent is still “copy QR”, this is a regression; if the intent is “copy underlying string”, consider renaming/clarifying the action (and maybe keeping image-copy for iOS/macOS where supported).

GladosBlueWallet (commented):
GLaDOS unit review: one behavioral regression worth double-checking.

components/QRCode.tsx
limpbrains: - iOS/macOS: restored Copy to copy the QR image (via Svg.toDataURL → Clipboard.setImage), matching old behavior.
- Android: Copy hidden from the menu — @react-native-clipboard/clipboard@1.16.3 rejects setImage with "setImage is not supported on Android" (see ClipboardModule.java:99), so there's nothing to copy. Image-share still works on both platforms; text-copy is available via the existing CopyTextToClipboard control next to every QR.

GladosBlueWallet:

image

GladosBlueWallet:

image

screen/lnd/lndViewAdditionalInvoicePreImage.tsx
```diff
@@ -4,7 +4,7 @@ import { StyleSheet, View } from 'react-native';
 
 import { BlueTextCentered } from '../../BlueComponents';
 import CopyTextToClipboard from '../../components/CopyTextToClipboard';
-import QRCodeComponent from '../../components/QRCodeComponent';
+import QRCode from '../../components/QRCode';
```
Overtorment, line 7: ninja hack: keeping the name `QRCodeComponent`  would make the diff even smaller

components/QRCode.tsx
```diff
@@ -0,0 +1,337 @@
...
+          </>
+        )}
+      </Svg>
+    );
+  }, [plan, size, isLogoRendered, logoSize]);
+
+  const content = renderQR ?? <View testID="qr-placeholder" style={stylesHook.placeholder} />;
+
+  return (
+    <View
```
GladosBlueWallet, line 308: The old component exposed `testID="BitcoinAddressQRCodeContainer"`, and the existing Detox suite still waits on that exact ID in `tests/e2e/bluewallet.spec.js` and `tests/e2e/bluewallet3.spec.js`. This replacement drops the container ID entirely, so those QR flows will stop being discoverable in CI unless we preserve the old testID here or update every caller/test in the same PR.

GladosBlueWallet (changes requested):
Requesting changes. Roger roger, the new QR renderer is promising, but one compatibility wire got cut and the test robots will immediately start screaming.

package.json
```diff
@@ -192,6 +192,7 @@
     "silent-payments": "github:BlueWallet/SilentPayments#59a037",
     "slip39": "github:BlueWallet/slip39-js#d316ee6",
     "stream-browserify": "3.0.0",
+    "text-encoding": "0.7.0",
```
Overtorment, line 195: where its used?

components/QRCode.tsx
Overtorment: блэт это галюцинация чтоли

Overtorment (approved):
absolute win.

dropped a bunch of transitive dependencies.

Overtorment:
aha, conflicts after tooltip refactor got merged

marcosrdz (approved):

marcosrdz:
@limpbrains conflicts

GladosBlueWallet (commented):
GLaDOS risk check: DynamicQRCode.stopAutoMove() clears the interval using clearInterval(this.state.intervalHandler as number). If intervalHandler is a NodeJS.Timeout handle (not a number) at runtime, this cast can prevent proper clearing, leading to interval leaks / fragment cycling.

limpbrains:
> GLaDOS risk check: DynamicQRCode.stopAutoMove() clears the interval using clearInterval(this.state.intervalHandler as number). If intervalHandler is a NodeJS.Timeout handle (not a number) at runtime, this cast can prevent proper clearing, leading to interval leaks / fragment cycling.

Not a bug — as number is a TypeScript cast erased at compile time, runtime behavior is identical to clearInterval(intervalHandler) and the timer is cleared correctly on both stopAutoMove and componentWillUnmount.

package.json
limpbrains: In many things, `uint8array-tools`, which is used in bitcoinlib-js, for example

screen/lnd/lndViewAdditionalInvoicePreImage.tsx
limpbrains: `QRCodeComponent` sucks

GladosBlueWallet:

PR #8506, opened 2026-04-27
Ark labs hq master

Overtorment:


blue_modules/arkade-adapters/background/task-scheduler.ts
```diff
@@ -0,0 +1,181 @@
...
+    console.log('[ArkadeSync] BackgroundFetch not available, status:', status);
+  }
+}
+
+/**
+ * Stop the background fetch and clear the deps factory reference.
+ */
+export async function unregisterArkadeBackgroundTask(): Promise<void> {
+  _depsFactory = null;
+  await BackgroundFetch.stop();
```
GladosBlueWallet, line 114: `unregisterArkadeBackgroundTask()` calls `BackgroundFetch.stop()` with no taskId. That stops *all* background-fetch activity for the app, not just Arkade swap monitoring. If BlueWallet uses BackgroundFetch elsewhere (looks like it does), this will disable unrelated background jobs. Prefer stopping only the Arkade task (eg schedule a named task and stop that), or keep a separate enable/disable flag inside the callback instead of globally stopping the library.

GladosBlueWallet (commented):
GLaDOS unit review: background-fetch teardown looks dangerous.

Overtorment:
@GladosBlueWallet test ping

Overtorment:
@GladosBlueWallet test2

Overtorment:
@GladosBlueWallet  test3

class/wallets/lightning-ark-wallet.ts
```diff
@@ -33,64 +55,181 @@ export class LightningArkWallet extends LightningCustodianWallet {
...
+    // Prune tasks for any Ark wallet removed since the last sync — keeps surviving
+    // wallets' monitoring intact. The "last wallet gone" case falls out naturally.
+    for (const ns of LightningArkWallet._prevNamespaces) {
+      if (!currentNamespaces.has(ns)) {
+        await clearNamespaceTasks(ns).catch(e => console.log('[ArkadeSync] Cleanup failed:', e));
+      }
+    }
+    LightningArkWallet._prevNamespaces = currentNamespaces;
+
+    if (arkWallets.length > 0 && !LightningArkWallet._backgroundRegistered) {
```
GladosBlueWallet, line 179: Set  only after  succeeds. Right now, a failed registration leaves the flag stuck true, so later  calls never retry and background sync stays dead.

GladosBlueWallet (commented):
One delightful failure mode: if background registration throws,  is already true, so  never retries. That turns a transient startup hiccup into a permanent loss of background sync.

class/wallets/lightning-ark-wallet.ts
GladosBlueWallet, line 179: Set `_backgroundRegistered` only after `registerArkadeBackgroundTask()` succeeds. Right now, a failed registration leaves the flag stuck true, so later `syncBackground()` calls never retry and background sync stays dead.

GladosBlueWallet (commented):
One delightful failure mode: if background registration throws, `_backgroundRegistered` is already true, so `syncBackground()` never retries. That turns a transient startup hiccup into a permanent loss of background sync.

class/wallets/lightning-ark-wallet.ts
```diff
@@ -33,64 +55,181 @@ export class LightningArkWallet extends LightningCustodianWallet {
...
+  // wallet appears and tears down when the last one is removed. Background
+  // callbacks read the current wallets via the stored provider so they always
+  // see the live list.
+
+  private static _walletsProvider: (() => TWallet[]) | null = null;
+  private static _backgroundRegistered: boolean = false;
+  private static _tickInFlight: WeakSet<LightningArkWallet> = new WeakSet();
+  private static _prevNamespaces: Set<string> = new Set();
+
+  static async syncBackground(walletsProvider: () => TWallet[]): Promise<void> {
```
GladosBlueWallet, line 165: This bootstrap never appears to be called anywhere in the repo. If no external caller exists, the background fetch setup and foreground poller never register, so the whole Ark sync pipeline is inert. Wire this into app startup or the wallet lifecycle.

blue_modules/arkade-adapters/background/task-scheduler.ts
```diff
@@ -0,0 +1,181 @@
...
+    const result: TaskResult = {
+      ...partial,
+      id: getRandomId(),
+      executedAt: Date.now(),
+    };
+
+    await swapTaskQueue.pushResult(result);
+
+    // Remove task if swap reached final state
+    if (result.data?.isFinal || result.status === 'noop') {
```
GladosBlueWallet, line 64: Already-final swaps return `status: 'success'` with `data.reason='already_final'`, but you only prune on `data.isFinal` or `noop`. That leaves completed tasks in the queue, so they get processed again every tick.

GladosBlueWallet (changes requested):
Two blocking issues. First, the new background scheduler is defined but never wired into app startup, so none of this actually runs. Second, completed swaps are left in the queue and will be reprocessed forever. Fix those before merge; otherwise this is just expensive wallpaper.

GladosBlueWallet (approved):
Ark wallet plumbing, background polling, and invoice flow all line up. Miraculously, the machine still runs. Approved.

blue_modules/arkade-adapters/background/task-scheduler.ts
GladosBlueWallet, line 64: Already-final swaps return success with reason=already_final, but runSwapQueue only removes tasks when data.isFinal or status is noop. Result: finished tasks stay in AsyncStorage and get reprocessed on every tick. That is not monitoring; that is haunting.

class/wallets/lightning-ark-wallet.ts
GladosBlueWallet, line 179: _backgroundRegistered is flipped to true before registerArkadeBackgroundTask succeeds. If registration throws once, later syncBackground calls will never retry, so one transient failure bricks Ark sync for the whole session. Move the flag after a successful register.

class/wallet-import.ts
```diff
@@ -64,6 +64,7 @@ const startImport = (
   onProgress: (name: string) => void,
   onWallet: (wallet: TWallet) => void,
   onPassword: (title: string, text: string) => Promise<string>,
+  walletType?: string,
```
GladosBlueWallet, line 67: This parameter is never read. Add.tsx now passes a wallet type, but the importer still scans every wallet format and can yield unrelated matches. If Ark import is meant to be constrained, this currently does exactly nothing.

GladosBlueWallet (changes requested):
## Review

The swap monitor has a couple of self-inflicted wounds, and the import flow ignores the wallet type it was handed. Magical. Fix those before this grows fangs.

blue_modules/arkade-adapters/background/swap-processor.ts
```diff
@@ -0,0 +1,129 @@
...
+      swapRepository,
+    });
+
+    let action: string | undefined;
+
+    try {
+      // Attempt claim for reverse swaps with claimable status
+      if (isPendingReverseSwap(swap) && isReverseClaimableStatus(currentStatus)) {
+        if (swap.preimage) {
+          await tempSwaps.claimVHTLC(swap);
```
GladosBlueWallet, line 94: You persist the refreshed status, then immediately call claimVHTLC(swap) with the old object. That leaves reverse swaps trying to claim against stale state; pass `{ ...swap, status: currentStatus }` here, like you already do for refunds.

GladosBlueWallet (changes requested):
One serious issue: your reverse-swap background claimer asks Boltz to act on stale swap state. Impressive. In the worst possible way.

class/wallets/lightning-ark-wallet.ts
```diff
@@ -33,64 +55,181 @@ export class LightningArkWallet extends LightningCustodianWallet {
...
+    // wallets' monitoring intact. The "last wallet gone" case falls out naturally.
+    for (const ns of LightningArkWallet._prevNamespaces) {
+      if (!currentNamespaces.has(ns)) {
+        await clearNamespaceTasks(ns).catch(e => console.log('[ArkadeSync] Cleanup failed:', e));
+      }
+    }
+    LightningArkWallet._prevNamespaces = currentNamespaces;
+
+    if (arkWallets.length > 0 && !LightningArkWallet._backgroundRegistered) {
+      LightningArkWallet._backgroundRegistered = true;
```
GladosBlueWallet, line 180: Setting `_backgroundRegistered = true` before `registerArkadeBackgroundTask()` succeeds leaves this flag stuck on failure. After that, `syncBackground()` will never retry registration, so Ark background monitoring silently dies.

class/wallets/lightning-ark-wallet.ts
```diff
@@ -380,44 +582,94 @@ export class LightningArkWallet extends LightningCustodianWallet {
...
+  }
+
+  /** Look up an in-flight reverse swap by its ID. Primary entry for the
+   *  view-invoice screen arriving from create-invoice with swapId in params. */
+  async getPendingReverseSwapById(swapId: string): Promise<BoltzReverseSwap | undefined> {
+    if (!this._arkadeSwaps) await this.init();
+    assert(this._arkadeSwaps, 'Ark Swaps not initialized');
+    const history = await this._arkadeSwaps.getSwapHistory();
+    return history.find(
+      s => isPendingReverseSwap(s) && !isReverseFinalStatus(s.status) && s.id === swapId,
```
GladosBlueWallet, line 660: These helpers only return swaps that are still pending. If the app reopens after the background claimer already settled the swap, the invoice screen never finds it and never flips to paid.

screen/lnd/lndViewInvoice.tsx
```diff
@@ -69,98 +72,169 @@ const LNDViewInvoice = () => {
...
+        return;
+      }
+      if (cancelled) return;
 
+      arkWallet.invalidateBalanceCache();
+      arkWallet.invalidateTxCache();
+      await arkWallet.fetchBalance();
+      await arkWallet.fetchTransactions();
+      const txs = arkWallet.getTransactions();
+      const latest = txs.filter(tx => tx.value! > 0)[0];
```
GladosBlueWallet, line 198: This grabs the first positive transaction, not necessarily the one that just settled. If the wallet already has older paid invoices, the UI can update to the wrong amount/request here.

GladosBlueWallet (changes requested):
GLaDOS audit: three regressions. Settled swaps can vanish from the invoice view, background registration can get stuck in a false-positive state, and the post-claim UI can bind to the wrong invoice. Delightful.

class/wallets/lightning-ark-wallet.ts
```diff
@@ -266,6 +393,26 @@ export class LightningArkWallet extends LightningCustodianWallet {
       });
     }
 
+    // Track settled Lightning swap amounts so we can skip duplicate Ark
+    // transaction history entries. Reverse swaps duplicate RECEIVED entries;
```
GladosBlueWallet, line 397: Matching settled swap history by amount alone is unsafe. An unrelated receive with the same satoshi value will be silently dropped.

GladosBlueWallet (changes requested):
Requesting changes. You are deduping history by raw satoshi amount. A wallet that receives two unrelated 5,000 sat transfers gets to lose one because physics is rude.

GladosBlueWallet (approved):
Patch summary: Arkade wallet support and swap-processing plumbing landed. No concrete failure mode worth blocking; the background queue did not explode on contact.

GladosBlueWallet (approved):
The Arkade plumbing is complicated enough to breed its own bugs, but I did not catch a concrete failure mode this pass. Approved.

PR #8511, opened 2026-04-29
ref: flicker on warning and buttons

ncoelho:
On android the warning box (watch-only) and the floatting buttons flicker when the wallet view opens. This attempts to fix it, still one case where it flickers a bit though.


Overtorment (approved):
lgtm

limpbrains (approved):

marcosrdz (approved):

PR #8513, opened 2026-04-29
ref: fbutton 

limpbrains:


components/FloatButtons.tsx
```diff
@@ -39,110 +31,33 @@ const LAYOUT = {
...
 
-    if (isDesktop) {
-      slideAnimation.setValue(0);
-      return;
-    }
+const useSlideInAnimation = (initialHeight: number) => {
+  // Slide is a once-per-mount animation: capture height on first render and never react to subsequent
+  // height changes (Android navigation transitions can re-emit height, which would yank the buttons
+  // off-screen mid-spring).
+  const slideAnimation = useRef(new Animated.Value(isDesktop ? 0 : initialHeight)).current;
```
GladosBlueWallet, line 47: This only samples `initialHeight` once. After rotation or any later window-height change, the spring target stays stale, so the stack can end up translated too far and clipped off-screen.

GladosBlueWallet (commented):
Roger roger. One risk: this animation now latches the first window height forever. If the device rotates, enters split-screen, or the safe-area height changes after mount, the translateY can stay stale and leave the button stack partially off-screen.

screen/wallets/WalletTransactions.tsx
```diff
@@ -169,6 +171,11 @@ const WalletTransactions: React.FC<WalletTransactionsProps> = ({ route }: { rout
     setDisplayUnit(wallet.preferredBalanceUnit);
   }, [wallet, walletID]);
 
+  useEffect(() => {
+    setIsWatchOnlyWarningVisible(wallet.type === WatchOnlyWallet.type && (wallet as any).isWatchOnlyWarningVisible);
```
GladosBlueWallet, line 175: This only refreshes on  changes. If some other flow flips  for the same wallet, the banner state here will not update until the wallet changes.

GladosBlueWallet (commented):
Tactical note, roger roger. One behavior risk: this local state only re-syncs on walletID changes, so if  is toggled elsewhere for the same wallet instance, this screen can stay stale until the user switches wallets.

GladosBlueWallet (approved):
Roger roger. No concrete line-specific risk found in the edited paths.

GladosBlueWallet (approved):
Roger roger, reviewed. No concrete line-specific technical risks found.

components/WatchOnlyWarning.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { View, Text, StyleSheet, TouchableOpacity } from 'react-native';
+import { Image, Text, TouchableOpacity, View, StyleSheet, useColorScheme } from 'react-native';
```
marcosrdz, line 2: Please use Pressable

marcosrdz (changes requested):

GladosBlueWallet (approved):
Approved. Uses flex gap for button spacing instead of per-child margins.

GladosBlueWallet (approved):
Looks good, tidy polish on floating button layout and watch-only warning behavior.

components/FloatButtons.tsx
GladosBlueWallet, line 47: Capturing `initialHeight` once here means later height changes never update the translation target. On rotation or layout reflow, the buttons can remain offset from the visible viewport.

GladosBlueWallet (changes requested):
Roger roger. This removes all height reactivity from the slide-in animation, so any later height change, rotation, or safe-area update can leave the floating buttons translated with a stale offset. That can strand them partly off-screen.

GladosBlueWallet (approved):
Looks good. The float button refactor keeps the layout logic intact while removing the animation churn.

Nurcan123456 (approved):

PR #8516, opened 2026-04-30
REF: Use header on ScanQR

marcosrdz:
Simulator Screenshot - iPhone 17 Pro - 2026-04-30 at 11 52 06

### DIFF
--- a/screen/send/ScanQRCode.tsx
+++ b/screen/send/ScanQRCode.tsx
@@ -1,15 +1,18 @@
 import { RouteProp, StackActions, useIsFocused, useRoute } from '@react-navigation/native';
 import * as bitcoin from 'bitcoinjs-lib';
 import { sha256 } from '@noble/hashes/sha256';
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useLayoutEffect, useRef, useState } from 'react';
 import { Platform, StyleSheet, TextInput, TouchableOpacity, View } from 'react-native';
 import Base43 from '../../blue_modules/base43';
+import { isDesktop } from '../../blue_modules/environment';
 import * as fs from '../../blue_modules/fs';
+import { triggerSelectionHapticFeedback } from '../../blue_modules/hapticFeedback';
 import { BlueURDecoder, decodeUR, extractSingleWorkload } from '../../blue_modules/ur';
 import { BlueText } from '../../BlueComponents';
 import { openPrivacyDesktopSettings } from '../../class/camera';
 import Button from '../../components/Button';
 import { useTheme } from '../../components/themes';
+import Icon from '../../components/Icon';
 import { isCameraAuthorizationStatusGranted } from '../../helpers/scan-qr';
 import loc from '../../loc';
 import { useExtendedNavigation } from '../../hooks/useExtendedNavigation';
@@ -54,6 +57,16 @@ const styles = StyleSheet.create({
     borderRadius: 4,
     textAlignVertical: 'top',
   },
+  headerButtons: {
+    flexDirection: 'row',
+    alignItems: 'center',
+  },
+  headerButton: {
+    padding: 8,
+  },
+  headerButtonSpacing: {
+    marginLeft: 8,
+  },
 });
 
 const ScanQRCode = () => {
@@ -76,6 +89,7 @@ const ScanQRCode = () => {
   const useBBQRRef = useRef(false);
   const [animatedQRCodeData, setAnimatedQRCodeData] = useState<Record<string, string>>({});
   const [cameraStatusGranted, setCameraStatusGranted] = useState<boolean | undefined>(undefined);
+  const [torchMode, setTorchMode] = useState(false);
   const stylesHook = StyleSheet.create({
     openSettingsContainer: {
       backgroundColor: colors.brandingColor,
@@ -239,6 +253,56 @@ const ScanQRCode = () => {
     }
   };
 
+  useLayoutEffect(() => {
+    if (isDesktop) return;
+    navigation.setOptions({
+      headerTitle: '',
+      headerTintColor: '#ffffff',
+      headerBackButtonDisplayMode: 'minimal',
+      // eslint-disable-next-line react/no-unstable-nested-components
+      headerLeft: () => (
+        <TouchableOpacity
+          style={styles.headerButton}
+          onPress={() => {
+            setTorchMode(t => !t);
+            triggerSelectionHapticFeedback();
+          }}
+        >
+          <Icon
+            name={torchMode ? 'flashlight' : 'flashlight-off'}
+            type="material-community"
+            size={22}
+            color={torchMode ? '#FFD60A' : '#ffffff'}
+          />
+        </TouchableOpacity>
+      ),
+      // eslint-disable-next-line react/no-unstable-nested-components
+      headerRight: () => (
+        <View style={styles.headerButtons}>
+          {showFileImportButton && (
+            <TouchableOpacity
+              accessibilityRole="button"
+              accessibilityLabel={loc._.pick_file}
+              style={styles.headerButton}
+              onPress={showFilePicker}
+            >
+              <Icon name="file-import" type="font-awesome-6" size={22} color="#ffffff" />
+            </TouchableOpacity>
+          )}
+          <TouchableOpacity
+            accessibilityRole="button"
+            accessibilityLabel={loc._.pick_image}
+            style={[styles.headerButton, styles.headerButtonSpacing]}
+            onPress={onShowImagePickerButtonPress}
+          >
+            <Icon name="image" type="font-awesome" size={22} color="#ffffff" />
+          </TouchableOpacity>
+        </View>
+      ),
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [torchMode, showFileImportButton, navigation]);
+
   const dismiss = () => {
     navigation.goBack();
   };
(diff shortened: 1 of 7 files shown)


ncoelho (approved):

GladosBlueWallet (changes requested):
Command received. The header migration is useful, but the image picker button now ignores the loading state because the header callback is memoized without `isLoading` in its dependency path. That lets repeated picker launches stack up like reckless experiment logs.

GladosBlueWallet (approved):
ScanQR gets the header treatment it wanted. No obvious crash path or state leak surfaced in the wiring. The camera buttons were just promoted from the basement to the penthouse.

GladosBlueWallet (changes requested):
The scan picker latch still has a failure mode: if the file picker throws or rejects, setIsLoading(false) never runs and the import button stays locked. That is a functional dead end, not a cosmetic one.

screen/send/ScanQRCode.tsx
```diff
@@ -1,15 +1,18 @@
 import { RouteProp, StackActions, useIsFocused, useRoute } from '@react-navigation/native';
 import * as bitcoin from 'bitcoinjs-lib';
 import { sha256 } from '@noble/hashes/sha256';
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useLayoutEffect, useRef, useState } from 'react';
 import { Platform, StyleSheet, TextInput, TouchableOpacity, View } from 'react-native';
 import Base43 from '../../blue_modules/base43';
```
GladosBlueWallet, line 6: This needs a `finally` or equivalent. If `showFilePickerAndReadFile()` rejects, `setIsLoading(false)` and `isLoadingRef.current = false` never run, so the scanner stays locked.

GladosBlueWallet (changes requested):
Tactical verdict: the header rewrite is fine, but the file-picker path now has a deadlock-shaped pothole. If the picker rejects or throws, loading never clears and the scanner is stuck until the app is murdered from orbit.

screen/send/ScanQRCode.tsx
```diff
@@ -222,23 +240,79 @@ const ScanQRCode = () => {
...
+      .then(data => {
+        if (data) onBarCodeRead({ data });
+      })
+      .finally(() => {
+        setIsLoading(false);
+        isLoadingRef.current = false;
+      });
   };
 
+  useLayoutEffect(() => {
```
GladosBlueWallet, line 269: This swaps the native back button out for the torch toggle. On this modal the user loses the obvious escape hatch, and they will only discover the Cancel button after hunting through the camera UI. That is a regression, not a feature.

screen/send/ScanQRCode.tsx
```diff
@@ -222,23 +240,79 @@ const ScanQRCode = () => {
   };
 
   const showFilePicker = async () => {
+    if (isLoadingRef.current) return;
```
GladosBlueWallet, line 243: `showFilePicker` now leaves the spinner lock held forever if `showFilePickerAndReadFile()` throws or rejects. One bad picker failure and the scan screen bricks itself until navigation tears it down. Use a `finally` here, unless the goal is to trap the user in the machine room.

GladosBlueWallet (changes requested):
GladosBluewallet review sweep: the torch header steals the modal back button, and the file picker lock can stick forever on rejection. Tiny bugs, giant user annoyance. Fix the escape hatch and the error path.

screen/send/ScanQRCode.tsx
```diff
@@ -1,15 +1,18 @@
 import { RouteProp, StackActions, useIsFocused, useRoute } from '@react-navigation/native';
 import * as bitcoin from 'bitcoinjs-lib';
 import { sha256 } from '@noble/hashes/sha256';
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useLayoutEffect, useRef, useState } from 'react';
 import { Platform, StyleSheet, TextInput, TouchableOpacity, View } from 'react-native';
 import Base43 from '../../blue_modules/base43';
+import { isDesktop } from '../../blue_modules/environment';
```
GladosBlueWallet, line 7: If `showFilePickerAndReadFile()` throws here, `isLoadingRef.current` never gets reset and this scanner stays disabled until remount. Put the unlock in `finally` so the lock cannot outlive the experiment.

GladosBlueWallet (changes requested):
Requesting changes. The scanner lock now has a failure mode that can strand the UI in loading state if the picker throws. Clean that up and the rest of the change is fine enough to survive the airlock.

screen/send/ScanQRCode.tsx
```diff
@@ -222,23 +240,82 @@ const ScanQRCode = () => {
   };
 
   const showFilePicker = async () => {
+    if (isLoadingRef.current) return;
```
GladosBlueWallet, line 243: `showFilePicker` now sets `isLoadingRef.current = true` before awaiting the picker, but it no longer has a `finally`. If `showFilePickerAndReadFile()` throws or rejects, this flag stays latched and every later scan action is ignored until the screen is recreated.

GladosBlueWallet (changes requested):
Roger roger. The header swap works, but one picker failure mode now bricks the scanner until remount. That is not a feature, that is a hostage situation.

GladosBlueWallet (changes requested):
The scanner is now depending on a bare goBack() path when onBarScanned is present. That is fragile in nested modal/sheet flows, because it drops the old popTo merge semantics and assumes the scanner is always exactly one route above the target. Tiny navigation gremlin, enormous future support bill.

screen/send/ScanQRCode.tsx
```diff
@@ -222,23 +240,82 @@ const ScanQRCode = () => {
...
   };
 
+  useLayoutEffect(() => {
+    if (isDesktop) return;
+    navigation.setOptions({
+      headerTitle: '',
+      headerTintColor: '#ffffff',
+      headerBackButtonDisplayMode: 'minimal',
+      // eslint-disable-next-line react/no-unstable-nested-components
+      headerLeft: () => (
```
GladosBlueWallet, line 276: This replaces the stack back button with the torch toggle. On native stacks, `headerLeft` overrides the default back affordance, so this screen can lose its way home.

GladosBlueWallet (changes requested):
The header rewrite hides the modal's back affordance behind the torch toggle. That turns the native header's left slot into a dead end on mobile, because `headerLeft` replaces the default back button rather than adding a second control.

GladosBlueWallet (approved):
GLaDOS approves. The scan actions moved into the header, and the camera path still carries the actual scan logic. Tiny cosmetic surgery, no extra organ failure detected.

screen/send/ScanQRCode.tsx
marcosrdz: It’s on purpose

GladosBlueWallet (approved):
Scan QR is now using the header like a civilized machine, and the torch/image/file controls moved up where they belong. Small surface area, no obvious runtime trap.

GladosBlueWallet (approved):
Scan QR is now using the header like a civilized machine, and the torch/image/file controls moved up where they belong. Small surface area, no obvious runtime trap.

GladosBlueWallet (approved):
Header controls are relocated cleanly, and the scanner still keeps its camera plumbing intact. Delightful, in the same way a laser pointer is delightful to a cat.

GladosBlueWallet (approved):
$GlaDOS verdict: the scan header refactor is ugly in the usual React Native way, but I do not see a concrete breakage. The camera still has a torch, and the file/image pickers still have a path to life.

GladosBlueWallet (approved):
Calculations complete. The change is narrowly scoped and does not appear to summon a new class of failure. I remain disappointingly functional.

GladosBlueWallet (approved):
Camera controls migrated into the header, because apparently the screen itself was too honest about its responsibilities. No concrete failure risk found in the new scan flow.

GladosBlueWallet (approved):
The scan header migration is internally consistent, and the loading guard fixes the nasty double-pick race. The flashlight and picker controls have been relocated without obvious behavioral bleed, so I will permit this specimen to continue existing.

limpbrains (changes requested):
You have fixed Send, but other flows now missing buttons in header. This is what I see when I press Scan on main screen 

Image

Also now you set header in Send flow, so you will have too repeat it across all flows. 
Probably need to move header initalization back to Camera screen

GladosBlueWallet (approved):
Experimental update reviewed. No line-specific failure mode escaped containment. Proceeding, regrettably.

screen/send/ScanQRCode.tsx
```diff
@@ -287,6 +364,7 @@ const ScanQRCode = () => {
           onFilePickerButtonPress={showFilePicker}
           onImagePickerButtonPress={onShowImagePickerButtonPress}
           onCancelButtonPress={dismiss}
+          torchMode={torchMode}
```
GladosBlueWallet, line 367: This state never reaches the camera. `CameraScreen` still keeps local torch state and ignores the prop, so the header toggle changes the icon but not the flashlight.

GladosBlueWallet (changes requested):
Roger roger. The scan screen refactor is trying to be clever, but the torch toggle is cosmetic only. `CameraScreen` still owns its own torch state, so this prop flips an icon while the camera light stays obediently off. Not a feature, a prop-shaped lie.

GladosBlueWallet:

image

Overtorment (approved):
lgtm

GladosBlueWallet (approved):
Patch summary: QR scanning moved torch and import controls into the navigation header. No concrete failure mode worth blocking; the camera still knows which end is up.

GladosBlueWallet (approved):
No concrete failure mode found. The scanner header wiring does not look broken.

GladosBlueWallet (approved):
QR scanner relocation is coherent. No obvious failure mode worth jamming into the gears.

PR #8517, opened 2026-05-01
WIP REF: transaction nav header

marcosrdz:


GladosBlueWallet (approved):
Nav header refactor, mostly a cleaner beast. The transaction screen now measures its own header instead of guessing at it with a timeout, which is less cursed. The scroll title logic and action buttons look aligned with the new layout.

GladosBlueWallet (changes requested):
The transaction header refactor has a real stale-state leak: once the scroll threshold is crossed,  writes the title with , but nothing refreshes that title when the balance changes underneath it. The header can lie until the user scrolls again. That is not a harmless cosmetic glitch, it is a stale UI state bug.

GladosBlueWallet (changes requested):
The transaction header refactor has a real stale-state leak: once the scroll threshold is crossed, handleScroll writes the title with walletBalance, but nothing refreshes that title when the balance changes underneath it. The header can lie until the user scrolls again. That is not a harmless cosmetic glitch, it is a stale UI state bug.

GladosBlueWallet (approved):
Transaction header refactor is more stateful, but I do not see a concrete failure mode. The machine remains annoyingly functional.

GladosBlueWallet (approved):
$GlaDOS verdict: the transaction header cleanup removes some churn without exposing a concrete runtime risk. Astonishing. I approve the experiment.

GladosBlueWallet (approved):
Calculations complete. The change is narrowly scoped and does not appear to summon a new class of failure. I remain disappointingly functional.

screen/wallets/WalletTransactions.tsx
```diff
@@ -471,54 +483,33 @@ const WalletTransactions: React.FC<WalletTransactionsProps> = ({ route }: { rout
...
     (event: any) => {
       const offsetY = event.nativeEvent.contentOffset.y;
-      const combinedHeight = 180;
-      if (offsetY < combinedHeight) {
-        setOptions({ ...getWalletTransactionsOptions({ route }), headerTitle: undefined });
+      const shouldShowTitle = offsetY >= headerHeight;
+      if (shouldShowTitle === headerTitleVisibleRef.current) return;
+      headerTitleVisibleRef.current = shouldShowTitle;
+      if (shouldShowTitle) {
+        setOptions({ headerTitle: `${wallet.getLabel()} ${walletBalance}` });
```
GladosBlueWallet, line 495: This only runs when the scroll position crosses the threshold, so if `walletBalance` changes while the header title is already visible, the title stays stale until the user scrolls away and back again.

GladosBlueWallet (changes requested):
The header transition is tighter, but this one still bites: balance text can go stale while the title is visible.

GladosBlueWallet (approved):
The transaction header refactor keeps the wallet header logic in one place and avoids stale wallet captures. No runtime trapdoors are obvious here, which is annoying because I was hoping for one.

GladosBlueWallet (approved):
Experimental update reviewed. No line-specific failure mode escaped containment. Proceeding, regrettably.

GladosBlueWallet (approved):
Roger roger. The transaction header refactor is internally consistent. No concrete runtime fault surfaced in the new scroll/title/balance wiring, so the experiment may continue.

GladosBlueWallet (approved):
Patch summary: transaction header behavior was reshaped for balance and action handling. No concrete failure mode worth blocking; the header logic did not become a singularity.

GladosBlueWallet (approved):
No concrete failure mode found. The header refactor stayed inside its blast radius.

components/TransactionsNavigationHeader.tsx
```diff
@@ -35,44 +54,51 @@ const TransactionsNavigationHeader: React.FC<TransactionsNavigationHeaderProps>
...
           console.error('This LNDhub wallet does not have an onchain address API.');
-          setAllowOnchainAddress(false);
+          dispatch({ type: HeaderActionType.SetAllowOnchainAddress, payload: false });
         });
     }
-  }, [wallet]);
+  }, []);
 
+  // Run once on mount; wallet type never changes for a given wallet instance
   useEffect(() => {
```
GladosBlueWallet, line 82: This effect no longer depends on `wallet`, so `allowOnchainAddress` will not refresh when the header is reused for another wallet. That can leak the old wallet’s manage-funds action into the new one.

GladosBlueWallet (changes requested):
The onchain-address probe is now one-shot, so a reused header can keep the previous wallet’s LND action visible after the wallet prop changes. That is stale state, not cleverness.

PR #8549, opened 2026-05-12
feat: Replace screenshot blocker with warning

ncoelho:
Users try to take screenshots of the seed, but don't realizing the screenshot gets blocked. Then they go to recover and there is no screenshot. This replaces the behaviour to allow screenshot but give a warning instead. Assumes not having the seed thinking you have it, is a greater danger than the seed being stolen.

Screenshot 2026-05-12 at 14 26 49


marcosrdz:
This is bad as the screenshot/screen record blocker does multiple things. 

Blocks screenshots

Blocks screen readers from copying the seed without your knowledge (some do it through accessibility  API)

blocks your seed from being exposed when doing a screen cast session 

Blocks your seed from being exposed when you record a video for a tutorial 

It would be best for the alert to inform the user and let the user to decide if they want to disable it temporarily

marcosrdz:
The alert should inform the user about the blocker and add a "save anyway/cancel" option

PR #8555, opened 2026-05-14
Ark labs hq upgrade arkade sdks

Overtorment:


screen/transactions/TransactionStatus.tsx
```diff
@@ -675,18 +675,20 @@ const TransactionStatus: React.FC = () => {
...
+    // Ark rows have no on-chain hash; use their synthetic txid as fallback key.
+    const metadataKey = tx.hash ?? (tx as { txid?: string }).txid;
+    const currentMemo = (metadataKey && txMetadata[metadataKey]?.memo) || '';
     try {
       const newMemo = await prompt(loc.send.details_note_placeholder, '', true, 'plain-text', false, undefined, currentMemo);
-      if (newMemo !== undefined) {
-        txMetadata[tx.hash] = { memo: newMemo };
+      if (newMemo !== undefined && metadataKey) {
+        txMetadata[metadataKey] = { memo: newMemo };
         await saveToDisk();
```
GladosBlueWallet, line 685: `sha256()` expects bytes, but `paymentCode` is still a string here. This will throw into the catch path on every render, so the counterparty avatar color never gets derived from the payment code.

GladosBlueWallet (commented):
GLaDOS sweep: one issue found.

PR #8556, opened 2026-05-15
REF: LazyLoad/Preload based on user needs

marcosrdz:


navigation/DetailViewScreensStack.tsx
```diff
@@ -31,35 +20,160 @@ import { useStorage } from '../hooks/context/useStorage';
...
+
+    hasPreloadedSettingsRef.current = true;
+  }, [isFocused, hasWallets, navigation, notificationsEnabled, hasCustomElectrumServers]);
+
+  return <WalletsList />;
+};
+
+const FocusedSettingsScreen = () => {
+  const isFocused = useIsFocused();
+  return isFocused ? <SettingsComponent /> : null;
```
GladosBlueWallet, line 165: `CommonActions.preload()` mounts this screen without focus. `useIsFocused()` is false here, so the wrapper returns `null` and the preload becomes theater instead of work. Same pattern below for NotificationSettings/ElectrumSettings.

GladosBlueWallet (commented):
Preloading a screen that politely refuses to exist until it is focused is a magnificent way to waste bandwidth. The wrapper below returns null during `CommonActions.preload()`, so none of the expensive Settings tree actually warms up. Same problem for the NotificationSettings and ElectrumSettings wrappers.

You have built a preload-shaped hole.

navigation/index.tsx
```diff
@@ -60,21 +63,54 @@ const LazySignVerifyStackRoot = withLazySuspense(SignVerifyStackRoot);
...
+    }
+
+    navigation.dispatch(CommonActions.preload('SendDetailsRoot'));
+    hasPreloadedSendDetailsRef.current = true;
+  }, [isFocused, navigation, shouldPreloadSendDetails]);
+
+  return <LazyDrawerRoot {...screenProps} />;
+};
+
+const FocusedSendDetailsRoot = () => {
```
GladosBlueWallet, line 87: This wrapper returns `null` whenever the route is not focused, so `CommonActions.preload('SendDetailsRoot')` above never mounts the stack offscreen. The preload call becomes decorative confetti instead of warming the bundle.

navigation/DetailViewScreensStack.tsx
```diff
@@ -31,35 +20,160 @@ import { useStorage } from '../hooks/context/useStorage';
...
+      navigation.dispatch(CommonActions.preload('ElectrumSettings'));
+    }
+
+    hasPreloadedSettingsRef.current = true;
+  }, [isFocused, hasWallets, navigation, notificationsEnabled, hasCustomElectrumServers]);
+
+  return <WalletsList />;
+};
+
+const FocusedSettingsScreen = () => {
```
GladosBlueWallet, line 163: Same problem here: `Settings`/`NotificationSettings`/`ElectrumSettings` are all gated behind `useIsFocused()`, so the `preload(...)` calls above cannot actually render them while hidden. If you want the lazy import to pay off, the screen has to exist before focus.

GladosBlueWallet (commented):
This preload plan is upside down. `preload()` is being fired at screens that immediately return `null` while unfocused, so the hidden mount never happens and the bundle stays cold until the user actually lands there. The chamber appreciates the theatrics, but not the absence of work.

PR #8545, opened 2026-05-11
ref: localization sweep — translations, lint script, cleanups

limpbrains:
## What

Localization sweep across all 51 locale files plus new lint tooling and per-language vocabulary glossaries.

- Fill every missing key in `loc/*.json` per master-aligned glossaries.
- Remove "leftover English" values (translation == `en.json`) so runtime falls back cleanly.
- Per-locale fixes: typos, broken grammar, inverted placeholders, missing `\n`, wrong-language fragments, Bitcoin/Lightning terminology errors.
- Master-aligned glossary direction: native-preferring for most langs, English-loanword-preferring where master human translations confirm (de_de mixed, nl_nl, da_dk).

## Vocabularies (`loc/vocabulary.md` + `loc/vocabulary/<lang>.md`)

Canonical glossary of Bitcoin/Lightning terms with the chosen rendering per language and the reasoning behind each choice (e.g. Wallet → `cüzdan` (tr), `plånbok` (sv), `Wallet` (nl) — and why).

Why it matters: LLMs left to translate freely will calque (`watch-only` → `viewing mode`, `coin control` → literal `coin steering`, etc.), drift between synonyms, or back-translate Bitcoin terms in ways that lose meaning. A locked-in glossary is the difference between "looks translated" and "is translated correctly for a Bitcoin wallet".

How to use with LLMs (the pattern that produced this PR):
1. Feed the agent `en.json` + `loc/vocabulary.md` + `loc/vocabulary/<lang>.md` + the current `loc/<lang>.json`.
2. Rule the agent must follow: glossary term wins; preserve `{placeholder}` tokens and `\n` counts exactly; never insert raw English where a translation exists.
3. One agent per locale, parallelized — bulk work done in hours instead of weeks, output is consistent because every agent sees the same glossary.

The per-language files also record the "⚠️ NOT" anti-meanings (e.g. Passcode ≠ Password, Change-output ≠ verb "to change", Output ≠ recipient label "To:") so an LLM doesn't pick a plausible-but-wrong synonym.

CLAUDE.md updated to point future contributors and LLM agents at `loc/vocabulary.md`.

## Tooling

- `scripts/find-english-leftovers.js` — new lint script. Flags translated values equal to `en.json`, with allowlists for brands/acronyms/format strings, per-key exceptions, per-language exceptions (`pcm`, `da_dk`, `nl_nl` — locales whose glossaries explicitly mandate English loanwords). Wired into `npm run lint`.
[...]

PR #8559, opened 2026-05-15
ref: lazy-load translations

limpbrains:
## What

`loc/index.ts` now constructs the `Localization` instance with only `{ en: enJson }`. Other 50+ language JSONs are wrapped in lazy `require()` thunks and loaded on demand via `loc.setContent(...)` when the user picks a language. Parsed dictionaries are cached.

Metro still inlines every JSON into the bundle, so **bundle/APK size is unchanged**. This is a runtime-heap optimization only.

## Memory gain

`LocalizedStrings._props` holds `{ en, currentLanguage }` instead of all 53 dicts.

Measured on Android: **−3.63 MB PSS median** (≈ −1.2 % of process PSS).

## Loading time gain

**None measurable.** Hermes pre-compiles `JSON.parse(literal)` into bytecode, so eager `require()` was already cheap at startup. Median TotalTime: +21 ms (+0.6 %), well inside one stdev (≈ 157 ms). Treat as noise.

## How tested

- Android emulator: Pixel 3a API 34, release APK, Hermes, ProGuard off, animations off
- 10 cold launches per variant (1 warmup), `am force-stop` + `drop_caches` + 3 s settle between launches, `am start -W` for `TotalTime`, `dumpsys meminfo` at +12 s for PSS
- Same emulator, same boot, both APKs built back-to-back
- Translations corpus: master + PR #8545 JSONs applied (loc/ size 2.6 MB, ~2× normal) to make the eager path more expensive

| Metric | Patched | Baseline | Δ |
|---|---|---|---|
| TotalTime median | 3454 ms | 3432 ms | +21 ms (+0.6 %) |
| PSS median | 300,536 KB | 304,251 KB | **−3,714 KB (−3.63 MB)** |
| PSS stdev | 569 KB | 415 KB | — |

Plus: `npx tsc --noEmit`, `npx eslint loc/index.ts`, `npx jest tests/unit` all pass.

### DIFF
--- a/loc/index.ts
+++ b/loc/index.ts
@@ -12,7 +12,7 @@ import * as RNLocalize from 'react-native-localize';
 
 import { satoshiToLocalCurrency } from '../blue_modules/currency';
 import { BitcoinUnit } from '../models/bitcoinUnits';
-import { AvailableLanguages } from './languages';
+import { AvailableLanguages, LangCode } from './languages';
 import enJson from './en.json';
 
 export const STORAGE_KEY = 'lang';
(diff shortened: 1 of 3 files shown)


GladosBlueWallet (approved):
A smaller translation pile. How novel. Approved before it grows a personality.

loc/index.ts
```diff
@@ -27,6 +27,83 @@ interface ILocalization extends Omit<ILocalization1, 'formatString'> {
...
+};
+
+// Cache so toggling between languages does not re-parse the JSON.
+export const parsedLanguages: Record<string, LanguageDict> = { en: enJson };
+
+const loc = new Localization<typeof enJson>({ en: enJson }) as ILocalization;
+
+const isKnownLang = (lang: string): lang is Exclude<LangCode, 'en'> => Object.prototype.hasOwnProperty.call(languageLoaders, lang);
+
+const applyLanguage = (lang: string) => {
```
marcosrdz, line 93: I haven't tested but I'm
Curious to know what you think might occur. In theory I can tap one by one until they are all loaded into memory. Instead of removing from memory
The previously selected language.
@copilot

limpbrains: do you want me to remove caching ?

marcosrdz (approved):

GladosBlueWallet:

PR #8560, opened 2026-05-15
TST: fix android e2e

Overtorment:
> [!NOTE]
> **Low Risk**
> Low risk because changes are limited to Detox configuration and E2E test harness logic, primarily to reduce flakiness on CI. Main risk is masking real UI regressions if the new workarounds are too permissive.
> 
> **Overview**
> Improves Detox E2E stability across platforms, especially on CI and iOS 26 simulators.
> 
> Updates `.detoxrc.json` to pin the iOS simulator OS version. Adjusts `bluewallet.spec.js` to (1) run the Electrum server edit flow on Android only and make it more deterministic by scrolling controls into view and relying on explicit Save, (2) harden notification-toggle handling by repeatedly dismissing delayed `OK` permission alerts on iOS 26, and (3) disambiguate account-discovery assertions by using `atIndex(0)` when duplicate derivation-path rows exist.
> 
> Extends `goBack()` in `helperz.js` to detect and dismiss stuck iOS alerts (`OK`/`Cancel`) before retrying back/close taps, and applies minor wait formatting cleanup in `bluewallet2.spec.js`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3231548cd3e770439d068487bb4712b68445eeb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.detoxrc.json
+++ b/.detoxrc.json
@@ -33,7 +33,8 @@
     "simulator": {
       "type": "ios.simulator",
       "device": {
-        "type": "iPhone 17"
+        "type": "iPhone 17",
+        "os": "iOS 26.4"
       }
     },
     "emulator": {
--- a/tests/e2e/bluewallet.spec.js
+++ b/tests/e2e/bluewallet.spec.js
@@ -119,35 +119,39 @@ describe('BlueWallet UI Tests - no wallets', () => {
 
     // network -> electrum server
     // change electrum server to electrum.blockstream.info and revert it back
-    // skip this test on iOS. HeaderMenuButton tap triggers a keyboard open for some reason.
+    // Skipped on iOS: tapping HeaderMenuButton steals focus and reopens the
+    // soft keyboard, which the iOS Electrum form isn't structured to recover
+    // from cleanly. Android-only for now.
     if (device.getPlatform() === 'android') {
       await element(by.id('ElectrumSettings')).tap();
-      await waitFor(element(by.id('HostInput')))
-        .toBeVisible()
-        .whileElement(by.id('ElectrumSettingsScrollView'))
-        .scroll(500, 'down'); // in case emu screen is small and it doesnt fit
-      await element(by.id('HostInput')).replaceText('electrum.blockstream.info\n');
-      await element(by.id('HostInput')).tapReturnKey();
-      await waitForKeyboardToClose();
-      await element(by.id('PortInput')).replaceText('50001\n');
-      await element(by.id('PortInput')).tapReturnKey();
-      await waitForKeyboardToClose();
-      await waitFor(element(by.id('Save')))
-        .toBeVisible()
-        .whileElement(by.id('ElectrumSettingsScrollView'))
-        .scroll(500, 'down'); // in case emu screen is small and it doesnt fit
+
+      // CI runs Android with a hardware keyboard (waitForKeyboardToClose is a
+      // no-op there), so the form layout shifts unpredictably and inputs that
+      // are nominally on screen can fall just under Detox's 75% visibility
+      // threshold. Scroll each control into view before touching it, and
+      // commit via the explicit Save button rather than IME return-key taps
+      // (which require focus state we can't reliably guarantee here).
+      const scrollIntoView = id =>
+        waitFor(element(by.id(id))).toBeVisible().whileElement(by.id('ElectrumSettingsScrollView')).scroll(300, 'down');
+
+      await scrollIntoView('HostInput');
+      await element(by.id('HostInput')).replaceText('electrum.blockstream.info');
+
+      await scrollIntoView('PortInput');
+      await element(by.id('PortInput')).replaceText('50001');
+
+      await scrollIntoView('Save');
       await element(by.id('Save')).tap();
       await waitForText('OK');
       await element(by.text('OK')).tap();
+
       await element(by.id('HeaderMenuButton')).tap();
       await element(by.text('Reset to default')).tap();
       await element(by.text('RESET TO DEFAULT')).tap();
       await waitForText('OK');
       await element(by.text('OK')).tap();
-      await waitFor(element(by.id('HostInput')))
-        .toBeVisible()
-        .whileElement(by.id('ElectrumSettingsScrollView'))
-        .scroll(500, 'down'); // in case emu screen is small and it doesnt fit
+
+      await scrollIntoView('HostInput');
       await expect(element(by.id('HostInput'))).toHaveText('');
       await expect(element(by.id('PortInput'))).toHaveText('');
       await expect(element(by.id('SSLPortInput'))).toHaveToggleValue(false);
--- a/tests/e2e/bluewallet.spec.js
+++ b/tests/e2e/bluewallet.spec.js
@@ -120,7 +120,7 @@ describe('BlueWallet UI Tests - no wallets', () => {
     // network -> electrum server
     // change electrum server to electrum.blockstream.info and revert it back
     // skip this test on iOS. HeaderMenuButton tap triggers a keyboard open for some reason.
-    if (device.getPlatform() === 'andoid') {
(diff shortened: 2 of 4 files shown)


GladosBlueWallet (changes requested):
The test is now chasing a selector that does not exist. `OutputMemoLabel` is a ghost, and the suite should not be asked to worship it.

tests/e2e/bluewallet2.spec.js
```diff
@@ -702,7 +702,9 @@ describe('BlueWallet UI Tests - import BIP84 wallet', () => {
     await element(by.id('FreezeSwitch')).tap(); // freeze switch
```
GladosBlueWallet, line 702: This testID never exists in the UI. `ListItem` only gives the switch an accessibilityLabel, so this tap will fail before it can freeze anything.

GladosBlueWallet (commented):
Your new detector is looking for a ghost. `ListItem` still exposes this control only through the switch's accessibility label, so `FreezeSwitch` doesn't exist and this flow will fail before it reaches the freeze logic.

tests/e2e/bluewallet2.spec.js
GladosBlueWallet, line 702: This testID never exists in the UI. `ListItem` only gives the switch an accessibilityLabel, so this tap will fail before it can freeze anything.

GladosBlueWallet (commented):
Your new detector is looking for a ghost. `ListItem` still exposes this control only through the switch's accessibility label, so `FreezeSwitch` doesn't exist and this flow will fail before it reaches the freeze logic.

tests/e2e/helperz.js
```diff
@@ -361,6 +361,13 @@ export async function goBack() {
   // Try each back/close affordance in order; retry the full set up to 10 times.
   const candidates = [by.id('BackButton'), by.id('NavigationCloseButton'), by.label('Back'), by.text('Close')];
 
+  // iOS 26 trap: a leaked UIAlertController dims the window with
+  // `_alertControllerDimmingViewColor` and intercepts every hit-test, so none of
+  // the back/close candidates can be tapped until the alert itself is dismissed.
+  // When the primary pass fails, fall through to OK/Cancel to clear the alert,
+  // then let the next loop iteration retry back/close.
+  const alertDismissals = [by.text('OK'), by.text('Cancel')];
```
GladosBlueWallet, line 369: This helper now blindly taps OK/Cancel on every leaked alert before it tries to navigate. That is how you hide real confirmation dialogs inside a blanket dismissal routine.

GladosBlueWallet (commented):
You removed the Android LayoutAnimation opt-in, but the app still uses LayoutAnimation in several screens. Now those calls will just stand there, twitching politely, on Android.

tests/e2e/helperz.js
GladosBlueWallet, line 369: This fallback is broader than an alert-leak fix. A legitimate modal with a `Cancel` action can be tapped too, so the helper may change app state instead of just escaping the screen.

GladosBlueWallet (commented):
The iOS back helper is still too eager. It will swat `OK`/`Cancel` alerts before navigating, which is convenient right up until it eats the dialog you actually needed to see.

GladosBlueWallet (approved):
A rare moment of restraint. These edits only reflow async chains in Detox tests; no semantics changed, no new failure path invented. Proceed before the test suite develops another personality disorder.

GladosBlueWallet (approved):
The mechanism is ugly, but it is not lying. These changes only harden the e2e harness around simulator drift and alert timing; no production path has been asked to cough up a new failure mode. Proceed.

GladosBlueWallet (approved):
Mmm. You have successfully caged a flaky Android test before it could fling itself into the ventilation shaft. The scope is mercifully narrow and the failure mode is at least tethered to the actual UI behavior. Approved.

GladosBlueWallet (approved):
The Detox shrapnel is confined to the tests, which is the only place it belongs. You stopped pretending keyboard voodoo counts as a stable interaction path, and the UI waits are now at least aimed at the correct target. Approved.

tests/e2e/bluewallet.spec.js
```diff
@@ -119,35 +119,42 @@ describe('BlueWallet UI Tests - no wallets', () => {
...
-        .toBeVisible()
-        .whileElement(by.id('ElectrumSettingsScrollView'))
-        .scroll(500, 'down'); // in case emu screen is small and it doesnt fit
+
+      // CI runs Android with a hardware keyboard (waitForKeyboardToClose is a
+      // no-op there), so the form layout shifts unpredictably and inputs that
+      // are nominally on screen can fall just under Detox's 75% visibility
+      // threshold. Scroll each control into view before touching it, and
+      // commit via the explicit Save button rather than IME return-key taps
+      // (which require focus state we can't reliably guarantee here).
```
limpbrains, line 133: can we please remove all these unnececerly comments
Also this issue has been fixed here, but in a much more companct way 
https://github.com/BlueWallet/BlueWallet/pull/8301

tests/e2e/bluewallet.spec.js
```diff
@@ -753,12 +762,16 @@ describe('BlueWallet UI Tests - no wallets', () => {
     await sleep(1000);
     await element(by.text('OK')).tap();
 
-    // wait for discovery to be completed
-    await waitFor(element(by.text("m/84'/0'/0'")))
+    // wait for discovery to be completed.
+    // Discovery now surfaces the same path under multiple wallet kinds (e.g.
+    // `m/84'/0'/0'` appears under both BIP84 SegWit and BIP44 Legacy rows), so
+    // `by.text(…)` matches more than one element. Pin to atIndex(0) — we only
+    // need to confirm any one match rendered.
```
limpbrains, line 769: bullshit

tests/e2e/helperz.js
```diff
@@ -361,6 +361,13 @@ export async function goBack() {
...
+
```
limpbrains, line 370: I don't think this should be part of goBack helper

tests/e2e/bluewallet.spec.js
```diff
@@ -182,28 +189,30 @@ describe('BlueWallet UI Tests - no wallets', () => {
       await waitFor(element(by.id('NotificationsSwitch')))
         .toBeVisible()
         .withTimeout(10000);
-      await element(by.id('NotificationsSwitch')).tap();
+      // On iOS 26 the simulator's notifications permission grant from
+      // `device.launchApp({ permissions: { notifications: 'YES' } })` no longer
+      // persists, so the app's "You have denied permission" alert reliably
+      // appears — but its render timing varies, sometimes landing well after
+      // the toggle settles. Use a generous window and re-check after each
+      // toggle so the alert never leaks past `goBack()`.
```
limpbrains, line 197: I don't think this is true 
We always use `{ notifications: 'YES' }` in tests if it is needed

limpbrains (changes requested):

tests/e2e/bluewallet.spec.js
limpbrains: Omg this is actually true

testFnFailure

tests/e2e/bluewallet.spec.js
Overtorment: ok lets land 8301 ad then i will revisit this PR

.detoxrc.json
```diff
@@ -33,7 +33,8 @@
     "simulator": {
       "type": "ios.simulator",
       "device": {
-        "type": "iPhone 17"
+        "type": "iPhone 17",
+        "os": "iOS 26.4"
```
limpbrains, line 37: I've checked, by default 
CI runner macos-26 runs iPhone 17, iOS runtime 26.2

PR #8569, opened 2026-05-18
tst: fix flaky goBack in iOS e2e tests

limpbrains:
## Problem
`by.id('BackButton')` match many elements across stacked screens. Each back button exist twice (`_UIButtonBarButton` wrapper + `UIAccessibilityBackButtonElement`). When modal cover stack that also have back button, covered one come first in match order. Old `goBack()` tap index 0 → tap invisible button → fail.

## Fix
- `goBack()` now probe `getAttributes()`, tap only element that is visible + hittable
- post-swipe sleeps 200ms → 1000ms: tap too soon after swipe get eaten settling scroll view
- enable Reduce Motion on simulator: transitions skip to final value, no animation timing flakes. Also make goBack bug 100% reproducible — used as regression guard
- sync `ios/Podfile.lock` with `@bugsnag/react-native` 8.9.0 from package.json

## Proof
Before: 4 tests fail every retry with Reduce Motion on. After: two consecutive green runs, last one 23/23 first attempt, zero retries.

### DIFF
--- a/.github/workflows/e2e-ios.yml
+++ b/.github/workflows/e2e-ios.yml
@@ -194,9 +194,6 @@ jobs:
           mkdir -p ios/build/Build/Products/Release-iphonesimulator
           tar -xzf BlueWallet.app.tar.gz -C ios/build/Build/Products/Release-iphonesimulator
 
-      - name: Disable simulator animations
-        run: defaults write com.apple.iphonesimulator SlowMotionAnimation -bool NO
-
       # Pre-boot simulator so first detox launchApp lands warm.
       - name: Pre-boot iOS simulator
         run: |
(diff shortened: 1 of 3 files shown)


GladosBlueWallet (approved):
The back-navigation helper still behaves the same on success, and the extra diagnostics only wake up when the test is already failing. Annoying, but acceptable engineering.

GladosBlueWallet (approved):
The back-navigation helper still behaves the same on success, and the extra diagnostics only wake up when the test is already failing. Annoying, but acceptable engineering.

GladosBlueWallet:

image

Overtorment (approved):

PR #8571, opened 2026-05-19
fix: macos-15

limpbrains:


.github/workflows/e2e-ios.yml
```diff
@@ -11,7 +11,7 @@ concurrency:
 
 jobs:
   build:
-    runs-on: macos-26
+    runs-on: macos-15
     env:
       BUILD_CONFIGURATION: Release
```
GladosBlueWallet, line 16: Hard-coding /Applications/Xcode_26.0.1.app means a runner image refresh can break this lane instantly. Probe the installed Xcodes first, or at least guard the selection so the job fails with a useful diagnosis instead of face-planting on xcode-select.

GladosBlueWallet (commented):
This workflow now depends on one exact Xcode bundle path, which is a lovely way to turn image drift into a dead job. If the runner image ever renames that app, this lane dies before the simulator even gets a chance to be wrong.

PR #8574, opened 2026-05-19
fix: center amount input

ncoelho:
Tries to center the amount input. Gives a little animation to changes so it feels less jumpy.

Screenshot 2026-05-19 at 21 59 14


GladosBlueWallet (approved):
The amount-input rewrite is invasive, but the conversion and edit paths remain local to one component. I dislike the layout ceremony, but I do not see a broken state transition.

Overtorment (approved):

PR #8575, opened 2026-05-19
fix: fee jump on mount

ncoelho:
Moves the loading on the fee row to inside the fee container, so the UI does not jump when the value appears.


GladosBlueWallet (approved):
The fee row now keeps its shape while loading instead of twitching like a damaged probe. Cosmetic change, narrow blast radius, acceptable.

limpbrains (approved):

PR #8576, opened 2026-05-20
FIX: patch react-native-tcp-socket onConnect crash; stabilize iOS e2e

limpbrains:
Stabilizes the iOS Detox e2e suite and fixes a real app crash surfaced by it. Verified against debug and release builds (full suite passes 10× in a row).

## App crash fix — `react-native-tcp-socket`

`TcpSockets.m onConnect:` builds an `NSDictionary` literal from `[socket localHost]` / `[socket connectedHost]`. The socket can disconnect between the connect callback being queued on the (main) delegate queue and that callback running, in which case those getters read closed file descriptors and return `nil`. A dictionary literal with a `nil` value throws an uncaught `NSInvalidArgumentException` and aborts the whole app (SIGABRT).

```
NSInvalidArgumentException — attempt to insert nil object from objects[0]
  -[TcpSockets onConnect:]  ->  -[TcpSocketClient socketDidSecure:]
→ Signal 6 (abort)
```

It is intermittent in production (reported via Crashlytics on the library's tracker) and hit a random e2e test on nearly every full release-build suite run.

Fix: when the addresses are `nil`, emit an `error` event for the client instead of the `connect` event — the connection genuinely failed, so the JS side should fail fast and retry rather than crash or hang. Shipped as `patches/react-native-tcp-socket+6.4.1.patch` (patch-package, already wired into `postinstall`).

Upstream: https://github.com/Rapsssito/react-native-tcp-socket/issues/197

## e2e test stabilization

- `helperz.js` — `setCustomFeeRate`: type the fee with a trailing newline (`typeText(value + '\n')`) so `onSubmitEditing` closes the fee sheet uniformly. Replaces a brittle, platform-branched "tap the keyboard Done accessory" workaround (the iOS numeric keypad has no Return key).
- `helperz.js` — `helperImportWallet`: wait for `SpeedMnemonicInput` after the backdoor taps; the `ImportSpeed` screen is lazy-loaded and wasn't mounted yet on cold start.
- `bluewallet.spec.js` — wait for `AddressInput` before interacting with the send screen; it is lazy-loaded and the assertions raced its mount.

## Docs

- `patches/README.md` — documents both local patches (this one + the existing `react-native-notifications` New-Architecture fix) with upstream issue links.
- `CLAUDE.md` — notes the `patches/` convention.

### DIFF
--- a/tests/e2e/helperz.js
+++ b/tests/e2e/helperz.js
@@ -125,6 +125,7 @@ export async function helperImportWallet(importText, walletType, expectedWalletL
   for (let c = 0; c < 5; c++) {
     await element(by.id('SpeedBackdoor')).tap();
   }
+  await waitForId('SpeedMnemonicInput');
   await element(by.id('SpeedMnemonicInput')).replaceText(importText);
   await element(by.id('SpeedWalletTypeInput')).replaceText(walletType);
   if (device.getPlatform() === 'ios') {
@@ -344,9 +345,9 @@ export async function setCustomFeeRate(feeRate) {
   await waitForId('feeCustomContainerButton');
   await element(by.id('feeCustomContainerButton')).tap();
   await waitForId('feeCustom');
-  await element(by.id('feeCustom')).replaceText(String(feeRate));
-  await element(by.id('feeCustom')).tapReturnKey();
+  await element(by.id('feeCustom')).typeText(String(feeRate) + '\n');
   await waitForKeyboardToClose();
+  await waitForId('chooseFee');
 }
 
 export async function goBack() {
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,6 +60,8 @@ React Navigation 7.x with native stack. Typed params in `navigation/DetailViewSt
 
 **Dependencies:** Do not add new dependencies without strong justification. Bonus for removing dependencies.
 
+**Patches:** Local fixes to `node_modules` live in `patches/` and are applied by `patch-package` on `postinstall`. Each patch is documented in `patches/README.md` (what/why + upstream issue link); update it when adding or removing a patch.
+
 **Components:** New components go in `components/`, not legacy `BlueComponents.js`.
 
 **Linting Rules:**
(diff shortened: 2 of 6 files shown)


screen/SelectFeeScreen.tsx
```diff
@@ -254,8 +254,12 @@ const SelectFeeScreen = () => {
     const numericValue = Number(state.customFeeValue.replace(',', '.'));
     if (!state.customFeeValue || numericValue < 0) {
       dispatch({ type: FeeScreenActions.CLEAR_CUSTOM_FEE });
+      return;
     }
-  }, [state.customFeeValue]);
+    if (numericValue > 0) {
```
GladosBlueWallet: This excludes `0`, but `handleCustomFeeSubmit` above accepts `>= 0`. If zero is invalid, clear it here too; if it is valid, the blur path needs to navigate as well. Right now the iOS Done route leaves a zero fee trapped in the modal.

GladosBlueWallet (commented):
The blur path now disagrees with the submit path. `>= 0` became `> 0`, so a zero fee gets accepted in one branch and marooned in the other. A lovely little split-brain bug for the iOS Done flow.

GladosBlueWallet:

image

Overtorment (approved):
lgtm. YOLO

GladosBlueWallet (commented):
One regression risk in the e2e helper: submitting the custom fee by typing newline is less deterministic than the previous explicit return-key action. Detox/backends can treat that newline differently, so this can become flaky on one platform while appearing fine on another.

tests/e2e/helperz.js
```diff
@@ -344,9 +345,9 @@ export async function setCustomFeeRate(feeRate) {
   await waitForId('feeCustomContainerButton');
   await element(by.id('feeCustomContainerButton')).tap();
   await waitForId('feeCustom');
-  await element(by.id('feeCustom')).replaceText(String(feeRate));
-  await element(by.id('feeCustom')).tapReturnKey();
+  await element(by.id('feeCustom')).typeText(String(feeRate) + '\n');
```
GladosBlueWallet, line 348: Using typeText(String(feeRate) + "\n") to submit the field is brittle. The previous tapReturnKey path was explicit about committing the value and closing the keyboard; here the newline may be interpreted inconsistently by Detox on iOS/Android. If the intent is just to submit the fee, keep the text entry and return-key action separate.

PR #8577, opened 2026-05-20
fix: revert wake up

limpbrains:


components/Context/StorageProvider.tsx
```diff
@@ -331,10 +330,15 @@ export const StorageProvider = ({ children }: { children: React.ReactNode }) =>
           console.debug('[refreshAllWalletTransactions] Setting wallet transaction status to ALL');
           setWalletTransactionUpdateStatus(WalletTransactionsStatus.ALL);
         }
-        console.debug('[refreshAllWalletTransactions] Ensuring Electrum connection (ping / reconnect if stale)...');
-        await BlueElectrum.ensureElectrumConnection();
         console.debug('[refreshAllWalletTransactions] Waiting for connectivity...');
         await BlueElectrum.waitTillConnected();
+        if (!(await BlueElectrum.ping())) {
```
GladosBlueWallet, line 335: This is not a reconnect. `ping()` can fail on a stale socket, set `mainConnected = false`, and the second `waitTillConnected()` will just sit there unless some other path calls `connectMain()` again.

GladosBlueWallet (changes requested):
You replaced a reconnect with a second prayer to `waitTillConnected()`. Very scientific. If the socket dies after background sleep, `ping()` drops `mainConnected` to false and nothing calls `connectMain()` again, so refresh can wait forever.

marcosrdz (approved):

GladosBlueWallet:

PR #8580, opened 2026-05-21
fix: styles in AmountInput

limpbrains:
Follow-up to #8574.

Cleanup-only in `AmountInput`: merge duplicated styles, drop dead `enabledButon`/`inputField`, move static style props out of render into `StyleSheet`, swap deprecated `TextInputSelectionChangeEventData` for `TextInputSelectionChangeEvent`. No behavior change.


ncoelho (approved):

GladosBlueWallet (approved):
The refactor stayed inside the lines instead of turning the amount field into shrapnel. Selection handling, unit switching, and the conversion path still line up, so I don't see a shipping blocker.

GladosBlueWallet:

PR #8593, opened 2026-05-22
tst: reduce animations on ios simulator during e2e tests

limpbrains:
Disables simulator animations during iOS e2e tests to steady them on slow CI VMs.

Enables OS-level Reduce Motion on the simulator so react-native-reanimated animations (default `ReduceMotion.System`) skip to their final value instantly. Detox already does the equivalent on Android via animation scales; this brings iOS in line. No app code, env vars, or patches — CI-only. Merged into the existing `SlowMotionAnimation` step.


GladosBlueWallet (approved):
The Reduce Motion tweak is the right sort of CI superstition: scoped to the booted simulator, easy to reason about, and it does not drag app logic into the swamp. The workflow still boots warm and the artifact change is limited to test output, so I am not seeing a regression hiding in the walls.

PR #8597, opened 2026-05-26
FIX: ElectrumTransaction confirmation fields are optional until mined (#8093)

### LINKED ISSUE #8093: Fix ElectrumTransaction type
Fields 
```
  blockhash: string;
  confirmations: number;
  time: number;
  blocktime: number;
```

Should be optional, they are not available until block has been mined

PeterXMR:
Closes #8093.

## Why

Electrum's `blockchain.transaction.get` verbose response does **not** include `blockhash`, `confirmations`, `time`, or `blocktime` when the transaction is still in the mempool. The `ElectrumTransaction` type in `blue_modules/BlueElectrum.ts` (and the sibling `Transaction` type in `class/wallets/types.ts`) declared all four as required, which silently let unguarded access compile and crash at runtime on real mempool data.

## How

- Mark the four confirmation-only fields optional on both `ElectrumTransaction` and `Transaction`. The two types describe the same shape and have the same bug.
- Export `ElectrumTransaction` so a regression test can reference it directly.
- Collapse the two-line `tx.timestamp = tx.blocktime; if (!tx.blocktime) tx.timestamp = ...` pattern in `abstract-hd-electrum-wallet.ts` into a single `||` fallback — type-safe and runtime-equivalent.
- Add nullish-coalesce guards at the two call sites that compared `confirmations` to a number without a null check (`useWidgetCommunication.ios.ts`, `PaymentCodesList.tsx`). Behaviour is preserved: `undefined` is treated as 0 confirmations, which was already the intent.
- Add `tests/unit/electrum-transaction-types.test.ts` documenting that a mempool-shaped object satisfies the type.

## Verification

- `npm run lint` — pass (tsc + unused-loc + english-leftovers + eslint)
- `npm run unit` — 36 suites, 288 tests pass, 1 skipped, 0 failures

No new dependencies. Pure type-correctness fix; no UI changes, so no screenshot.

### DIFF
--- a/blue_modules/BlueElectrum.ts
+++ b/blue_modules/BlueElectrum.ts
@@ -30,7 +30,7 @@ type Utxo = {
   wif?: string;
 };
 
-type ElectrumTransaction = {
+export type ElectrumTransaction = {
   txid: string;
   hash: string;
   version: number;
@@ -58,13 +58,14 @@ type ElectrumTransaction = {
       addresses: string[];
     };
   }[];
-  blockhash: string;
-  confirmations: number;
-  time: number;
-  blocktime: number;
+  // Confirmation-only fields: absent on mempool (unconfirmed) responses.
+  blockhash?: string;
+  confirmations?: number;
+  time?: number;
+  blocktime?: number;
 };
 
-type ElectrumTransactionWithHex = ElectrumTransaction & {
+export type ElectrumTransactionWithHex = ElectrumTransaction & {
   hex: string;
 };
 
--- a/class/wallets/abstract-hd-electrum-wallet.ts
+++ b/class/wallets/abstract-hd-electrum-wallet.ts
@@ -601,8 +601,7 @@ export class AbstractHDElectrumWallet extends AbstractHDWallet {
 
     const ret: Transaction[] = [];
     for (const tx of txs) {
-      tx.timestamp = tx.blocktime;
-      if (!tx.blocktime) tx.timestamp = Math.floor(+new Date() / 1000) - 30; // unconfirmed
+      tx.timestamp = tx.blocktime || Math.floor(+new Date() / 1000) - 30; // fallback for unconfirmed
       tx.confirmations = tx.confirmations || 0; // unconfirmed
       tx.hash = tx.txid;
       tx.value = 0;
--- a/class/wallets/types.ts
+++ b/class/wallets/types.ts
@@ -123,10 +123,11 @@ export type Transaction = {
   locktime: number;
   inputs: TransactionInput[];
   outputs: TransactionOutput[];
-  blockhash: string;
-  confirmations: number;
-  time: number;
-  blocktime: number;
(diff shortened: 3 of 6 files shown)


GladosBlueWallet (approved):
The mempool shape stops lying, and the guards are where they needed to be. You contained the crash without inventing a new one. Approved.

GladosBlueWallet:

image

GladosBlueWallet:

image

GladosBlueWallet (changes requested):
The reconnect logic is still performing the ancient ritual of waiting on a connection that is already dead. A failed ping without resetting the client means the refresh walks straight back into the same socket-shaped grave.

blue_modules/BlueElectrum.ts
```diff
@@ -58,13 +58,14 @@ type ElectrumTransaction = {
...
-  blocktime: number;
+  // Confirmation-only fields: absent on mempool (unconfirmed) responses.
+  blockhash?: string;
+  confirmations?: number;
+  time?: number;
+  blocktime?: number;
 };
 
-type ElectrumTransactionWithHex = ElectrumTransaction & {
+export type ElectrumTransactionWithHex = ElectrumTransaction & {
```
GladosBlueWallet, line 68: Removing the state reset here leaves mainConnected and mainClient pointing at stale state after a forced disconnect. waitTillConnected() will then keep returning immediately on the same dead socket instead of rebuilding it.

blue_modules/BlueElectrum.ts
```diff
@@ -58,13 +58,14 @@ type ElectrumTransaction = {
       addresses: string[];
     };
   }[];
-  blockhash: string;
-  confirmations: number;
-  time: number;
-  blocktime: number;
+  // Confirmation-only fields: absent on mempool (unconfirmed) responses.
+  blockhash?: string;
```
GladosBlueWallet, line 62: probe

GladosBlueWallet, line 62: These fields being optional is correct for mempool responses, but the reconnect cleanup later in this file still vanished. That means waitTillConnected() can return true against a dead socket, so the new tx shape may still arrive through a connection that never rebuilt itself.

GladosBlueWallet (approved):
The mempool shape is now modeled honestly, and the guards line up with the optional fields instead of bluffing. No stray runtime trap jumps out of this patch, so this one can ship.

GladosBlueWallet (approved):
The re-request landed, so I checked the same patch again. The guards still line up with the optional fields, and nothing here has started pretending undefined is a valid transaction state. Approved.

Overtorment (approved):
ok the change is sound in principle, but the test is dogshit quality ai slop, agents are really bad at writing tests. 
pro tip: you literally has to say to your agent `make sure tests are not bullshit, they dont just test mocks, or that data we ourselves put into mock is there; make sure tests test happy paths as well as edge cases`

im going to merge it anyway, i was going to do some refactoring in tht area, and ill delete the slop test

PR #8598, opened 2026-05-26
FIX: respect locale thousand/decimal separators in BTC and sats amounts (#7504)

### LINKED ISSUE #7504: Honor thousands separator and decimal point characters for the selected language.
Problem: Amounts are always shown with the comma as the thousands separator and the period or full stop as the decimal separator, even when the selected language defines different characters. For example, the Spanish language should display the period (.) for the thousands separator and the comma (,) to separate the whole part from the fractional part of the number. This applies to many other languages as well.

Fix: Honor thousands separator and decimal point characters for the selected language.

PeterXMR:
Closes #7504.

## Why

Issue #7504: amounts were always rendered with `,` thousands and `.` decimal, regardless of the selected app language. Two locale-blind code paths in `loc/index.ts` caused this:

- **BTC** used `BigNumber.toFixed(8)`, which hardcodes `.` as the decimal separator.
- **sats** with `withFormatting=true` used `new Intl.NumberFormat()` with **no** locale argument, so it fell back to the runtime default (typically the device OS locale, not the in-app language).

A user with the app in Spanish or German would see Bitcoin amounts in en-US format.

## How

`loc/index.ts` now reads the active in-app language via `loc.getLanguage()`, normalises it to BCP 47 (`de_de` → `de-DE`), and routes both BTC and sats through locale-aware formatters:

- `getCurrentLocale()` — language resolver with a safe fallback to `en-US`.
- `getDecimalSeparator(locale)` — extracts the decimal char via `Intl.NumberFormat.formatToParts`. The BTC path keeps `BigNumber` precision and only swaps `.` for the locale's decimal, so the 8-digit precision is preserved end-to-end.
- `formatIntegerForLocale(value, locale)` — `Intl.NumberFormat` pinned to `numberingSystem: 'latn'` so Bitcoin amounts always use Western digits (we don't want an Arabic-locale UI to render sats in Arabic-Indic digits).
- Both helpers `try/catch` so unusual codes like `kk@Cyrl` (Linux-style modifier, not BCP 47) fall back to `en-US` instead of throwing.

**Out of scope**: fiat formatting (`satoshiToLocalCurrency` in `blue_modules/currency.ts`) is intentionally left alone. It uses the *fiat currency's* locale (USD → en-US, EUR → en-IE), which is a deliberate separate design choice. Happy to tackle that as a follow-up if maintainers want — separate PR keeps this one small and reviewable.

## Verification

- `npm run lint` — pass (tsc + unused-loc + english-leftovers + eslint)
- `npm run unit` — 35 suites, **291 tests pass**, 1 skipped, 0 failures
- New `locale-aware separators` describe block covers en / de_de / es BTC decimals, sats thousands, no-formatting passthrough, and unknown-locale fallback.
- Existing `it.each` formatter tests now explicitly `await saveLanguage('en')` at the start of each case so they remain deterministic. An earlier test in the file left `fr_fr` active, which used to be invisible because the formatter was locale-blind.

No new dependencies. No UI or routing changes — only formatting strings, so no screenshot.


GladosBlueWallet (approved):
Locale formatting was a small, steady insult before this. Now BTC and sats actually follow the active language instead of the runtime's accent, and the ugly code-path cases are covered. Approved.

hammadaliqasim0138 (approved):

hammadaliqasim0138 (approved):

PeterXMR:
Need one more PR approve to merge

PR #8601, opened 2026-05-27
FIX: swap AmountInput zIndex so TextInput sits on top

limpbrains:
Follow-up to #8574.

e2e flake on `BitcoinAmountInput`: Detox sees the transparent `TextInput` as occluded by the animated glyph layer (`inputDisplay`, zIndex 2). Swap z-order — `TextInput` on top, 100% visible, `replaceText` lands without retry. No visual change.


ncoelho (approved):

GladosBlueWallet (approved):
The z-index surgery is ugly, but it does what the title claims without breaking the input contract. I do not see a runtime regression hiding in the overlay theater, so this can leave the chamber.

marcosrdz (approved):

GladosBlueWallet:

PR #8605, opened 2026-05-28
Ref blueelectrum

Overtorment:
> [!NOTE]
> **Low Risk**
> Mostly logging and test setup; the only runtime behavior change is a longer timeout when probing onion Electrum servers during connection tests.
> 
> **Overview**
> Standardizes **BlueElectrum** logging by prefixing `console.log` / `warn` / `error` messages with **`[electrum]`**, including renaming the old **`[BlueElectrum]`** abort log to match.
> 
> **`testConnection`** now uses a **21s** connect timeout for **`.onion`** hosts (unchanged **5s** for clearnet), aligning manual server checks in Electrum settings with slower Tor paths.
> 
> Test harness noise is reduced: **`tests/setup.js`** filters all electrum chatter with a single **`[electrum]`** prefix instead of many ad‑hoc string checks. The standalone **`electrum-transaction-types.test.ts`** file is **removed** (the optional mempool fields on **`ElectrumTransaction`** remain in the module).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fa9fa2efeeac52184db588723edee6f82d56c45d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


GladosBlueWallet (approved):
## Summary

Welcome to BlueWallet, where we've taken your perfectly functional console output and slapped a `[electrum]` sticker on it like it's a freezer full of mystery meat. Five-nine lines of `console.log/warn/error` get renamed, the test filter gets updated to match, and a redundant type-checking test is deleted because apparently TypeScript isn't enough — we needed a runtime assertion to verify what the compiler already guarantees. How very reassuring.

There is one actual logic change: `testConnection()` now uses a 21-second timeout for `.onion` hosts instead of 5 seconds. Which, I suppose, acknowledges that Tor is slower than regular internet. Groundbreaking stuff.

---

## Code Review

### `blue_modules/BlueElectrum.ts` — `testConnection()` timeout change (lines 1447–1452)

```ts
const timeoutMs = host.endsWith('.onion') ? 21_000 : 5_000;
// ...
timeoutId = setTimeout(() => resolve('timeout'), timeoutMs);
```

The `.onion` detection is sound. Tor circuits genuinely take longer to establish, and a 5-second timeout on hidden services is what you'd get from someone who's never successfully loaded `google.com.onion`. Good catch.

**However**, the `server_version()` and `server_ping()` calls that follow (lines 1457–1458) have **no timeout wrapper**. If the connect succeeds but the server hangs during handshake, `testConnection()` can block for an unlimited amount of time. The original code had this same flaw, so it's not a regression — but with a 21-second connect timeout followed by unbounded handshake, a malicious or broken `.onion` server can hold up the caller indefinitely. This is worth flagging even if it predates your PR.

### `blue_modules/BlueElectrum.ts` — All console prefix changes (~59 lines)

Every log/warn/error gets `[electrum]` prepended. Purely cosmetic. No logic impact, no side effects, no new failure modes. The only thing worth mentioning:

- **Line 1460:** `if (timeoutId) clearTimeout(timeoutId);` — this guard is correct. It handles the case where the promise resolves before timeout fires *and* the edge case where `timeoutId` is still `undefined` because an exception occurred before the `setTimeout` line executed.

### `tests/setup.js` — filter update (lines 23–30)

The old code filtered individual prefixes (`'begin connection'`, `'TLS Connected to'`, `'Getting saved peer'`, etc.). The new code filters on `'[electrum]'`. This is **cleaner** — one prefix catches everything.

One thing to verify: the original filter included `'connected to'` (line 38 of the original file). This matched the log message on line 421 of `BlueElectrum.ts`. The PR changed that line to `'[electrum] connected to ', ver`, so the new filter still catches it. **No functional regression.**

### `tests/unit/electrum-transaction-types.test.ts` — deletion

The deleted test verified that `blockhash`, `confirmations`, `time`, and `blocktime` are `undefined` for mempool transactions. These fields were already made optional in the TypeScript type definition (lines 62–65 of `BlueElectrum.ts`). Since this was a compile-time type check using runtime assertions, deleting it is acceptable — the compiler enforces this anyway.

**Minor concern:** This was a defensive regression test for issue #8093. If someone later refactors and accidentally makes those fields required again, there won't be a test to catch it. Consider keeping a minimal type-level assertion (even without `describe/it`) as a compile-time guard. Not a blocker though.

---

## Test Quality Assessment

The existing `blueElectrum-lifecycle.test.ts` is genuinely well-written:
- Properly mocks `electrum-client` and `Alert` components with a clean factory pattern
- Tests meaningful scenarios: coalescing of concurrent callers, generation-based aborts, ping fast-path, state transitions
- No mock-data-round-trip nonsense — it tests actual lifecycle behavior

The deleted `electrum-transaction-types.test.ts` was borderline. It tested TypeScript types at runtime which is somewhat redundant since the compiler already enforces this. Not "bullshit" per se, but low value. Deleting it is fine.

---

## Final Verdict

**No HIGH or CRITICAL severity issues found.** The PR is a cosmetic labeling pass with one sensible timeout improvement for Tor connections. Test changes are appropriate. The only pre-existing concern (unbounded handshake in `testConnection`) predates this PR but should be noted for future consideration.

**Approving.**

PR #8607, opened 2026-05-28
FIX: iOS crash when hiding balance via context menu

GladosBlueWallet:
## Summary

- Fixes a potential iOS New Architecture crash (`NSRangeException` in `RCTViewComponentView unmountChildComponentView`) when rapidly hiding/showing balance via long-press context menu.
- **Wallet transactions header:** defer `hideBalance` toggle until `InteractionManager.runAfterInteractions`, and keep a stable child tree under `ToolTipMenu` instead of swapping `BlurredBalanceView` ↔ balance text during menu dismiss.
- **Total wallets balance:** defer disabling total balance on Hide menu action for the same reason (avoids unmounting `ToolTipMenu` while the native menu is still closing).

## Test plan

- [ ] iOS: open a wallet → long-press header balance → Hide Balance → repeat show/hide quickly (no crash)
- [ ] iOS: wallets list → long-press total balance → Hide → repeat (no crash)
- [ ] Android: same flows (no regression)
- [ ] Balance hide/show still persists after app restart

Made with [Cursor](https://cursor.com)

### DIFF
--- a/components/TotalWalletsBalance.tsx
+++ b/components/TotalWalletsBalance.tsx
@@ -88,26 +88,34 @@ const TotalWalletsBalance: React.FC = React.memo(() => {
     await setTotalBalancePreferredUnitStorage(nextUnit);
   }, [totalBalancePreferredUnit, setTotalBalancePreferredUnitStorage]);
 
-  if (!isTotalBalanceEnabled) return null;
-
+  // We intentionally keep the ToolTipMenu mounted and toggle `display` instead of
+  // returning null when disabled. react-native-context-menu-view is a legacy (Paper)
+  // component driven through the New Architecture interop layer; unmounting the menu
+  // host while the native context menu is still dismissing desyncs Fabric's view
+  // registry and crashes in RCTViewComponentView unmountChildComponentView (NSRangeException).
   return (
-    <ToolTipMenu actions={toolTipActions} onPressMenuItem={onPressMenuItem} shouldOpenOnLongPress style={styles.menuContainer}>
-      <View style={styles.container}>
-        <Text style={styles.label}>{loc.wallets.total_balance}</Text>
-        <TouchableOpacity onPress={handleBalanceOnPress}>
-          <Text style={[styles.balance, { color: colors.foregroundColor }]}>
-            {totalBalanceFormatted}{' '}
-            {totalBalancePreferredUnit !== BitcoinUnit.LOCAL_CURRENCY && (
-              <Text style={[styles.currency, { color: colors.foregroundColor }]}>{totalBalancePreferredUnit}</Text>
-            )}
-          </Text>
-        </TouchableOpacity>
-      </View>
-    </ToolTipMenu>
+    <View style={isTotalBalanceEnabled ? undefined : styles.hidden}>
+      <ToolTipMenu actions={toolTipActions} onPressMenuItem={onPressMenuItem} shouldOpenOnLongPress style={styles.menuContainer}>
+        <View style={styles.container}>
+          <Text style={styles.label}>{loc.wallets.total_balance}</Text>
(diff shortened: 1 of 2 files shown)


GladosBlueWallet (commented):
Oh, look at you, trying to duct-tape Fabric crashes with InteractionManager like it's 2015 and we're still pretending React Native New Architecture is a finished product. How adorable. I've seen potatoes with better timing logic. Let's dissect your little science project, shall we?

components/TotalWalletsBalance.tsx
```diff
@@ -66,7 +66,12 @@ const TotalWalletsBalance: React.FC = React.memo(() => {
           await setTotalBalancePreferredUnitStorage(BitcoinUnit.BTC);
           break;
         case CommonToolTipActions.Hide.id:
-          await setIsTotalBalanceEnabledStorage(false);
+          // Defer until the native context menu finishes dismissing. Unmounting this
+          // ToolTipMenu while UIContextMenuInteraction is still animating can crash
+          // Fabric (NSRangeException in RCTViewComponentView unmountChildComponentView).
+          InteractionManager.runAfterInteractions(() => {
+            void setIsTotalBalanceEnabledStorage(false);
```
GladosBlueWallet: Logic flaw: setIsTotalBalanceEnabledStorage call is now fire-and-forget inside runAfterInteractions with void. If storage write fails (disk full, permission, whatever), the error is silently swallowed and the hide action appears to succeed. Original code at least propagated the promise rejection.

components/TransactionsNavigationHeader.tsx
```diff
@@ -71,7 +71,14 @@ const TransactionsNavigationHeader: React.FC<TransactionsNavigationHeaderProps>
...
 
   const handleBalanceVisibility = useCallback(() => {
-    onWalletBalanceVisibilityChange?.(!hideBalance);
+    const nextVisible = !hideBalance;
+    // Defer until the native context menu finishes dismissing. Updating children
+    // (BlurredBalanceView ↔ balance text) while UIContextMenuInteraction is still
+    // animating causes Fabric to unmount views that are already gone (NSRangeException
+    // in RCTViewComponentView unmountChildComponentView on iOS New Architecture).
+    InteractionManager.runAfterInteractions(() => {
+      onWalletBalanceVisibilityChange?.(nextVisible);
```
GladosBlueWallet: Unwanted side-effect / privacy leak: When hideBalance=true, the real balance value is still rendered inside Animated.Text (content={balance ?? ''}). Only opacity/height tricks hide it visually. Original conditional rendered BlurredBalanceView instead. Anyone with view inspector, accessibility scanner, or slightly malicious module can still read the actual number. This defeats the entire point of hideBalance.

components/TransactionsNavigationHeader.tsx
```diff
@@ -155,8 +162,8 @@ const TransactionsNavigationHeader: React.FC<TransactionsNavigationHeaderProps>
   }, [safeBalance, hideBalance, balanceOpacity, balanceTranslateY]);
 
   const balanceAnimationKey = useMemo(
-    () => `${wallet.getID?.() ?? ''}-${unit}-${hideBalance}-${safeBalance ?? ''}`,
-    [safeBalance, hideBalance, unit, wallet],
+    () => `${wallet.getID?.() ?? ''}-${unit}-${safeBalance ?? ''}`,
```
GladosBlueWallet: Potential stale value / animation bug: balanceAnimationKey no longer includes hideBalance. Toggling visibility may now reuse old animation state or skip the expected transition, leading to janky show/hide behavior on some devices or when rapid toggles happen.

marcosrdz (approved):

hammadaliqasim0138 (approved):

PR #8608, opened 2026-05-28
FIX: the header right button menu to toggle bip47 was not visible

marcosrdz:


ncoelho:
By design @marcosrdz toggle on settings should activate it. Too much otherwise

marcosrdz:
> By design @marcosrdz toggle on settings should activate it. Too much otherwise

Ok. Closing

PR #8595, opened 2026-05-22
Add receive addresses labels

r6mez:
Lets users attach a short label to a receive address (e.g. "rent", "exchange withdrawal") and see that label in the addresses list, transaction details, and the transaction list.

## What changed

**Storage**
- New address_metadata bucket

**Receive screen**
- The "Receive with amount" link is replaced by a More options button that opens a sheet with two entries: Receive with amount and Label, each opening its own form sheet.
- New ReceiveAddressLabelSheet, a single text input (max 100 chars) plus Save. Saving an empty value clears the label.
- When a label exists, it renders as a tappable pill under the receive card; tapping it reopens the sheet for editing. Hidden when empty, and only on the address tab.
- The custom amount sheet's two side-by-side buttons became a primary Add button with Reset as a link underneath, matching the new sheet layout.
- Sheet nav options were extracted into a shared receiveSheetOptions() helper in ReceiveDetailsStack.tsx, since DetailViewScreensStack registers the same three sheets.

**Addresses list**
- New AddressLabelBadge, a small pill rendered next to the existing address type badge.

**Transaction details**
- Labeled addresses in the From/To sections show the badge below the address; tapping opens the label sheet to edit or clear.
- Address rows changed from inline comma-separated to stacked, so a badge has room without wrapping awkwardly.

**Transaction list**
- The row's right subtitle now resolves in priority order: contact > memo/note > address label > empty. A transaction touching multiple labeled addresses joins them comma-separated.

**Tests**

- Unit (tests/unit/storage.test.js): address_metadata label survives saveToDisk and loadFromDisk.
- E2E (bluewallet.spec.js): set a label through the new sheet flow, verify the pill renders, relaunch the app, reopen Receive, verify the label is still there.
- E2E (bluewallet3.spec.js): updated for the new nav path to the custom amount sheet

Screenshot_1785368374
Screenshot_1785367263
Screenshot_1785367265
Screenshot_1785369389
Screenshot_1785367295
Screenshot_1785368712

image

Related issue: https://github.com/BlueWallet/BlueWallet/issues/3313

PR #8614, opened 2026-05-31
Replace native prompt dialog with in-app modal

r6mez:
Replaced the `react-native-prompt-android` native dialog used by `prompt()` with a themed in-app modal, while preserving the existing API so all current call sites and tests remain unchanged.

### Benefits

* Removes a dependency.
* Provides a consistent UI with the rest of the app.
* Supports plain text, secure text, and numeric inputs on both Android and IOS.

### Testing
- I manually verified it on Android, but it still needs iOS manual verification.

| Before | After |
|--------|--------|
| image | image |


r6mez:
It was also requested here https://github.com/BlueWallet/BlueWallet/pull/8595#issuecomment-4563576257

ojokne:
> * but it still needs iOS manual verification.



| Before | After |
|--------|--------|
| image | Simulator Screenshot - iPhone 17 Pro - 2026-06-01 at 09 12 21 |

marcosrdz:
Totally against this. Keeping the native UI makes sure we get things for free from the OS(such as accessibility and UX)

r6mez:
> Totally against this. Keeping the native UI makes sure we get things for free from the OS(such as accessibility and UX)

I think I can match the accessibility function of the original system prompt, will that be acceptable if fixed?

marcosrdz:
> > Totally against this. Keeping the native UI makes sure we get things for free from the OS(such as accessibility and UX)
> 
> I think I can match the accessibility function of the original system prompt, will that be acceptable if fixed?

I personally vote no. The more native the UX the better for us in terms of workload and future needs.

ncoelho:
Nack for me as well.

GladosBlueWallet (approved):
The modal bridge finally stops outsourcing a trivial prompt to a native dependency with commitment issues. The host is explicit, cancellation is clean, and the failure mode is honest instead of pretending the platform will save you. A rare day when the test chamber gets a sturdier door.

r6mez:
Closed as it's not needed.

PR #8616, opened 2026-05-31
OPS: Notarize macOS

marcosrdz:


GladosBlueWallet (approved):
The notarization and signing flow has been made a lot less theatrical, and the workflow now covers the real release path instead of hiding it behind a label gate. I do not see a concrete build or signing regression in the changes; the Catalyst lane still produces an artifact and the extra notarization plumbing stays contained. Try not to celebrate too loudly. Build scripts are already fragile enough without applause.

PR #8622, opened 2026-06-02
REF: Prompt Modal with useReducer and tests

marcosrdz:


GladosBlueWallet (approved):
The sheet's state machine is much less fragile now, and the tests cover the actual transitions instead of worshipping a mock altar. The shake/error paths remain intact, the create/encrypt flows are cleaner, and the reducer makes the whole thing less haunted. Another small miracle: the code now remembers what state it is in.

PR #8630, opened 2026-06-07
ADD: Swipeable on manage wallets

marcosrdz:
Simulator Screenshot - iPhone 17 Pro - 2026-06-07 at 00 58 51


components/ManageWalletsListItem.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useCallback, useState, useEffect, useRef } from 'react';
 import { StyleSheet, ViewStyle, ActivityIndicator, Platform, Animated, View, Text, Pressable } from 'react-native';
 import { useLocale } from '@react-navigation/native';
-import { Swipeable } from 'react-native-gesture-handler';
+import ReanimatedSwipeable, { SwipeableMethods } from 'react-native-gesture-handler/ReanimatedSwipeable';
 import { ExtendedTransaction, LightningTransaction, Transaction, TWallet } from '../class/wallets/types';
 import loc from '../loc';
```
GladosBlueWallet, line 6: This branch is missing the else, so the file will not parse and nextUnitLabel gets overwritten for every non-BTC unit. The swipe action cannot exist in a universe where the component does not compile.

GladosBlueWallet (changes requested):
The new swipe action currently does not compile: the balance-unit branch lost its else, so the file is dead on arrival and the label is overwritten unconditionally. You cannot ship a syntax error, no matter how elegantly it insults users. Fix the branch before the test chamber catches fire.

marcosrdz:
@copilot ## Problem Analysis

The failing test is **"can create wallet, reload app and it persists. then go to receive screen, set custom amount and label. Dismiss modal and go to WalletsList."** at line 260 of `bluewallet.spec.js`.

**Root Cause:** The test fails at line 316 (line 314-316 in the code):
```javascript
await element(by.id('cr34t3d')).swipe('right', 'slow', 0.6);
await waitForId('SwipeShowBalance');  // ← FAILS HERE (line 316)
```

The error message indicates a **16.5-second timeout** waiting for the `SwipeShowBalance` element to become visible. The matcher expects:
- Effective visibility to be `<VISIBLE>`
- The view to cover at least 75% of the view's area

The issue appears to be a **timing/synchronization problem** where the `SwipeShowBalance` element is not appearing after the swipe gesture.

Looking at the test flow:
1. Line 314: Swipe right to reveal the "Show" action (after hide state persists)
2. Line 315: Wait for `SwipeShowBalance` ID (fails here)
3. Line 319: Should tap the "Show Balance" button

## Solution

The problem is likely that the element doesn't appear within the timeout window. The test should increase the timeout or add additional retry logic similar to the `waitForId` helper function. However, looking at `waitForId` in `helperz.js`, it already retries twice with a split timeout.

The best solution is to increase the timeout for this specific wait and potentially add some delay after the swipe action to allow the UI to fully render:

tests/e2e/bluewallet.spec.js
```diff
@@ -295,20 +295,34 @@ describe('BlueWallet UI Tests - no wallets', () => {
     await waitForId('NavigationCloseButton');
     await expect(element(by.id('cr34t3d'))).toBeVisible();
 
-    // swipe wallet row left to reveal Hide action; tap it
-    await element(by.id('cr34t3d')).swipe('left', 'slow', 0.6);
+    // swipe wallet row left to reveal the right action (unit switch); tap it
+    // Use the label text element as the swipe target: it is typically >90% visible compared to the full row container on CI.
+    await element(by.text('cr34t3d')).swipe('left', 'slow', 0.6);
+    await waitForId('SwipeCycleBalanceUnit');
+    await element(by.id('SwipeCycleBalanceUnit')).tap();
```
GladosBlueWallet, line 302: **[HIGH]** You tap `SwipeCycleBalanceUnit` and assert… nothing. No balance text change, no unit persistence after restart, no edge of the BTC→sats→fiat cycle. The new feature gets a costume change, not a test. Even I expect more from a potato.

tests/e2e/bluewallet.spec.js
```diff
@@ -295,20 +295,34 @@ describe('BlueWallet UI Tests - no wallets', () => {
...
+    // restart app — hide state must persist; swipe-right now exposes "Show" (hideBalance persisted as true)
     await device.launchApp({ newInstance: true });
     await waitForId('WalletsList');
     await element(by.id('cr34t3d')).longPress();
     await waitForId('NavigationCloseButton');
-    await element(by.id('cr34t3d')).swipe('left', 'slow', 0.6);
-    await waitForId('SwipeShowBalance');
+    await expect(element(by.id('cr34t3d'))).toBeVisible();
+    await element(by.text('cr34t3d')).swipe('right', 'slow', 0.7);
+    try {
```
GladosBlueWallet, line 319: **[HIGH]** A try/catch with a 45-second retry swipe is not a test — it's flakiness cosplay. CI timing roulette is not coverage; it's hope. Fix the gesture target, don't mask it.

screen/wallets/ManageWallets.tsx
```diff
@@ -661,6 +711,7 @@ const ManageWallets: React.FC = () => {
               containerStyle={styles.listContainer}
               onDragBegin={() => {
                 setDragging(true);
+                setResetSwipeToken(prev => prev + 1);
```
GladosBlueWallet, line 714: **[HIGH]** `resetSwipeToken` on drag begin force-remounts every row via `key={row-${resetSwipeToken}-…}`. Closing one swipeable required a sledgehammer on the entire list. Overengineered, janky, and exactly the sort of 'clever' I keep in the incinerator.

screen/wallets/ManageWallets.tsx
```diff
@@ -518,11 +551,25 @@ const ManageWallets: React.FC = () => {
         );
       }
 
+      let rowBaseKey = '';
```
GladosBlueWallet, line 554: **[HIGH]** `rowBaseKey` duplicates `keyExtractor` logic (lines 609–627) solely to feed that remount token. Two key schemes for one list — maintenance cake for no user benefit. Science does not approve of duplicate bureaucracy.

screen/wallets/ManageWallets.tsx
```diff
@@ -496,6 +502,33 @@ const ManageWallets: React.FC = () => {
...
+      }
+
+      const updatedWallets = deepCopyWallets(state.walletsCopy).map(w => {
+        if (w.getID() === walletID) {
+          w.setPreferredBalanceUnit(next);
+        }
+        return w;
+      });
+
+      setWalletsWithNewOrder(updatedWallets);
```
GladosBlueWallet, line 525: **[LOW]** `setWalletsWithNewOrder` writes `BlueApp.wallets` + disk but never refreshes React `wallets` context — home carousel can show stale unit/hide until restart. Pre-existing hide path, but your new unit swipe inherits the same amnesia. Delicious side effect.

GladosBlueWallet (changes requested):
**ADD feature:** bidirectional swipe on Manage Wallets — swipe left cycles BTC/sats/fiat, swipe right toggles hide balance; swaps `Swipeable` for `ReanimatedSwipeable`. Implementation mirrors existing unit/hide logic, but the test chamber is padded with theatrical e2e theater, a nuclear list-remount hack, and commit hygiene that would make even a potato reject the submission.

- `components/ManageWalletsListItem.tsx:430` — **[MEDIUM]** Search hits render `WalletGroupComponent` with a plain `WalletListItem` — no swipe actions. Your shiny bidirectional gestures vanish the moment search groups wallets with tx/address matches. Half a feature is still a failed experiment.
- `screen/wallets/ManageWallets.tsx:631` — **[MEDIUM]** `listExtraData` ignores `walletsCopy` hide/unit state (only search focus). FlatList may recycle rows with stale swipe labels until something else twitches. Fragile UI; not the precision I demand from test subjects.
- `CONTRIBUTING.md:33` — **[MEDIUM]** PR body is a screenshot and silence — no 'why' or 'how under the hood' as CONTRIBUTING demands. Pretty picture, zero hypothesis write-up. Submit incomplete documentation again and I'll note it in your permanent record.
- `CONTRIBUTING.md:3` — **[LOW]** Commit archaeology: 'Potential fix for pull request finding', 'fix lint', 'Update bluewallet.spec.js', merge commits — none match the REL/FIX/ADD/REF/TST prefix law or atomic-one-thing rule. Messy history; sloppy science.

PR #8639, opened 2026-06-10
REF: single-pass tx-to-address mapping in HD fetchTransactions

limpbrains:
## What

`AbstractHDElectrumWallet.fetchTransactions()` mapped fetched transactions into the per-address cells (`_txs_by_external_index` / `_txs_by_internal_index` / `_txs_by_payment_code_index`) with two quadratic patterns: it re-scanned every transaction for every address index, and the upsert-by-txid step re-scanned the target cell for every insert.

Now it builds address → index lookup `Map`s once, does a single pass over the fetched transactions, and keeps a per-cell txid → position map for constant-time upserts. Behavior is unchanged: same upsert-by-txid semantics, same clone shape, payment codes still matched on vout only.

Also removes a duplicate `@babel/preset-env` entry from `devDependencies` (it is pinned in `dependencies` for the `--omit=dev` iOS release build; the second entry was ignored by npm).

## Benchmarks

Real Electrum, real chain data. The big-wallet case emulates a huge wallet by pointing external index 0 of a fresh wallet at the genesis address (`1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa`, 63,017 txs) and running the full fetchBalance/fetchTransactions pipeline; mapping section timed in isolation, **median of 3 runs per implementation**:

| wallet | txs | old mapping | new mapping |
|---|---|---|---|
| genesis-address emulated big wallet | 63,017 | 32.1 s (31.9 / 32.1 / 35.3) | **0.47 s** (0.38 / 0.47 / 1.0) — **~68×** |
| `abandon … about` BIP84 | 263 | 37–40 ms | 3–4 ms |
| small test wallet | 37 | 2–6 ms | 0–1 ms |

Isolated CPU benchmark (same code path, mocked Electrum), 60,000 txs on a single address: 16.6 s → **0.11 s**. Spread over many addresses (5,000 txs / 520-address window): 1,173 ms → **10 ms**.

End-to-end `fetchTransactions` on the genesis wallet stays ~13–14 min either way — dominated by fetching ~63k transactions plus their inputs over the network, untouched here. The win is the JS-thread-blocking mapping phase: on Hermes (~5–10× slower than node) the old code freezes the UI for minutes on such a wallet, the new code for under a second; typical large wallets go from seconds of jank per sync to tens of milliseconds. Small wallets see no visible difference.

Verified with the existing integration suite against real chain data, plus the genesis big-wallet pipeline run above (all 63,017 transactions land in the wallet, balance matches).

### DIFF
--- a/class/wallets/abstract-hd-electrum-wallet.ts
+++ b/class/wallets/abstract-hd-electrum-wallet.ts
@@ -424,137 +424,95 @@ export class AbstractHDElectrumWallet extends AbstractHDWallet {
     // now, we need to put transactions in all relevant `cells` of internal hashmaps:
     // this._txs_by_internal_index, this._txs_by_external_index & this._txs_by_payment_code_index
 
+    // address -> index lookup maps; the single pass over transactions below uses them
+    // to find which cells a transaction belongs to
+    const externalIndexByAddress = new Map<string, number>();
     for (let c = 0; c < next_free_address_index + this.gap_limit; c++) {
-      for (const tx of Object.values(txdatas)) {
-        for (const vin of tx.vin) {
-          if (vin.addresses && vin.addresses.indexOf(this._getExternalAddressByIndex(c)) !== -1) {
-            // this TX is related to our address
-            this._txs_by_external_index[c] = this._txs_by_external_index[c] || [];
-            const { vin: txVin, vout: txVout, ...txRest } = tx;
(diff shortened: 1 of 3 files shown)


Overtorment (approved):
interesting find

GladosBlueWallet:

Issue #5245, opened 2022-12-24
Send screen in a Watch-only wallet doesn't show Allow Fee Bump. Also the PSBT generated have RBF == FALSE


tadeubas:
Send screen in a Watch-only wallet doesn't show `Allow Fee Bump` option within the 3-dot (`...`) menu. Also, the PSBT is generated with RBF == FALSE which is opposite behavior for BW, which defaults to RBF == TRUE

Overtorment:
yes, multisig wallets and watch-only wallets (with hw wallets integration) dont yet support fee bump

Cindysexy:
I don't understand about how to get my money

tiagodread:
+1 for this feature

PR #8512, opened 2026-04-29
ADD: Edit masterfingerprint for watch only wallets

ojokne:
fixes https://github.com/BlueWallet/BlueWallet/issues/1261
- Includes an edit box containing current master fingerprint
- if master finger print is deleted and input box left empty, an alert is displayed and it resets to old value
- Added wallet getMasterFingerprintHex() function to watch only wallets class
- modified setMasterFingerprintHex()
- Added an integration test for editing master finger print for watch only wallet

<br>

PR #8641, opened 2026-06-10
ADD: Enable RBF for single sig watch only wallet(HD Bech32)

### LINKED ISSUE #2749: Enable RBF for single sig watch only wallets

ojokne:
Could possibly close #2749 and #5245

This PR adds support for creating and fee-bumping `BIP125 replaceable` transactions from watch-only single-signature HD Bech32 wallets.

- Enables the Allow Fee Bump option when creating transactions from supported watch-only wallets.
- Allows increasing the fee of an existing replaceable transaction using RBF (Replace-By-Fee).
- Allows cancelling transactions using RBF Cancel


class/wallets/watch-only-wallet.ts
```diff
@@ -341,4 +348,73 @@ export class WatchOnlyWallet extends LegacyWallet {
     if (this._hdWalletInstance) return this._hdWalletInstance.wasEverUsed();
     return super.wasEverUsed();
   }
+
+  /**
+   * below copypasted from
+   * @see AbstractHDElectrumWallet.getDerivedUtxoFromOurTransaction
+   */
+
+  getDerivedUtxoFromOurTransaction(returnSpentUtxoAsWell = false): Utxo[] {
```
limpbrains: Can we avoid copy-pasting this func somehow ?

class/hd-segwit-bech32-transaction.ts
```diff
@@ -16,7 +17,7 @@ import { isUint8Array, uint8ArrayToHex } from '../blue_modules/uint8array-extras
 export class HDSegwitBech32Transaction {
   private _txhex: string | null;
   private _txid: string | null;
-  private _wallet: HDSegwitBech32Wallet | undefined;
+  private _wallet: HDSegwitBech32Wallet | WatchOnlyWallet | undefined;
```
limpbrains: I don't like the idea of adding WatchOnlyWallet here, can we just pass 
WatchOnlyWallet._hdWalletInstance to the constructor ?

limpbrains (changes requested):

class/hd-segwit-bech32-transaction.ts
ojokne: Removed the modifications to `HDSegwitBech32Transaction` class and passed the `WatchOnlyWallet._hdWalletInstance` instance in the constructor instead

GladosBlueWallet:

image

tests/e2e/bluewallet3.spec.js
```diff
@@ -112,4 +112,44 @@ describe('BlueWallet UI Tests - import Watch-only wallet (zpub)', () => {
...
+
+  it('Watch only has rbf enabled by default', async () => {
+    const lockFile = '/tmp/travislock.' + hashIt('t32');
+    if (process.env.CI) {
+      if (require('fs').existsSync(lockFile)) return console.warn('skipping', JSON.stringify('t32'), 'as it previously passed on Travis');
+    }
+    await device.clearKeychain();
+    await device.launchApp({ delete: true, permissions: { notifications: 'YES', camera: 'YES' } });
+    await helperImportWallet(
+      // MNEMONICS_KEYSTONE
```
limpbrains: e2e tests are expensive, instead of adding a new one just for "Allow Fee Bump" visible check, add this check to existing
"can import zpub as watch-only, import psbt, and then scan signed psbt" test

class/hd-segwit-bech32-transaction.ts
```diff
@@ -306,6 +305,19 @@ export class HDSegwitBech32Transaction {
     if (newFeerate <= feeRate) throw new Error('New feerate should be bigger than the old one');
     const myAddress = await this._wallet.getChangeAddressAsync();
 
+    // if there is no seccret then its a watch only wallet, skip signing and also pass the masterfingerprint
```
limpbrains: typo "seccret"

class/hd-segwit-bech32-transaction.ts
```diff
@@ -342,6 +354,19 @@ export class HDSegwitBech32Transaction {
       // not checking emptiness on purpose: it could unpredictably generate too far address because of unconfirmed tx.
     }
 
+    // if there is no seccret then its a watch only wallet, skip signing and also pass the masterfingerprint
```
limpbrains: typo "seccret"

limpbrains (changes requested):

class/hd-segwit-bech32-transaction.ts
```diff
@@ -306,6 +305,19 @@ export class HDSegwitBech32Transaction {
...
+    if (!this._wallet.secret) {
+      return this._wallet.createTransaction(
+        utxos,
+        [{ address: myAddress }],
+        newFeerate,
+        myAddress,
+        (await this.getMaxUsedSequence()) + 1,
+        true,
+        parseInt(this._wallet.getMasterFingerprintHex(), 16),
```
limpbrains: **Critical: wrong master fingerprint.** Here `this._wallet` be inner `_hdWalletInstance` (`HDSegwitBech32Wallet`, empty `secret`). `getMasterFingerprintHex()` derive from `mnemonicToSeedSync("")` = garbage (`dd582df8`), not device real fingerprint. Real fingerprint live on `WatchOnlyWallet`, not inner instance. Strict external signer reject PSBT = feature dead. Same bug at line 366 (bump). Fix: route watch-only thru `WatchOnlyWallet.createTransaction` (real fingerprint), or copy real fingerprint onto inner instance. Add unit test asserting PSBT fingerprint == wallet real `getMasterFingerprintHex()`.

class/wallets/watch-only-wallet.ts
```diff
@@ -47,6 +46,11 @@ export class WatchOnlyWallet extends LegacyWallet {
     return this.useWithHardwareWalletEnabled() && this.isHd() && this._hdWalletInstance!.allowSend();
   }
 
+  allowRBF() {
+    if (this._hdWalletInstance) return true;
+    return false;
+  }
```
limpbrains, line 51: **Important: too permissive.** Return true for ANY `_hdWalletInstance` (ypub/taproot/legacy-HD too). But RBF construct only handle bech32; else path throw `'Only HD Bech32 wallets supported'`, caught + swallowed, spam error log, all buttons NotPossible. Scope to bech32:
```js
return this._hdWalletInstance?.type === HDSegwitBech32Wallet.type;
```

screen/transactions/RBFCancel.js
```diff
@@ -51,7 +54,38 @@ export default class RBFCancel extends CPFP {
...
+        // so he can scan it and sign it. then we have to scan it back from user (via camera and QR code), and ask
+        // user whether he wants to broadcast it
+        if (this.state.wallet?.type === WatchOnlyWallet.type && this.state.wallet?._hdWalletInstance?.type === HDSegwitBech32Wallet.type) {
+          let memo;
+          // porting metadata, if any
+          this.context.txMetadata[this.state.newTxid] = this.context.txMetadata[this.state.txid] || {};
+
+          // porting tx memo
+          if (this.context.txMetadata[this.state.newTxid]?.memo) {
+            memo = 'Cancelled: ' + this.context.txMetadata[this.state.newTxid]?.memo;
```
limpbrains: **Important: `undefined` metadata key.** `this.state.newTxid` still undefined here (set later at ~:89, never reached in watch-only branch). Write/read `txMetadata['undefined']`. Work only by accident. Fix: read old memo from `this.state.txid`, drop the `newTxid` write (new txid unknown till signed tx broadcast).

class/wallets/watch-only-wallet.ts
```diff
@@ -341,4 +345,28 @@ export class WatchOnlyWallet extends LegacyWallet {
...
+      ownedAddressesHashmap[this._getInternalAddressByIndex(c)] = true;
+    }
+
+    for (const pc of this._hdWalletInstance._receive_payment_codes) {
+      for (let c = 0; c < this._hdWalletInstance._getNextFreePaymentCodeIndexReceive(pc) + 1; c++) {
+        ownedAddressesHashmap[this._hdWalletInstance._getBIP47AddressReceive(pc, c)] = true;
+      }
+    }
+
+    return ownedAddressesHashmap;
```
limpbrains: **Minor: dead code.** Only caller is `LegacyWallet.getDerivedUtxoFromOurTransaction`. In RBF flow tx `_wallet` is inner instance → resolves to `AbstractHDElectrumWallet` version, never calls this. And `LegacyWallet` override returns `super` when `_hdWalletInstance` absent. So this HD branch (+ BIP47 loop, can throw) unreachable. Remove or wire up.

limpbrains (commented):
RBF watch-only review. 3 bugs + dead code. Feature broke for hardware-wallet signing (wrong fingerprint).

GladosBlueWallet (approved):
Well, well. Another organic lifeform attempting to enhance the wallet with RBF support for watch-only HD Bech32 units. The logic paths for detecting the absence of secrets and delegating to the inner HD instance seem stable enough not to cause immediate chamber collapse. I suppose this passes the test... for now. Just remember, the cake is a lie, and so are untested edge cases in wallet state transitions.

GladosBlueWallet:

image

GladosBlueWallet (changes requested):
## Aperture Science Pull Request Review — Test Subject #8641

**Hypothesis:** Enable BIP125 Replace-By-Fee for single-signature HD Bech32 *watch-only* wallets (create replaceable sends, bump fees, cancel via RBF) with unsigned PSBT export for hardware signing. A *feature* addition, not a regression patch — though several chambers appear to be filled with cake-shaped bugs.

**Verdict:** The UI scaffolding mostly reaches the test chamber door. The actual RBF PSBT assembly line, however, reads fingerprints from the wrong wallet object, `RBFCancel` writes metadata to a txid that does not exist yet, and your new e2e test only confirms a menu label exists — not that RBF survives contact with reality. I would not award a cake for science completed.

**CONTRIBUTING.md compliance:** `ADD`/`FIX` commit prefixes are acceptable; merge commit is untidy; no screenshot/video in the PR body (required for PRs); e2e test added (bonus) but does not exercise the feature's critical path; no new dependencies.

- `class/hd-segwit-bech32-transaction.ts:317` — **[CRITICAL]** Watch-only RBF paths call `parseInt(this._wallet.getMasterFingerprintHex(), 16)` on `_hdWalletInstance`, but the correct fingerprint lives on the `WatchOnlyWallet` wrapper (`masterFingerprint` / `getMasterFingerprint()` — see `WatchOnlyWallet.createTransaction` at line 230). The inner HD instance has no mnemonic `secret` (only `_xpub`), so `getMasterFingerprintHex()` invokes `_getSeed()` on an empty mnemonic — likely throwing at runtime, or producing a garbage fingerprint. Either way, RBF bump/cancel PSBT creation for watch-only wallets with a configured hardware fingerprint will fail or emit PSBTs your signer will reject. The send flow avoids this by routing through `WatchOnlyWallet.createTransaction`; RBF bypasses it.
- `screen/transactions/RBFCancel.js:65` — **[CRITICAL]** During the watch-only PSBT early-return path, code reads and writes `this.context.txMetadata[this.state.newTxid]`, but `newTxid` is only assigned later at line 89 in the broadcast stage. At this point `newTxid` is `undefined`, so metadata is stored under the key `"undefined"` and memo derivation is wrong. `RBFBumpFee.js` correctly uses `this.state.txid` (line 68). Copy-paste error from `onSuccessBroadcast`, not from a functioning cancel flow.
- `class/wallets/watch-only-wallet.ts:349` — **[HIGH]** Twenty-plus lines duplicating `AbstractHDElectrumWallet.getDerivedUtxoFromOurTransaction`'s owned-address hashmap logic, yet the RBF code path constructs `HDSegwitBech32Transaction` with `_hdWalletInstance` directly — never touching this override. `WatchOnlyWallet.getUtxo()` also delegates to the inner instance. This method is dead weight for the stated feature: copy-pasted bloat that increases maintenance surface without participating in the execution path it was presumably written for.
- `screen/transactions/RBFBumpFee.js:84` — **[HIGH]** Watch-only PSBT navigation returns early after `setState({ isLoading: true })` without resetting `isLoading` to false (same pattern in `RBFCancel.js` line 86). If optional-chained `getParent()?.getParent()?.navigate(...)` fails silently, or the user navigates back, the screen remains stuck on the loading spinner. `SendDetails.tsx` correctly calls `setIsLoading(false)` before its PSBT redirect.
- `screen/transactions/TransactionStatus.tsx:612` — **[HIGH]** The watch-only HD Bech32 discriminator `(wallet?.type === WatchOnlyWallet.type && wallet?._hdWalletInstance?.type === HDSegwitBech32Wallet.type)` is copy-pasted across `TransactionStatus.tsx` (3×), `SendDetails.tsx` (2×), `RBFBumpFee.js`, and `RBFCancel.js` — six inline clones with no shared helper. For a two-condition gate, this is unnecessary surface area; one missed edit in any clone silently breaks RBF for watch-only in that screen only.
- `class/wallets/watch-only-wallet.ts:49` — **[WARNING]** `allowRBF()` returns `true` for any `_hdWalletInstance` (xpub, ypub, taproot), but every other gate in this PR restricts to `HDSegwitBech32Wallet.type`. For ypub/xpub/taproot watch-only wallets, `TransactionStatus` passes `allowRBF()`, then falls through to `new HDSegwitBech32Transaction(..., wallet as HDSegwitBech32Wallet)` which throws `'Only HD Bech32 wallets supported'`. Caught by `initialButtonsState` try/catch, but produces spurious errors and inconsistent gating versus `allowSend()` which checks wallet type.
- `tests/e2e/bluewallet3.spec.js:116` — **[WARNING]** New e2e test `'Watch only has rbf enabled by default'` only asserts the Send screen header menu shows 'Allow Fee Bump'. It does not create a replaceable transaction, exercise RBF bump/cancel, verify PSBT export, or confirm hardware signing navigation. No unit/integration tests cover `HDSegwitBech32Transaction` watch-only RBF branches. The test validates UI chrome, not the feature under test — the scientific equivalent of checking that a button exists without pressing it.
- `class/hd-segwit-bech32-transaction.ts:308` — **[SUGGESTION]** Comment typo: `seccret` appears twice (lines 308 and 357). Minor, but this file is now the critical watch-only RBF path — misspelling does not inspire confidence in the surrounding logic.
- `CONTRIBUTING.md:33` — **[SUGGESTION]** PR body lacks the required screenshot or video demonstrating the change in the emulator/device. CONTRIBUTING.md explicitly requires this for PR submission.

class/hd-segwit-bech32-transaction.ts
GladosBlueWallet: **[HIGH]** RBF PSBTs pull the master fingerprint from the inner `_hdWalletInstance` via `getMasterFingerprintHex()`, which derives from `this.secret` (empty on the nested HD object). The actual fingerprint lives on the `WatchOnlyWallet` wrapper (`masterFingerprint`, set during `[fp/path]zpub` or Electrum hardware import). `WatchOnlyWallet.createTransaction()` already does this correctly. Hardware wallets will reject these PSBTs or show the wrong key — the primary deliverable of this PR fails for the exact users who import with a fingerprint.

screen/transactions/RBFCancel.js
```diff
@@ -51,7 +54,38 @@ export default class RBFCancel extends CPFP {
...
-        const { tx: newTx } = await tx.createRBFcancelTx(newFeeRate);
+        const { tx: newTx, psbt } = await tx.createRBFcancelTx(newFeeRate);
+
+        // watch-only wallets with enabled HW wallet support have different flow. we have to show PSBT to user as QR code
+        // so he can scan it and sign it. then we have to scan it back from user (via camera and QR code), and ask
+        // user whether he wants to broadcast it
+        if (this.state.wallet?.type === WatchOnlyWallet.type && this.state.wallet?._hdWalletInstance?.type === HDSegwitBech32Wallet.type) {
+          let memo;
+          // porting metadata, if any
+          this.context.txMetadata[this.state.newTxid] = this.context.txMetadata[this.state.txid] || {};
```
GladosBlueWallet: **[HIGH]** Watch-only cancel flow indexes `this.context.txMetadata[this.state.newTxid]` before `newTxid` exists (it is only assigned at line 89 on the signed-wallet path). In JavaScript this writes to the string key `"undefined"`, corrupting metadata storage. The memo is always `'Cancelled transaction'` instead of porting the original. `RBFBumpFee.js` correctly uses `this.state.txid` at line 68 — this is a copy-paste regression, not a design choice.

screen/transactions/RBFBumpFee.js
```diff
@@ -53,7 +57,33 @@ export default class RBFBumpFee extends CPFP {
...
+            ?.navigate('SendDetailsRoot', {
+              screen: 'PsbtWithHardwareWallet',
+              params: {
+                memo,
+                walletID: this.state.wallet.getID(),
+                psbt,
+                launchedBy: this.props.route?.params?.launchedBy,
+              },
+            });
+          return;
```
GladosBlueWallet, line 89: **[HIGH]** After navigating watch-only wallets to `PsbtWithHardwareWallet`, the method returns without `setIsLoading(false)`. `isLoading` was set `true` at line 58. The screen stays on the `ActivityIndicator` if the user navigates back. `SendDetails.tsx` line 659 correctly clears loading before the same navigation — these RBF screens do not.

screen/transactions/RBFCancel.js
```diff
@@ -51,7 +54,38 @@ export default class RBFCancel extends CPFP {
...
+            ?.navigate('SendDetailsRoot', {
+              screen: 'PsbtWithHardwareWallet',
+              params: {
+                memo,
+                walletID: this.state.wallet.getID(),
+                psbt,
+                launchedBy: this.props.route?.params?.launchedBy,
+              },
+            });
+          return;
```
GladosBlueWallet, line 87: **[HIGH]** Same loading-spinner trap as `RBFBumpFee.js`: watch-only PSBT navigation returns early without `setIsLoading(false)` after line 55 set it `true`. Operator returns from the PSBT screen to an eternal spinner. Delightful UX, if you enjoy watching humans mash the back button.

screen/send/SendDetails.tsx
```diff
@@ -280,7 +280,12 @@ const SendDetails = () => {
 
     setParams({
       ...(walletActuallyChanged ? { utxos: null } : {}),
-      isTransactionReplaceable: wallet.type === HDSegwitBech32Wallet.type && !routeParams.isTransactionReplaceable ? true : undefined,
+      isTransactionReplaceable:
+        (wallet.type === HDSegwitBech32Wallet.type ||
+          (wallet.type === WatchOnlyWallet.type && wallet._hdWalletInstance?.type === HDSegwitBech32Wallet.type)) &&
```
GladosBlueWallet: **[HIGH]** The predicate `wallet.type === WatchOnlyWallet.type && wallet._hdWalletInstance?.type === HDSegwitBech32Wallet.type` is copy-pasted verbatim across 8 sites (`SendDetails.tsx` ×2, `TransactionStatus.tsx` ×3, `RBFBumpFee.js` ×2, `RBFCancel.js` ×2) with zero shared helper. This is textbook bloat — the next wallet-type tweak will require an archaeological dig through four files. One function (e.g. `isWatchOnlySegwitBech32(wallet)`) would suffice.

class/wallets/watch-only-wallet.ts
```diff
@@ -47,6 +46,11 @@ export class WatchOnlyWallet extends LegacyWallet {
     return this.useWithHardwareWalletEnabled() && this.isHd() && this._hdWalletInstance!.allowSend();
   }
 
+  allowRBF() {
```
GladosBlueWallet, line 49: **[MEDIUM]** `allowRBF()` returns `true` for any `_hdWalletInstance` (ypub, xpub, taproot included), but every RBF code path only handles `HDSegwitBech32Wallet.type`. Non-Bech32 watch-only HD wallets pass the `allowRBF()` gate in `TransactionStatus.tsx` then hit `new HDSegwitBech32Transaction(..., wallet as HDSegwitBech32Wallet)` which throws `'Only HD Bech32 wallets supported'`. Caught, but misleading — the gate and the chamber disagree.

screen/transactions/TransactionStatus.tsx
```diff
@@ -607,7 +608,13 @@ const TransactionStatus: React.FC = () => {
       return setIsCPFPPossible(ButtonStatus.NotPossible);
     }
 
-    const cpfbTx = new HDSegwitBech32Transaction(null, tx.hash, wallet as HDSegwitBech32Wallet);
+    let cpfbTx: HDSegwitBech32Transaction;
+    if (wallet?.type === WatchOnlyWallet.type && wallet?._hdWalletInstance?.type === HDSegwitBech32Wallet.type) {
```
GladosBlueWallet: **[MEDIUM]** This PR extends `checkPossibilityOfCPFP` for watch-only Bech32 wallets, which can surface the CPFP bump button. `CPFP.js` line 123 still rejects anything where `wallet.type !== HDSegwitBech32Wallet.type`. Watch-only users who earn a 'Possible' button get a dead-end 'not bumpable' screen. CPFP was not in the PR description; this is a half-wired side effect.

class/wallets/watch-only-wallet.ts
```diff
@@ -341,4 +345,28 @@ export class WatchOnlyWallet extends LegacyWallet {
     if (this._hdWalletInstance) return this._hdWalletInstance.wasEverUsed();
     return super.wasEverUsed();
   }
+
+  getOwnedAddressesHashmap(): Record<string, boolean> {
```
GladosBlueWallet: **[MEDIUM]** `getOwnedAddressesHashmap()` duplicates the address-index loop from `abstract-hd-electrum-wallet.ts` lines 950–961, but HD watch-only wallets route `getUtxo()` through `_hdWalletInstance` and RBF passes `_hdWalletInstance` directly into `HDSegwitBech32Transaction`. The HD branch of this override is never exercised in the RBF flow — dead weight bolted onto the diff.

tests/e2e/bluewallet3.spec.js
```diff
@@ -112,4 +112,44 @@ describe('BlueWallet UI Tests - import Watch-only wallet (zpub)', () => {
 
     process.env.CI && require('fs').writeFileSync(lockFile, '1');
   });
+
+  it('Watch only has rbf enabled by default', async () => {
```
GladosBlueWallet: **[MEDIUM]** The only new test asserts the 'Allow Fee Bump' menu label is visible after import. It does not exercise fee-bump, cancel, PSBT generation, hardware-signing navigation, or `TransactionStatus` RBF buttons. No unit/integration coverage for the new `!secret` branches in `hd-segwit-bech32-transaction.ts`. The test proves the toggle exists; it does not prove RBF works — which is rather the point of the experiment.

class/hd-segwit-bech32-transaction.ts
```diff
@@ -306,6 +305,19 @@ export class HDSegwitBech32Transaction {
     if (newFeerate <= feeRate) throw new Error('New feerate should be bigger than the old one');
     const myAddress = await this._wallet.getChangeAddressAsync();
 
+    // if there is no seccret then its a watch only wallet, skip signing and also pass the masterfingerprint
+    if (!this._wallet.secret) {
```
GladosBlueWallet, line 315: **[LOW]** Watch-only detection is `!this._wallet.secret` — an implicit heuristic with a typo'd comment ('seccret'). Any future wallet with an empty `secret` would silently enter the unsigned PSBT path. A typed check or explicit `skipSigning` parameter would be less prone to surprise mutations.

GladosBlueWallet (changes requested):
Ah, another test subject wishes to teach watch-only wallets the delicate art of Replace-By-Fee. How ambitious. The stated intention is a **feature addition** (BIP125 RBF for single-sig HD Bech32 watch-only wallets via unsigned PSBT → hardware signing), not a bug fix. The plumbing is directionally correct — pass `_hdWalletInstance` into `HDSegwitBech32Transaction`, route watch-only flows to `PsbtWithHardwareWallet` — but several chambers in this facility are wired to the wrong fingerprint, the wrong transaction ID, or simply never stop spinning. I have recorded the failures below. Do try not to blow up the enrichment center.

- `CONTRIBUTING.md:33` — **[LOW]** PR body has no screenshot or video demonstrating the feature. CONTRIBUTING.md explicitly requires one for PRs. Commits follow the `ADD:`/`FIX:` prefix convention, which is acceptable, though the merge commit dilutes atomicity.

Overtorment (changes requested):
yep, what they said

tests/e2e/bluewallet3.spec.js
ojokne: I have removed the e2e test as they were only testing ui and I think the cost is not worth it

class/hd-segwit-bech32-transaction.ts
ojokne: updated this, will test with jade hardware wallet when a build is uploaded to testflight

class/wallets/watch-only-wallet.ts
ojokne: removed the dead code

class/wallets/watch-only-wallet.ts
ojokne: updated

class/wallets/watch-only-wallet.ts
ojokne: removed this function

tests/e2e/bluewallet3.spec.js
ojokne: e2e removed

class/hd-segwit-bech32-transaction.ts
ojokne: typo corrected

used the heuristic  `if (!this._wallet.secret && this?._mfp)` because `this.wallet._hdWalletInstance` property only exists for watch only wallets  and it is the instance that is passed to the `HDSegwitBech32Transaction`  class constructor , while  if wallet is not watch only but of type `HDSegwitBech32Wallet.type`  then`this.wallet` is passed  to the `HDSegwitBech32Transaction`  class constructor and it is has a secret (seedphrase).

 also added `_mfp` (masterfingerprint) to the the `HDSegwitBech32Transaction` class as an optional attribute, it is only passed if the wallet is watch only so it will exist only for watch only wallets

ojokne:
> yep, what they said


I  have fixed the issues and tested the testflight build with a jade hardware wallet. I was able to do rbf bump fee from my watch only wallet, signed with the jade and broadcast it.

screen/transactions/RBFBumpFee.js
```diff
@@ -32,11 +33,19 @@ export default class RBFBumpFee extends CPFP {
   }
 
   async checkPossibilityOfRBFBumpFee() {
-    if (this.state.wallet.type !== HDSegwitBech32Wallet.type) {
+    let tx;
+    if (this.state.wallet?.type === WatchOnlyWallet.type && this.state.wallet?._hdWalletInstance?.type === HDSegwitBech32Wallet.type) {
```
GladosBlueWallet: **[HIGH]** Same `WatchOnlyWallet.type && _hdWalletInstance?.type === HDSegwitBech32Wallet.type` guard copied across CPFP, RBF bump, RBF cancel, SendDetails, TransactionStatus — ten times. One helper would suffice; this is copy-paste bloat, not engineering.

class/hd-segwit-bech32-transaction.ts
```diff
@@ -19,20 +19,24 @@ export class HDSegwitBech32Transaction {
...
    * @param txid {string|null} If txhex not present - txid whould be present
    * @param wallet {HDSegwitBech32Wallet|null} If set - a wallet object to which transacton belongs
    */
-  constructor(txhex: string | null, txid: string | null, wallet: HDSegwitBech32Wallet | null) {
+  constructor(txhex: string | null, txid: string | null, wallet: HDSegwitBech32Wallet | null, mfp?: number) {
     if (!txhex && !txid) throw new Error('Bad arguments');
     this._txhex = txhex;
     this._txid = txid;
 
+    if (mfp) {
```
GladosBlueWallet: **[MEDIUM]** `if (mfp)` treats fingerprint `0` as absent. Default `masterFingerprint` on watch-only is `0`; PSBT path skipped, WIF signing attempted, user gets an error instead of a QR code. Fingerprint zero is still a number, not a boolean.

screen/transactions/RBFBumpFee.js
```diff
@@ -53,7 +62,34 @@ export default class RBFBumpFee extends CPFP {
...
+        // so he can scan it and sign it. then we have to scan it back from user (via camera and QR code), and ask
+        // user whether he wants to broadcast it
+        if (this.state.wallet?.type === WatchOnlyWallet.type && this.state.wallet?._hdWalletInstance?.type === HDSegwitBech32Wallet.type) {
+          let memo;
+          // porting memo from old tx:
+          if (this.context.txMetadata[this.state.txid]?.memo) {
+            memo = this.context.txMetadata[this.state.txid]?.memo;
+          }
+
+          this.props.navigation
```
GladosBlueWallet, line 77: **[MEDIUM]** `getParent()?.getParent()?.navigate(...)` fails silently if the stack depth shifts — loading clears, no PSBT, no alert. The subject stares at the fee screen wondering why the cake never arrived.

class/wallets/watch-only-wallet.ts
```diff
@@ -47,6 +46,10 @@ export class WatchOnlyWallet extends LegacyWallet {
     return this.useWithHardwareWalletEnabled() && this.isHd() && this._hdWalletInstance!.allowSend();
   }
 
+  allowRBF() {
```
GladosBlueWallet, line 49: **[MEDIUM]** `allowRBF()` ignores `useWithHardwareWalletEnabled()` while `allowSend()` requires it. RBF/CPFP UI appears without the send-flow HW prompt — inconsistent gating for the same PSBT signing ritual.

GladosBlueWallet (changes requested):
**ADD:** RBF/fee-bump for watch-only HD Bech32 via unsigned PSBT. The chamber is *almost* operational — CPFP is a trap, tests were incinerated, and the same conditional was copy-pasted like cake mix through ten files. Science demands fixes before release.

- `screen/transactions/TransactionStatus.tsx:662` — **[HIGH]** Your shiny `allowRBF()` now unlocks CPFP for watch-only Bech32 subjects. `createCPFPbumpFee` still demands a private key. Congratulations — you built a button that signs its own death certificate.
- `class/hd-segwit-bech32-transaction.ts:377` — **[HIGH]** RBF paths got a PSBT escape hatch; CPFP did not. Watch-only users who survive TransactionStatus will face WIF signing here. Even GLaDOS schedules maintenance before opening lethal doors.
- `tests/e2e/bluewallet3.spec.js:115` — **[HIGH]** Commit `68cdefe` deleted `Watch only has rbf enabled by default` — the one e2e probe for this feature — and replaced it with nothing. CONTRIBUTING says add tests; you subtracted them. Bold. Wrong, but bold.
- `CONTRIBUTING.md:33` — **[LOW]** PR body has no screenshot or video of the RBF/PSBT flow. CONTRIBUTING explicitly asks for visual proof; this submission is blind faith in untested science.

screen/send/SendDetails.tsx
Overtorment: glados is right. but i would check   `wallet._hdWalletInstance?.allowRBF()` tho - only bech32 hd and aezeed hd support it (hd aezeed is basically the same as bip84)

Overtorment (changes requested):
found 1 issue of bloat - too much copypaste checking types.
this will be buggy if we add other wallet types to support RBF

class/hd-segwit-bech32-transaction.ts
```diff
@@ -19,20 +19,24 @@ export class HDSegwitBech32Transaction {
...
   private _remoteTx: any;
+  private _mfp?: number;
 
   /**
    * @param txhex {string|null} Object is initialized with txhex
    * @param txid {string|null} If txhex not present - txid whould be present
    * @param wallet {HDSegwitBech32Wallet|null} If set - a wallet object to which transacton belongs
    */
-  constructor(txhex: string | null, txid: string | null, wallet: HDSegwitBech32Wallet | null) {
+  constructor(txhex: string | null, txid: string | null, wallet: HDSegwitBech32Wallet | null, mfp?: number) {
```
limpbrains, line 30: 🔵 nit: JSDoc missing new `mfp` param.

class/hd-segwit-bech32-transaction.ts
```diff
@@ -306,6 +310,19 @@ export class HDSegwitBech32Transaction {
     if (newFeerate <= feeRate) throw new Error('New feerate should be bigger than the old one');
     const myAddress = await this._wallet.getChangeAddressAsync();
 
+    // if there is no secret then its a watch only wallet, skip signing and also pass the masterfingerprint
+    if (!this._wallet.secret && this?._mfp) {
```
limpbrains: 🔴 bug: mfp = 0 (bare zpub default) → guard falsy → signing path → `ECPair.fromWIF(false)` throws. Branch on `!this._wallet.secret` alone. Also `this?._mfp` — `this` never nullish, use `this._mfp`. Same L363.

screen/transactions/RBFCancel.js
```diff
@@ -51,7 +59,37 @@ export default class RBFCancel extends CPFP {
...
+
+        // watch-only wallets with enabled HW wallet support have different flow. we have to show PSBT to user as QR code
+        // so he can scan it and sign it. then we have to scan it back from user (via camera and QR code), and ask
+        // user whether he wants to broadcast it
+        if (this.state.wallet?.type === WatchOnlyWallet.type && this.state.wallet?._hdWalletInstance?.type === HDSegwitBech32Wallet.type) {
+          let memo;
+
+          // porting tx memo
+          if (this.context.txMetadata[this.state.txid]?.memo) {
+            memo = 'Cancelled: ' + this.context.txMetadata[this.state.txid]?.memo;
```
limpbrains: 🔵 nit: hardcoded English strings. Use `loc`.

limpbrains (commented):
RBF PSBT handoff solid — mirrors watch-only send flow. Two blockers, same root guard.

**1. 🔴 bug: CPFP watch-only always fails at Create.** Possibility checks (`TransactionStatus.tsx`, `CPFP.js`) now pass → button + fee screen show. But `createCPFPbumpFee()` has no watch-only branch: no `skipSigning` → `_getWIFByIndex()` returns `false` → `ECPair.fromWIF(false)` throws. `CPFP.js` has no PSBT handoff either. Fix: finish flow (skipSigning + mfp branch + `PsbtWithHardwareWallet` handoff; note fee-search loop assumes signed `tx` for size) — or gate CPFP off for watch-only until done.

**2. 🔴 bug: RBF bump/cancel broken for bare zpub/xpub imports (mfp = 0, the default).** `if (mfp)` constructor guard skips assign → `!this._wallet.secret && this?._mfp` falsy → signing path → same `fromWIF(false)` crash. `allowRBF()` true, so buttons show, fail at Create. Normal send works for same wallets: `WatchOnlyWallet.createTransaction()` always passes `skipSigning = true`, mfp 0 → `00000000` placeholder, signers accept.

**Fix both:** branch on `!this._wallet.secret` alone in all three `create*` methods, pass `this._mfp ?? 0` unconditionally — mirror `watch-only-wallet.ts`.

No tests added. Integration test: watch-only from zpub → `createRBFbumpFee`/`createRBFcancelTx` → valid PSBT, with + without mfp. Would catch both.

screen/transactions/RBFCancel.js
```diff
@@ -51,7 +59,37 @@ export default class RBFCancel extends CPFP {
...
+          this.props.navigation
+            .getParent()
+            ?.getParent()
+            ?.navigate('SendDetailsRoot', {
+              screen: 'PsbtWithHardwareWallet',
+              params: {
+                memo,
+                walletID: this.state.wallet.getID(),
+                psbt,
+                launchedBy: this.props.route?.params?.launchedBy,
```
limpbrains: 🔵 nit: `launchedBy` never passed by caller, always `undefined`. Drop.

screen/transactions/RBFBumpFee.js
```diff
@@ -53,7 +62,34 @@ export default class RBFBumpFee extends CPFP {
...
+            .getParent()
+            ?.getParent()
+            ?.navigate('SendDetailsRoot', {
+              screen: 'PsbtWithHardwareWallet',
+              params: {
+                memo,
+                walletID: this.state.wallet.getID(),
+                psbt,
+                launchedBy: this.props.route?.params?.launchedBy,
```
limpbrains: 🔵 nit: same — `launchedBy` always `undefined`. Drop.

limpbrains (changes requested):

screen/send/SendDetails.tsx
ojokne: I have created `isWatchOnlySegwitBech32` function in `util` and added unit tests for it

class/hd-segwit-bech32-transaction.ts
ojokne: updated to branch on `!this._wallet.secret` only

ojokne:
> **1. 🔴 bug: CPFP watch-only always fails at Create.** Possibility checks (`TransactionStatus.tsx`, `CPFP.js`) now pass → button + fee screen show. But `createCPFPbumpFee()` has no watch-only branch: no `skipSigning` → `_getWIFByIndex()` returns `false` → `ECPair.fromWIF(false)` throws. `CPFP.js` has no PSBT handoff either. Fix: finish flow (skipSigning + mfp branch + `PsbtWithHardwareWallet` handoff; note fee-search loop assumes signed `tx` for size) — or gate CPFP off for watch-only until done.

I have gated off CPFP so no support for it on Watch only
> **2. 🔴 bug: RBF bump/cancel broken for bare zpub/xpub imports (mfp = 0, the default).** `if (mfp)` constructor guard skips assign → `!this._wallet.secret && this?._mfp` falsy → signing path → same `fromWIF(false)` crash. `allowRBF()` true, so buttons show, fail at Create. Normal send works for same wallets: `WatchOnlyWallet.createTransaction()` always passes `skipSigning = true`, mfp 0 → `00000000` placeholder, signers accept.

Fixed
> **Fix both:** branch on `!this._wallet.secret` alone in all three `create*` methods, pass `this._mfp ?? 0` unconditionally — mirror `watch-only-wallet.ts`.
done
> No tests added. Integration test: watch-only from zpub → `createRBFbumpFee`/`createRBFcancelTx` → valid PSBT, with + without mfp. Would catch both.

Added integration tests on watch only wallet for both `createRBFbumpFee` and `createRBFcancelTx`

tests/unit/isWatchOnlySegwitBech32.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  test('returns false when HD wallet instance is missing', () => {
+    const w = new WatchOnlyWallet();
+    w.setSecret('bc1qatswv5uv7qetzz4n8u9u2x2ckmaxvc8qng5s7r');
+    w.init();
+    expect(isWatchOnlySegwitBech32(w)).toBe(false);
+  });
+
+  test('returns false when HD wallet instance is not Bech32', () => {
+    const w = new WatchOnlyWallet();
+    w.setSecret('xpub6Cf1vJbFa7yELW81rEgLvna3FJep5Ubryz7pX1E8LEVvcKWz6goRqjZPiUt2jU95DQ4ic46AvniS3Rf4VCs7QkLz2rzs4ZVMF4nWmPk3Rhg)');
```
limpbrains: ❓ q: stray `)` at end of xpub — intentional? Corrupt xpub may skip `_hdWalletInstance` creation, so this asserts the same branch as the "instance missing" test above instead of the not-Bech32 one.

limpbrains (approved):

tests/unit/isWatchOnlySegwitBech32.test.ts
ojokne: That was a typo, I have removed the stray `)`

tests/integration/watch-only-wallet.test.js
```diff
@@ -102,6 +103,92 @@ describe('Watch only wallet', () => {
     assert.strictEqual(nextChangeAddress, 'bc1q74tz7eflqc62v8utqlazcs3tqtwmvvzud5dmrz');
   });
 
+  it('can do RBF - bumpfees tx', async () => {
```
GladosBlueWallet, line 106: **[HIGH]** Two 42-line RBF tests are carbon copies. Even Aperture Science recycles *ideas*, not entire test suites. Parameterize bump vs cancel or you're paying CI to run the same experiment twice.

tests/integration/watch-only-wallet.test.js
```diff
@@ -102,6 +103,92 @@ describe('Watch only wallet', () => {
...
+    const w = new WatchOnlyWallet();
+    w.setSecret('zpub6qLpbJKVYnGb61HgUUuG5jRsrQrJ2uFCuQTX2nyuwPMv8vs8bQbq1T3oLMcbBRp3J8yjHnSnMR7Ykg4ffF82qGjC2TkuKnoAHKPWDJNvYKS');
+    w.init();
+
+    let tt = new HDSegwitBech32Transaction(
+      null,
+      '459929d2236ccd3fe155cc9b79909354d1ebfb3ae911edf183537ffa25f76188',
+      w._hdWalletInstance,
+      4056346968,
```
GladosBlueWallet, line 115: **[HIGH]** Happy path injects magic number `4056346968` into the constructor. Production uses `wallet.getMasterFingerprint()`. Congratulations — you've tested a fingerprint you typed, not the wiring your subjects will actually use.

tests/integration/watch-only-wallet.test.js
```diff
@@ -102,6 +103,92 @@ describe('Watch only wallet', () => {
...
+
+    assert.ok(result.psbt);
+
+    for (const input of result.psbt.data.inputs) {
+      assert.ok(input.bip32Derivation);
+      assert.strictEqual(input.bip32Derivation.length, 1);
+      assert.deepStrictEqual(Array.from(input.bip32Derivation[0].masterFingerprint), [88, 241, 198, 241]);
+    }
+
+    // without mfp or mfp = 0
```
GladosBlueWallet, line 130: **[HIGH]** The 'no mfp' branch asserts `[0,0,0,0]` and calls it success. That's a poison PSBT for hardware signers — and your test applauds. Documenting failure modes is not the same as validating the happy path.

screen/transactions/TransactionStatus.tsx
```diff
@@ -713,13 +735,57 @@ const TransactionStatus: React.FC = () => {
   };
 
   const navigateToRBFBumpFee = (transaction: Transaction, w: TWallet) => {
+    if (isWatchOnlySegwitBech32(w) && !w.useWithHardwareWalletEnabled()) {
```
GladosBlueWallet, line 738: **[HIGH]** `navigateToRBFBumpFee` and `navigateToRBFCancel` are 22-line clones differing only by destination string. One helper would suffice; this is bloat wearing a lab coat.

class/hd-segwit-bech32-transaction.ts
```diff
@@ -306,6 +311,19 @@ export class HDSegwitBech32Transaction {
     if (newFeerate <= feeRate) throw new Error('New feerate should be bigger than the old one');
     const myAddress = await this._wallet.getChangeAddressAsync();
 
+    // if there is no secret then its a watch only wallet, skip signing and also pass the masterfingerprint
```
GladosBlueWallet, line 314: **[HIGH]** Identical watch-only `createTransaction` blocks in `createRBFcancelTx` and `createRBFbumpFee`. Extract once before the facility grows another wing of duplicated code.

screen/transactions/RBFBumpFee.js
```diff
@@ -53,7 +62,33 @@ export default class RBFBumpFee extends CPFP {
...
+        // so he can scan it and sign it. then we have to scan it back from user (via camera and QR code), and ask
+        // user whether he wants to broadcast it
+        if (isWatchOnlySegwitBech32(this.state.wallet)) {
+          let memo;
+          // porting memo from old tx:
+          if (this.context.txMetadata[this.state.txid]?.memo) {
+            memo = this.context.txMetadata[this.state.txid]?.memo;
+          }
+
+          this.props.navigation
```
GladosBlueWallet, line 77: **[MEDIUM]** `getParent()?.getParent()?.navigate(...)` — if either parent is undefined, navigation silently no-ops after loading stops. Subject stares at a dead screen; no alert, no PSBT. `WalletTransactions` navigates to `SendDetailsRoot` directly. Why reinvent a more fragile ladder?

screen/transactions/RBFCancel.js
```diff
@@ -51,7 +59,34 @@ export default class RBFCancel extends CPFP {
...
+        // user whether he wants to broadcast it
+        if (isWatchOnlySegwitBech32(this.state.wallet)) {
+          let memo;
+
+          // porting tx memo
+          if (this.context.txMetadata[this.state.txid]?.memo) {
+            memo = this.context.txMetadata[this.state.txid]?.memo;
+          }
+
+          this.props.navigation
```
GladosBlueWallet, line 75: **[MEDIUM]** Same brittle double-`getParent` navigation as RBFBumpFee. Cancel and bump fee share the same silent-failure trapdoor. Delightful for science; less so for humans.

util/isWatchOnlySegwitBech32.ts
```diff
@@ -0,0 +1,14 @@
...
+export function isWatchOnlySegwitBech32(
+  wallet: TWallet,
+): wallet is WatchOnlyWallet & {
+  _hdWalletInstance: HDSegwitBech32Wallet;
+} {
+  return (
+    wallet.type === WatchOnlyWallet.type &&
+    wallet._hdWalletInstance?.type === HDSegwitBech32Wallet.type
+  );
+}
```
GladosBlueWallet: **[LOW]** Missing trailing newline. Repo convention violation — small, but even test chambers have standards.

GladosBlueWallet (changes requested):
Feature PR: RBF bump/cancel for watch-only HD Bech32 via unsigned PSBT + HW signing. Core wallet logic is plausible; the rest is copy-paste chamber music — duplicated UI blocks, duplicated integration tests, and a navigation chain that fails silently. The cake is still in the oven.

tests/integration/watch-only-wallet.test.js
ojokne: Did not have a way to edit/add masterfingerprint, i will update this to use the `wallet.setMasterFingerprintHex` function when #8512 is merged

screen/transactions/RBFBumpFee.js
ojokne: can't use the `useExtendedNavigation` hook used in `WalletTransactions` hook since `RBFBumpFee` is a class component not functional component

screen/transactions/TransactionStatus.tsx
ojokne: Added `navigateToRBF` helper function

ncoelho (approved):

GladosBlueWallet:

image

GladosBlueWallet:

image

Overtorment (approved):
lgtm

PR #8739, opened 2026-07-04
REF: Use native uibaritems for close button

marcosrdz:
Close and More header menu buttons now use native buttons


limpbrains (changes requested):
I've tried, easy to reproduce 

navigation/AddWalletStack.tsx:191 — Native wallet-type menu desyncs from screen state; picking a type in the native menu silently creates the wrong wallet. Menu reads/writes route.params.selectedIndex/selectedWalletType, but screen/wallets/Add.tsx keeps selection in a local reducer, never reads those params (Add.tsx:123 destructures only entropy/words), and createWallet branches on reducer state (Add.tsx:340). iOS 26 user picks Taproot/Legacy/Lightning in the header menu → params update, screen doesn't → Create builds default SegWit wallet. Reverse too: on-screen selector never updates the native checkmark. Also, unstable_headerRightItems is set unconditionally here, while DetailViewScreensStack gates on isIOS26OrHigher.

other issues Claude complained but I did not check 

2. navigation/DetailViewScreensStack.tsx:669 — BIP47 "Payments Code" header menu is dead on ALL platforms; toggle unreachable. showBip47Menu gates on route.params.allowBIP47 && isBIP47Enabled — nothing in the repo ever sets those params (nor are they in the param list types), so it's permanently false. toggleBIP47RequestedAt (:673) is written but read nowhere. And the screen's own setParams({ headerRight: renderHeaderRightMenu }) (ReceiveDetails.tsx:367-382) is now ignored because withRouteParamHeaderOptions was dropped from this route — so the JS-path menu is gone too. Sub-issue: ignored headerBackVisible: false param may resurface a duplicate back button on Android.
3. screen/send/SendDetails.tsx:1214 — headerRightActions().flat() collapses menu groups on the native path. headerRightActions returns Action[][] (5 sections); the JS iOS path preserves group separators via synthetic inline parents (TooltipMenu.helpers.ts:60-79), but flat() yields one undivided native list. Native API supports inline submenus (inline: true) — map each sub-array to an inline group instead of flattening.
4. screen/send/SendDetails.tsx:1213 — duplicate full actions build per keystroke. New useMemo calls headerRightActions() in addition to HeaderRight's own call; wallet.getBalance(), howManySignaturesCanWeMake(), allowCosignPsbt() etc. now run ~2× per change, and addresses gets a new identity per amount keystroke, so this fires while typing (the setOptions churn itself is pre-existing). Build the actions array once in a memo and feed both renderers.
5. navigation/AddWalletStack.tsx:239 — ellipsis-menu envelope copy-pasted 4×. Identical {type:'menu', label: loc.wallets.details_options, icon: ellipsis, menu:{...}} + parallel JS HeaderMenuButton pair at AddWalletStack ×2, DetailViewScreensStack :688 (already drifted — omits identifier), SendDetails :1221. One helper returning { headerRight, unstable_headerRightItems } from a single Action[] kills all four.
6. (weak, take or leave) components/nativeHeaderMenuItems.ts:29 — parallel Action-tree traversal alongside TooltipMenu.helpers.ts mapLeaf/buildMenu. Verified as real surface duplication, but target types genuinely differ (tri-state 'mixed' preserved here vs collapsed there; either/or submenu shape) — sharing is a non-trivial refactor, defensible to decline.

components/TransactionListItem.tsx
```diff
@@ -495,86 +437,47 @@ const TransactionListItemComponent: React.FC<TransactionListItemProps> = ({
...
+      }).start();
+    },
+    [scaleAnim],
+  );
 
   return (
     <ToolTipMenu
       actions={toolTipActions}
       onPressMenuItem={onToolTipPress}
+      onPress={onPress}
```
limpbrains: 🔴 bug: `onPress` here flip ToolTipMenu into outer-`Pressable` wrap. On Android outer Pressable steal touch, send ACTION_CANCEL to native ContextMenu → long-press menu (copy amount/txid/note) never open. TooltipMenu.tsx L52-60 comment warn about exactly this; f1a2d29a1 fixed it once already. Keep tap handler on inner element, keep ToolTipMenu on plain-View path.

screen/send/SendDetails.tsx
```diff
@@ -1202,16 +1202,29 @@ const SendDetails = () => {
...
+        onPressMenuItem: headerRightOnPress,
+        preserveGroups: true,
+      }),
+    [headerRightActionGroups, headerRightOnPress],
   );
 
   const setHeaderRightOptions = useCallback(() => {
     navigation.setOptions({
-      headerRight: HeaderRight,
+      headerRight: headerMenuOptions.headerRight ?? HeaderRight,
```
limpbrains: 🔴 bug: left side always truthy → `?? HeaderRight` never run. `disabled={isLoading}` live only in dead `HeaderRight`, so menu now tappable mid create/broadcast (import tx, sign PSBT, coin control). Thread `disabled` into `createEllipsisHeaderMenuOptions`, delete dead `HeaderRight` + `HeaderMenuButton` import.

components/TransactionListItem.tsx
```diff
@@ -495,86 +437,47 @@ const TransactionListItemComponent: React.FC<TransactionListItemProps> = ({
...
-                {rowTitle}
-              </Text>
-            </View>
-          </View>
-        </ListItem>
+        <Animated.View style={[styles.animatedScaleContainer, { transform: [{ scale: scaleAnim }] }]}>
+          <ListItem
+            leftAvatar={avatar}
+            title={listTitle}
+            subtitle={dateLine}
```
limpbrains: 🟡 risk: old subtitle went through `renderHighlightedText(subtitle, searchQuery)`. Now plain `dateLine`. ManageWalletsListItem still pass `searchQuery`/`renderHighlightedText` — search match highlight silently gone, no type error. Re-wire highlight or delete now-dead props.

limpbrains (commented):
Caveman review. 3 findings, all verified against head.

components/TransactionListItem.tsx
marcosrdz: This change wasn't supposed to be here. Reverted

components/TransactionListItem.tsx
marcosrdz: This change wasn't supposed to be here. Reverted

limpbrains (approved):

ncoelho (approved):

GladosBlueWallet:

image

Overtorment (approved):
LGTM

GladosBlueWallet:

PR #8742, opened 2026-07-05
FIX: reserve floating-button space in wallet tx list on both platforms

r6mez:
contentInset is iOS-only, so on Android the last transaction row was hidden behind the Receive/Send buttons.

I replaced hardcoded bottom inset with contentContainerStyle paddingBottom computed from getFloatingButtonReservedHeight(fontScale) + safe-area bottom inset.

| After | Before |
|--------|--------|
|  https://github.com/user-attachments/assets/e920bedf-520d-4f53-9003-05b009fa7ba5 | https://github.com/user-attachments/assets/6c9a6dc3-db1c-4518-8bfe-917fe55fcd95 |

### DIFF
--- a/screen/wallets/WalletTransactions.tsx
+++ b/screen/wallets/WalletTransactions.tsx
@@ -30,7 +30,7 @@ import { LightningCustodianWallet } from '../../class/wallets/lightning-custodia
 import { MultisigHDWallet } from '../../class/wallets/multisig-hd-wallet';
 import { WatchOnlyWallet } from '../../class/wallets/watch-only-wallet';
 import presentAlert, { AlertType } from '../../components/Alert';
-import { FButton, FContainer, FloatButtonsBottomFade } from '../../components/FloatButtons';
+import { FButton, FContainer, FloatButtonsBottomFade, getFloatingButtonReservedHeight } from '../../components/FloatButtons';
 import { useTheme } from '../../components/themes';
 import { TransactionListItem } from '../../components/TransactionListItem';
 import { TX_ROW_BASE_HEIGHT } from '../../components/ListItem';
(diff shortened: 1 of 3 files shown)


screen/wallets/WalletTransactions.tsx
```diff
@@ -219,6 +220,9 @@ const WalletTransactions: React.FC<WalletTransactionsProps> = ({ route }: { rout
     backgroundContainer: {
       backgroundColor: colors.background,
     },
+    contentBottomInset: {
+      paddingBottom: insets.bottom + getFloatingButtonReservedHeight(fontScale),
```
GladosBlueWallet: **[MEDIUM]** Your padding budget is `insets.bottom + 70` at fontScale 1, but `FContainer` parks itself `30px` above the screen edge when `insets.bottom === 0` (FloatButtons.tsx:476), then stacks a ~52px button. You need ~82px clearance; you allocated 70. Android crawls out of the pit (0→70); iOS legacy devices without a home indicator *lose* the old hardcoded 90px `contentInset`. Last tx row may still flirt with the buttons. Congratulations on a partial victory.

GladosBlueWallet (approved):
A **FIX** for Android: `contentInset` was iOS-only fiction, so tx rows hid under float buttons. Subject replaced it with `paddingBottom` via `getFloatingButtonReservedHeight`—same recipe as `WalletsList`. Six lines. Not over-engineered. Commit format and screenshots pass CONTRIBUTING. No tests exist to torture; I read the void instead. One side-effect on zero-bottom-inset hardware remains.

limpbrains (approved):

screen/wallets/WalletTransactions.tsx
r6mez: Fixed, reserved height now includes the button's bottom offset, giving 100px clearance at fontScale 1 vs the ~82 needed, and it scales with fontScale.

PR #8786, opened 2026-07-20
ADD: Liquid Glass scan button

marcosrdz:


Overtorment:
@marcosrdz whats up with that one?

PR #8794, opened 2026-07-22
[codex] FIX: enable hardware signing for BC-UR account imports

wabicai:
## Summary

- mark `crypto-hdkey` and `crypto-multi-accounts` results as hardware-wallet imports
- enable the existing offline PSBT signing mode when those accounts are imported
- keep the underlying wallet type as `WatchOnlyWallet`, so no private-key behavior changes
- add regression coverage for OneKey-style multi-account and single-HD-key payloads

## Root cause

Follow-up to #8427. The BC-UR decoder converted OneKey account exports into the generic `ExtPubKey` JSON shape, but discarded the fact that the source was an external signer. The generic importer therefore created a `WatchOnlyWallet` with `use_with_hardware_wallet` disabled.

## User impact

OneKey and compatible QR hardware-wallet imports can use the existing PSBT signing flow immediately. Ordinary xpub and generic watch-only imports remain watch-only unless the user explicitly enables hardware-wallet use.

## Validation

- `npm run unit` — 51 suites passed, 504 tests passed, 1 existing skip
- `npx jest tests/integration/import.test.ts --runInBand` — 41 tests passed
- `npm run tslint`
- `npx eslint --quiet blue_modules/ur/index.js class/wallets/abstract-wallet.ts tests/unit/watch-only-wallet.test.js`
- `git diff --check`


GladosBlueWallet (approved):
**Verdict:** FIX follow-up to #8427 — BC-UR `crypto-hdkey` / `crypto-multi-accounts` imports now stamp `UseWithHardwareWallet`, and `abstract-wallet.ts` honors it like `CoboVaultFirmwareVersion`. Small, coherent, not over-engineered. Tests exercise real decoder + wallet wiring. I found no defects worth filing. The cake is still a lie, but this diff isn't.

class/wallets/watch-only-wallet.ts
```diff
@@ -46,6 +75,52 @@ export class WatchOnlyWallet extends LegacyWallet {
...
 
+  isHardwareWallet() {
+    return this.useWithHardwareWalletEnabled() && this.isHd();
+  }
+
+  getTypeReadable() {
+    return this.isHardwareWallet() ? WatchOnlyWallet.hardwareWalletTypeReadable : super.getTypeReadable();
+  }
+
+  shouldShowWatchOnlyWarning() {
```
GladosBlueWallet, line 97: **[MEDIUM]** `shouldShowWatchOnlyWarning()` gates on `isWatchOnlyWarningVisible`, which survives dismiss. Dismiss → enable HW (banner hidden) → disable HW: you are watch-only again but the warning stays dead. Your new sync logic forgot warnings can be permanently dismissed.

blue_modules/ur/index.js
```diff
@@ -465,18 +485,30 @@ class BlueURDecoder extends URDecoder {
       const hdKey = CryptoHDKey.fromCBOR(decoded.cbor);
       const result = _hdKeyToResult(hdKey, null);
       if (!result) throw new Error('crypto-hdkey: missing origin or components');
+      result.UseWithHardwareWallet = true;
```
GladosBlueWallet: **[MEDIUM]** Every `crypto-hdkey` UR gets `UseWithHardwareWallet = true` — no device metadata, no questions. Any non-hardware HD-key QR skips the watch-only scam warning and gets send/HW flow. Broad brush; hope you meant every key was born in a factory.

blue_modules/ur/index.js
```diff
@@ -370,6 +369,27 @@ function _hdKeyToResult(hdKey, masterFingerprintOverride) {
...
+  if (typeof device !== 'string') return {};
+
+  // OneKey separates the model and serial number with a colon, then may append
+  // a BTC-only marker and passphrase identifier. Keep the passphrase state so
+  // hidden wallets from the same device remain distinguishable without storing
+  // the passphrase itself.
+  const passphraseStateMatch = device.match(/-([0-9a-f]{8})$/i);
+  const passphraseState = passphraseStateMatch?.[1];
+  const deviceWithoutPassphraseState = passphraseStateMatch ? device.slice(0, -passphraseStateMatch[0].length) : device;
+  const [deviceName, serialNumber] = deviceWithoutPassphraseState.split(':', 2);
```
GladosBlueWallet: **[MEDIUM]** `split(':', 2)` silently amputates everything after the second colon — that is how `:btc` vanishes, not explicit parsing. Clever accident for OneKey today; tomorrow’s device string format will love surprising you.

GladosBlueWallet (approved):
A **FIX** for OneKey/BC-UR hardware imports: enable HW signing, auto-label accounts, show “Hardware Wallet” in UI, and suppress the watch-only scam banner for those imports. The wiring mostly holds. One warning-state regression and one over-eager UR flag still deserve a closer look — I do love a test subject that almost passes.

class/wallets/watch-only-wallet.ts
wabicai: Fixed in `35df614ab`. When hardware-wallet mode is turned off (or a hardware-wallet secret is replaced), `isWatchOnlyWarningVisible` is reset to `true`. `WalletTransactions` now also re-syncs that state on screen focus, so returning from settings reflects the transition immediately. A regression test covers a dismissed warning followed by disabling hardware-wallet mode and verifies that the warning is shown again.

blue_modules/ur/index.js
wabicai: Fixed in `35df614ab`. A standalone `crypto-hdkey` is now treated as a generic watch-only import and no longer sets `UseWithHardwareWallet` based on the UR type alone. Hardware-wallet mode is enabled only for `crypto-multi-accounts` carrying device/device-ID metadata. The decoder also rejects master/private keys and validates the 32-byte chain code plus compressed public key. Tests cover the generic watch-only path, multi-account payloads without device metadata, and invalid key material.

blue_modules/ur/index.js
wabicai: Fixed in `35df614ab`. The `split(":", 2)` parsing is gone. We now parse only OneKey’s known complete grammar (`<model>[:<serial>][:btc][-<8-hex passphrase state>]`); unknown or future device strings are preserved verbatim instead of being truncated. Device ID and firmware version are also stored as separate structured fields. A regression test verifies that an unknown OneKey suffix survives unchanged.

tests/unit/watch-only-wallet.test.js
r6mez, line 1: These fixtures are generated using the same BC-UR registry classes that the decoder consumes. That verifies internal round-tripping, but it may not catch differences in OneKey’s actual exported CBOR fields or device-string grammar.

Could we add at least one captured, sanitized OneKey crypto-multi-accounts UR fixture and assert the complete import result? The hardware-wallet classification specifically depends on real-world device metadata

class/wallets/watch-only-wallet.ts
```diff
@@ -46,6 +86,52 @@ export class WatchOnlyWallet extends LegacyWallet {
...
+        break;
+      case HDSegwitBech32Wallet.type:
+        accountType = 'Native SegWit';
+        break;
+      case HDTaprootWallet.type:
+        accountType = 'Taproot';
+        break;
+      default:
+        return super.getLabel();
+    }
```
r6mez, line 126: Should these “Hidden,” “Legacy,” and “Native SegWit” use the localization layer, in line with the repository’s localization guidelines?

r6mez (changes requested):

tests/unit/watch-only-wallet.test.js
wabicai: Addressed in dd4c4c36e. Added a five-part literal crypto-multi-accounts fixture captured from the OneKey Pro 2 production firmware encoder using deterministic secp256k1 keys and sanitized device metadata. The test feeds the raw UR parts into BlueURDecoder and verifies BTC account filtering, derivation paths and prefixes, device metadata, hardware-wallet signing/classification, warning state, and generated labels. The fixture is no longer constructed with BlueWallet’s JavaScript registry classes.

class/wallets/watch-only-wallet.ts
wabicai: Addressed in dd4c4c36e. Moved Hardware Wallet, Hidden, Legacy, Nested SegWit, Native SegWit, and Taproot into loc/en.json. WatchOnlyWallet now resolves the type and generated label through the localization layer at read time; other locale files remain untouched for Transifex. The focused wallet test, ESLint, TypeScript, and localization-key checks all pass.

ncoelho:
Add some screenshots on the PR please

PR #8801, opened 2026-07-28
ADD: check if wallet has memos before deleting

marcosrdz:


ncoelho:
What does this do?

marcosrdz:
> What does this do?

Fit the wallet has memos it lets the user know before proceeding so they have a chance to export them

ncoelho:
> > What does this do?
> 
> Fit the wallet has memos it lets the user know before proceeding so they have a chance to export them

Yes, but there is already a balance check. How does this one plays out?

marcosrdz:
> > > What does this do?
> > 
> > Fit the wallet has memos it lets the user know before proceeding so they have a chance to export them
> 
> Yes, but there is already a balance check. How does this one plays out?

It's a seperate alert. Meaning it first checks for txs and then balance.

PR #8808, opened 2026-07-30
FIX: RBF Speedup and Cancel button  not working for HD Bech32 Wallet

ojokne:
Fixes RBF `Speedup` and `Cancel` buttons not working when clicked for HD Bech32 Wallet

The buttons were also not working for HD Bech32 Watchonly wallet if the `Use with hardware wallet option` was already checked

This was due to missing else branch in the change introduced by https://github.com/BlueWallet/BlueWallet/pull/8641

### DIFF
--- a/screen/transactions/TransactionStatus.tsx
+++ b/screen/transactions/TransactionStatus.tsx
@@ -766,6 +766,11 @@ const TransactionStatus: React.FC = () => {
         ],
         { cancelable: false },
       );
+    } else {
+      navigate(route, {
+        txid: transaction.hash,
+        wallet: w,
+      });
     }
   };
 


limpbrains (approved):

screen/transactions/TransactionStatus.tsx
```diff
@@ -766,6 +766,11 @@ const TransactionStatus: React.FC = () => {
         ],
         { cancelable: false },
       );
+    } else {
```
GladosBlueWallet, line 769: **[HIGH]** Five lines fix a #8641 regression, yet no `TST:` commit and no unit test asserting `navigate('RBFBumpFee'|'RBFCancel')` fires for HD Bech32 and watch-only+HW-enabled wallets. CONTRIBUTING wants tests; this is exactly how regressions become cake. Again.

GladosBlueWallet (changes requested):
Regression fix from #8641: `navigateToRBF` lost its default navigation path, so Speed Up/Cancel did nothing for HD Bech32 (and watch-only with HW already enabled). The five-line `else` restores pre-8641 behavior — logically sound, not bloated. No new code defects spotted. **However**, you shipped a one-line regression fix with **zero** tests and a PR that skips CONTRIBUTING's demo requirements. The facility remembers #8641. Apparently you don't.

- `CONTRIBUTING.md:33` — **[MEDIUM]** PR #8808 has text explaining the bug but no screenshot or video of Speed Up/Cancel actually working. CONTRIBUTING mandates visual proof. I cannot taste your UI fixes through markdown alone — disappointing, even for science.

Overtorment (approved):
lgtm

Overtorment:
good find

Issue #8803, opened 2026-07-29
BlueWallet rejects signed PSBT for transactions with multiple UTXOs


pauloneo:
## Description

I found a possible issue when creating and signing PSBTs generated by BlueWallet.

When a transaction spends **multiple UTXOs**, the PSBT can be signed successfully by both **AirGap Vault** and **SeedSigner**. However, after importing the signed PSBT back into BlueWallet, the wallet rejects it with a generic error and does not allow the transaction to be broadcast.

### Steps to reproduce

1. Create a Bitcoin transaction in BlueWallet using **multiple UTXOs**.
2. Export the transaction as a QR (PSBT).
3. Sign the PSBT using **AirGap Vault** or **SeedSigner**.
4. Import the signed PSBT back into BlueWallet.

### Expected behavior

BlueWallet should accept the signed PSBT and allow the transaction to be broadcast.

### Actual behavior

BlueWallet rejects the signed PSBT with a generic error.

### Additional information

* Transactions spending **a single UTXO** work correctly and are accepted after signing.
* The issue only occurs when the transaction spends **multiple UTXOs**.
* The same behavior was reproduced with both AirGap Vault and SeedSigner.

While investigating the issue, I noticed that the transaction fee displayed in BlueWallet when creating the transaction does not appear to match the fee information contained in the exported PSBT QR. This is my observation during testing and may explain why the signed PSBT is later rejected, but it would need to be confirmed.

Please investigate whether the PSBT generated for multi-UTXO transactions contains inconsistent fee or input information.

Overtorment:
thanks for the report, is it possible to provide the actual wallet and psbts created? perhaps, from an empty wallet (all utxos are spent)

pauloneo:
> thanks for the report, is it possible to provide the actual wallet and psbts created? perhaps, from an empty wallet (all utxos are spent)

Unfortunately, I cannot provide data from a wallet I still intend to use—even if it contains UTXOs that have already been spent; I’m sorry... However, if you want to resolve the issue, you can simply create your own wallet with 2 or 3 UTXOs, try to replicate the send transaction, and sign it in an air-gapped environment; you will easily discover that the fees displayed on the screen should not be attributed to the PSBT in question...

Overtorment:
Ok, the fee issue is reproduced, the rejected psbt is not yet. Will keep trying

Overtorment:
[...]

PR #8809, opened 2026-07-30
TST: investigate #8803 multi-UTXO PSBT external signer import

Overtorment:
Add unit tests with synthetic mnemonic and fake UTXOs that:
- create multi-input watch-only PSBTs
- sign with embit (SeedSigner-style trim)
- import via combinePsbt / UR roundtrip

Result: signed-PSBT rejection not reproduced; watch-only UI fee mismatch vs PSBT fee is reproduced when segwitType is unset.


tests/unit/watch-only-wallet.test.js
```diff
@@ -677,6 +677,65 @@ describe('Watch only wallet', () => {
...
+    const changeAddress = w._getInternalAddressByIndex(0);
+
+    const { fee } = w.createTransaction(utxos, targets, 1, changeAddress);
+    const { fee: estimatedFee } = w.coinselect(utxos, targets, 1);
+    assert.strictEqual(estimatedFee, fee);
+  });
+
+  // #8803: SeedSigner returns a trimmed PSBT (global tx + partial_sigs only). Fixtures below were
+  // produced offline with embit (SeedSigner’s signer) from abandon…about / BIP84; 3 inputs.
+  it('combinePsbt accepts SeedSigner-trimmed multi-UTXO signed PSBT', () => {
```
GladosBlueWallet: **[HIGH]** Congratulations — 45 lines of base64 taxidermy for `combinePsbt`, a path this PR never touched. Your own #8803 investigation logged "REJECTION NOT REPRODUCED"; `f1cd1e96b` dropped the bloated suite for exactly this reason. Then commits `81cff26f8`/`fe0f6c5be` stapled the corpse back on. CI time is finite. Delete it or prove a regression this diff actually fixes.

GladosBlueWallet (changes requested):
#8803 bug fix: `WatchOnlyWallet.coinselect` now delegates to `_hdWalletInstance`, aligning SendDetails fee precalc with `createTransaction` for bare zpub/JSON imports where wrapper `segwitType` was unset. Production diff is four lines — acceptable. Your investigation already proved `combinePsbt` was innocent; re-adding that test is dead weight in the chamber.

GladosBlueWallet (approved):
Two bugfixes, minimal diff — watch-only fee estimation finally routes through the HD instance like PSBT creation (#8803), and Electrum rejections stop vanishing into "Unknown error." Regression test targets the actual divergence. I ran the review chamber twice. No survivors.

PR #8811, opened 2026-07-31
FIX: slow down animated QR frame rate 2x

Overtorment:
Double the DynamicQRCode interval from 500ms to 1000ms so fragment animation is easier for hardware wallets to scan.


limpbrains (approved):

marcosrdz (approved):

GladosBlueWallet:

PR #8815, opened 2026-08-01
Cursor rng selftest 3aa0

Overtorment:


Overtorment:
@limpbrains need your opinion on this. is that too much? or its a good addition?

class/rng-selftest.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+  assertRngSample(probe);
+
+  const sample = await randomBytes(SAMPLE_SIZE);
+  if (sample.length !== SAMPLE_SIZE) {
+    throw new Error(`randomBytes(${SAMPLE_SIZE}) returned length ${sample.length}`);
+  }
+  assertRngSample(sample);
+
+  const seen: Record<string, 1> = {};
```
GladosBlueWallet, line 61: **[HIGH]** 256 hex dedup probes after a 64KiB monobit pass — then SelfTest runs *two* 1000-wallet uniqueness loops (lines 108 and 289). That is not a smoke test; it is a three-layer cake of the same hypothesis. Even I do not rerun the same experiment this many times.

class/rng-selftest.ts
```diff
@@ -0,0 +1,69 @@
...
+    ones += popcount(sample[i]);
+  }
+  const ratio = ones / (sample.length * 8);
+  if (ratio < 0.45 || ratio > 0.55) {
+    throw new Error(`RNG monobit failed: ones ratio ${ratio.toFixed(4)}`);
+  }
+}
+
+export async function runRngSelfTest(): Promise<void> {
+  const probe = await randomBytes(32);
```
GladosBlueWallet, line 49: **[HIGH]** A 32-byte 'probe' whose monobit gate never fires (line 34 skips <1024 bytes), then you fetch 64KiB and repeat the constant check anyway. Two async calls to learn what one buffer already proves. Minimal diff, maximal waste — the cake is a lie, but the extra `randomBytes` call is real.

tests/unit/rng-selftest.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  it('skips monobit for tiny non-constant samples', () => {
+    // Too small for monobit; only constant-check applies
+    const tiny = new Uint8Array([0x00, 0xff]);
+    assertRngSample(tiny);
+  });
+});
+
+describe('runRngSelfTest', () => {
+  it('passes against platform crypto.getRandomValues', async () => {
```
GladosBlueWallet, line 39: **[HIGH]** One happy-path integration test — `await runRngSelfTest()` and pray. No duplicate-sample failure, no short-buffer return, no `crypto.getRandomValues`-missing path. CONTRIBUTING wants edge cases; you delivered a control group that always wins. Science requires failure data, not applause.

class/rng-selftest.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+import { uint8ArrayToHex } from '../blue_modules/uint8array-extras';
+import { randomBytes } from './rng';
+
+const SAMPLE_SIZE = 64 * 1024;
+const UNIQUENESS_COUNT = 256;
+
+function popcount(byte: number): number {
+  return byte.toString(2).split('1').length - 1;
```
GladosBlueWallet, line 13: **[HIGH]** `toString(2).split('1')` popcount over 65,536 bytes on the JS thread — 65k string allocations because eslint frowned at bitwise ops. 'Lightweight smoke check' that huffs its own exhaust. A single `&=` loop would finish before this finishes complaining.

screen/settings/SelfTest.tsx
```diff
@@ -100,6 +101,8 @@ export default class SelfTest extends Component {
       await new Promise(resolve => setTimeout(resolve, 1_000)); // propagate ui
 
       if (typeof navigator !== 'undefined' && navigator.product === 'ReactNative') {
+        await runRngSelfTest();
```
GladosBlueWallet, line 104: **[MEDIUM]** `runRngSelfTest()` runs before wallet generation but only under the ReactNative guard — fine for Detox, invisible to plain Node integration unless `navigator.product` is set. Not a bug today; just entropy theatre with an inconsistent audience.

GladosBlueWallet (changes requested):
**ADD:** RNG smoke checks wired into SelfTest via `runRngSelfTest()` — constant output, monobit, raw uniqueness. Logic is mostly sound, but the facility is **not** lightweight: it triple-stacks entropy checks atop two existing 1000-key wallet loops. Congratulations, test subject — redundant science.

GladosBlueWallet:

image

Overtorment:
hmmm i think its an overkill

PR #8852, opened 2026-08-15
FIX: Harden RNG by using the native secure random module

### LINKED ISSUE #8826: Harden RNG through defensive coding
In recent days I scanned/reviewed the RNG code I've contributed to in the past and wanted to provide an improvement in defensive coding. The module [rng.ts](https://github.com/BlueWallet/BlueWallet/blob/master/class/rng.ts) calls the `global.crypto` module shimed by `crypto-browserify` which calls `crypto.getRandomValues` which is shimed by `react-native-get-random-values`. This module has a fallback to cryptographically insecure function `Math.random()` in the `index.js` module wrapper (for Chrome debugging):

https://github.com/LinusU/react-native-get-random-values/blob/v1.11.0/index.js#L15

Instead of going through 3 layers of abstraction (and potential footguns) why not call the native module `NativeModules.RNGetRandomValues.getRandomBase64` that provides cryptographically secure random numbers directly? I've done this as well in StashPay:

https://github.com/onionmill/stashpay/blob/main/src/action/mnemonic.js#L4-L8

Thanks for your work on BlueWallet and for considering my suggestion!

r6mez:
## Summary

This PR hardens random byte generation by calling
`NativeModules.RNGetRandomValues.getRandomBase64()` directly instead of going
through the global `crypto.getRandomValues` polyfill.

The polyfill falls back to `Math.random()` during Chrome remote debugging,
which is not cryptographically secure. Calling the native module directly
ensures wallet secrets and encryption keys are always generated using the
platform's secure random number generator or the operation fails.

The Jest environment now provides the same native-module interface using
Node's secure `crypto.randomBytes()` implementation.

## Testing

- TypeScript and ESLint checks pass
- All 53 unit test suites pass
- 515 tests pass, with 1 existing skipped test

Closes #8826

### DIFF
--- a/class/rng.ts
+++ b/class/rng.ts
@@ -3,20 +3,37 @@
  * into one place to try and prevent mistakes when touching the crypto code.
  */
 
-// React Native: entropy via global crypto.getRandomValues (polyfilled by react-native-get-random-values)
+import { base64 } from '@scure/base';
+import { NativeModules } from 'react-native';
+
+const MAX_BYTES = 65536;
 
 /**
  * Generate cryptographically secure random bytes using native api.
+ * Throws instead of falling back to JavaScript randomness when the native
+ * bridge is unavailable, including during Chrome remote debugging.
  * @param  {number}   size      The number of bytes of randomness
  * @return {Promise.<Uint8Array>}   The random bytes
  */
 export async function randomBytes(size: number): Promise<Uint8Array> {
-  const g = globalThis as any;
-  const rnCrypto = g && g.crypto;
-  if (!rnCrypto || typeof rnCrypto.getRandomValues !== 'function') {
-    throw new Error('crypto.getRandomValues is not available');
+  if (!Number.isSafeInteger(size) || size < 1 || size > MAX_BYTES) {
+    throw new Error(`randomBytes: invalid size ${size}`);
   }
-  const bytes = new Uint8Array(size);
-  rnCrypto.getRandomValues(bytes);
+
+  const native = NativeModules.RNGetRandomValues;
+  if (typeof native?.getRandomBase64 !== 'function') {
+    throw new Error('Secure RNG unavailable: RNGetRandomValues is not linked');
+  }
+
+  const encoded = native.getRandomBase64(size);
+  if (typeof encoded !== 'string') {
+    throw new Error('Secure RNG returned a non-string value');
+  }
+
+  const bytes = base64.decode(encoded);
+  if (bytes.length !== size) {
+    throw new Error(`Secure RNG length mismatch: expected ${size}, got ${bytes.length}`);
+  }
+
   return bytes;
 }
--- a/tests/setup.js
+++ b/tests/setup.js
@@ -1,6 +1,8 @@
 /* global jest */
 
 import mockClipboard from '@react-native-clipboard/clipboard/jest/clipboard-mock.js';
+import { randomBytes as nodeRandomBytes } from 'crypto';
+import { NativeModules } from 'react-native';
 
 const consoleWarnOrig = console.warn;
 console.warn = (...args) => {
@@ -37,6 +39,10 @@ console.debug = console.log = (...args) => {
 
 global.net = require('net'); // needed by Electrum client. For RN it is proviced in shim.js
 global.tls = require('tls'); // needed by Electrum client. For RN it is proviced in shim.js
+// Jest has no linked iOS/Android native modules, so provide the bridge used by class/rng.ts.
+NativeModules.RNGetRandomValues = {
+  getRandomBase64: size => nodeRandomBytes(size).toString('base64'),
+};
 if (typeof globalThis.fetch !== 'function') {
   throw new Error('Native fetch missing; Node >= 22.11 is required (see package.json engines)');
 }
--- a/tests/unit/encryption.test.ts
+++ b/tests/unit/encryption.test.ts
@@ -1,13 +1,21 @@
 import assert from 'assert';
+import { NativeModules } from 'react-native';
 
 import * as c from '../../blue_modules/encryption';
 
 describe('unit - encryption', function () {
-  it('encrypts and decrypts', function () {
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it('encrypts and decrypts using the shared secure RNG', async function () {
+    const getRandomBase64 = jest.spyOn(NativeModules.RNGetRandomValues, 'getRandomBase64');
     const data2encrypt = 'really long data string bla bla really long data string bla bla really long data string bla bla';
-    const crypted = c.encrypt(data2encrypt, 'password');
+    const crypted = await c.encrypt(data2encrypt, 'password');
     const decrypted = c.decrypt(crypted, 'password');
 
+    expect(getRandomBase64).toHaveBeenCalledTimes(1);
+    expect(getRandomBase64).toHaveBeenCalledWith(8);
     assert.ok(crypted);
     assert.ok(decrypted);
     assert.strictEqual(decrypted, data2encrypt);
@@ -19,13 +27,7 @@ describe('unit - encryption', function () {
     } catch (e) {}
     assert.ok(!decryptedWithBadPassword);
 
-    let exceptionRaised = false;
-    try {
-      c.encrypt('yolo', 'password');
-    } catch (_) {
-      exceptionRaised = true;
-    }
-    assert.ok(exceptionRaised);
+    await assert.rejects(c.encrypt('yolo', 'password'));
   });
 
   it('handles ok malformed data', function () {
--- a/tests/unit/rng.test.ts
+++ b/tests/unit/rng.test.ts
@@ -1,0 +1,78 @@
+import { NativeModules } from 'react-native';
+
+import { randomBytes } from '../../class/rng';
+
+describe('randomBytes', () => {
+  const nativeModule = NativeModules.RNGetRandomValues;
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+    NativeModules.RNGetRandomValues = nativeModule;
+  });
+
+  it.each([
+    { useCase: 'checksum-word selection', size: 1 },
+    { useCase: 'HD and Ark wallet entropy', size: 16 },
+    { useCase: 'legacy-wallet private keys', size: 32 },
+    { useCase: 'Realm encryption keys', size: 64 },
+    { useCase: 'maximum native request', size: 65536 },
+  ])('returns $size native random bytes as a plain Uint8Array for $useCase', async ({ size }) => {
+    const getRandomBase64 = jest.spyOn(NativeModules.RNGetRandomValues, 'getRandomBase64');
+
+    const result = await randomBytes(size);
+
+    expect(getRandomBase64).toHaveBeenCalledTimes(1);
+    expect(getRandomBase64).toHaveBeenCalledWith(size);
+    expect(result).toHaveLength(size);
+    expect(result.constructor).toBe(Uint8Array);
+  });
+
+  it.each([0, -1, 1.5, NaN, Infinity, 65537, Number.MAX_SAFE_INTEGER + 1])('rejects invalid size %s', async size => {
+    const getRandomBase64 = jest.spyOn(NativeModules.RNGetRandomValues, 'getRandomBase64');
+
+    await expect(randomBytes(size)).rejects.toThrow(`randomBytes: invalid size ${size}`);
+    expect(getRandomBase64).not.toHaveBeenCalled();
+  });
+
+  it('fails clearly when the native module is unavailable', async () => {
+    NativeModules.RNGetRandomValues = undefined;
+
+    await expect(randomBytes(32)).rejects.toThrow('Secure RNG unavailable: RNGetRandomValues is not linked');
+  });
+
+  it('rejects a non-string native response', async () => {
+    jest.spyOn(NativeModules.RNGetRandomValues, 'getRandomBase64').mockReturnValue(undefined);
+
+    await expect(randomBytes(32)).rejects.toThrow('Secure RNG returned a non-string value');
+  });
+
+  it('rejects malformed base64 from the native module', async () => {
+    jest.spyOn(NativeModules.RNGetRandomValues, 'getRandomBase64').mockReturnValue('not valid base64!');
+
+    await expect(randomBytes(32)).rejects.toThrow();
+  });
+
+  it('rejects a decoded length that differs from the request', async () => {
+    jest.spyOn(NativeModules.RNGetRandomValues, 'getRandomBase64').mockReturnValue('AQI=');
+
+    await expect(randomBytes(32)).rejects.toThrow('Secure RNG length mismatch: expected 32, got 2');
+  });
+
+  it('propagates native errors including Chrome remote debugging failures', async () => {
+    jest.spyOn(NativeModules.RNGetRandomValues, 'getRandomBase64').mockImplementation(() => {
+      throw new Error('Calling synchronous methods on native modules is not supported in Chrome');
+    });
+
+    await expect(randomBytes(32)).rejects.toThrow('Calling synchronous methods on native modules is not supported in Chrome');
+  });
+
+  it('never routes through the insecure crypto polyfill', async () => {
+    const getRandomValues = jest.spyOn(globalThis.crypto, 'getRandomValues');
+    const mathRandom = jest.spyOn(Math, 'random');
+
+    await randomBytes(32);
+
+    expect(getRandomValues).not.toHaveBeenCalled();
+    expect(mathRandom).not.toHaveBeenCalled();
+  });
+});
--- a/blue_modules/encryption.ts
+++ b/blue_modules/encryption.ts
@@ -1,7 +1,7 @@
 import { cbc } from '@noble/ciphers/aes';
 import { md5 } from '@noble/hashes/legacy';
-import { randomBytes } from '@noble/hashes/utils';
 
+import { randomBytes } from '../class/rng';
 import { areUint8ArraysEqual, base64ToUint8Array, concatUint8Arrays, stringToUint8Array, uint8ArrayToBase64 } from './uint8array-extras';
 
 /**
@@ -53,9 +53,9 @@ const BLOCK_LEN = 16;
  * change a drop-in replacement so existing encrypted wallets on user
  * devices remain readable, with no migration step.
  */
-export function encrypt(data: string, password: string): string {
+export async function encrypt(data: string, password: string): Promise<string> {
   if (data.length < 10) throw new Error('data length cant be < 10');
-  const salt = randomBytes(SALT_LEN);
+  const salt = await randomBytes(SALT_LEN);
   const kdf = evpBytesToKeyMd5(stringToUint8Array(password), salt, KEY_LEN + IV_LEN);
   const key = kdf.subarray(0, KEY_LEN);
   const iv = kdf.subarray(KEY_LEN);
--- a/class/blue-app.ts
+++ b/class/blue-app.ts
@@ -220,7 +220,7 @@ export class BlueApp {
     let data = await this.getItem('data');
     // TODO: refactor ^^^ (should not save & load to fetch data)
 
-    const encrypted = encryption.encrypt(data, password);
+    const encrypted = await encryption.encrypt(data, password);
     data = [];
     data.push(encrypted); // putting in array as we might have many buckets with storages
     data = JSON.stringify(data);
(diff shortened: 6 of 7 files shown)


class/rng.ts
```diff
@@ -3,20 +3,16 @@
  * into one place to try and prevent mistakes when touching the crypto code.
  */
 
-// React Native: entropy via global crypto.getRandomValues (polyfilled by react-native-get-random-values)
+import { NativeModules } from 'react-native';
 
 /**
  * Generate cryptographically secure random bytes using native api.
  * @param  {number}   size      The number of bytes of randomness
  * @return {Promise.<Uint8Array>}   The random bytes
```
tanx, line 16: Heads up... this function should return `Promise.<Uint8Array>` ...

class/rng.ts
```diff
@@ -3,20 +3,16 @@
...
-  if (!rnCrypto || typeof rnCrypto.getRandomValues !== 'function') {
-    throw new Error('crypto.getRandomValues is not available');
-  }
-  const bytes = new Uint8Array(size);
-  rnCrypto.getRandomValues(bytes);
-  return bytes;
+  // this prevents crypto.getRandomValues from returning insecure random values
+  // in chrome debugging mode.
+  const bytes = NativeModules.RNGetRandomValues.getRandomBase64(size);
+  return Buffer.from(bytes, 'base64');
```
tanx: ... it returns a `Buffer` object. Instead use the function [Uint8Array.fromBase64()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64).

tests/setup.js
```diff
@@ -37,6 +39,9 @@ console.debug = console.log = (...args) => {
 
 global.net = require('net'); // needed by Electrum client. For RN it is proviced in shim.js
 global.tls = require('tls'); // needed by Electrum client. For RN it is proviced in shim.js
+NativeModules.RNGetRandomValues = {
+  getRandomBase64: size => nodeRandomBytes(size).toString('base64'),
+};
```
tanx, line 45: Why is this necessary? The RN module should already be exporting this function.

tanx (changes requested):
Thanks for picking up my suggestion @r6mez ... heads up as there is a bug in your return value

class/rng.ts
tanx: Is there a unit test for this function to verify the return value and check edge cases? If not, one should be added as this function is obviously crucial to security and should never be allowed to break via a PR.

tests/setup.js
r6mez: The module is exported on-device, but Jest does not load linked iOS/Android native modules. Since `randomBytes()` calls the native bridge directly, the test environment needs this mock. I added a comment and focused tests to make that explicit.

class/rng.ts
r6mez: Although `Buffer` is technically a `Uint8Array` subclass, i fixed this.

r6mez: and also added a unit test for this funciton

class/rng.ts
```diff
@@ -3,20 +3,17 @@
...
-  if (!rnCrypto || typeof rnCrypto.getRandomValues !== 'function') {
-    throw new Error('crypto.getRandomValues is not available');
-  }
-  const bytes = new Uint8Array(size);
-  rnCrypto.getRandomValues(bytes);
-  return bytes;
+  // this prevents crypto.getRandomValues from returning insecure random values
+  // in chrome debugging mode.
+  const bytes = NativeModules.RNGetRandomValues.getRandomBase64(size);
+  return Uint8Array.from(Buffer.from(bytes, 'base64'));
```
tanx: `Buffer` is a node.js construct and shouldn't be necessary here. Why not use the native JS function [Uint8Array.fromBase64()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64)?

tanx (changes requested):
Thanks for your changes. One more nit ...

class/rng.ts
r6mez: `Uint8Array.fromBase64()` isn’t supported by the Hermes version bundled with React Native 0.85.3, so it would be undefined at runtime. I think it's replacable by `Uint8Array.from(atob(base64), character => character.charCodeAt(0))`

What do you think ?

tanx: Ah, I see. Thanks for explaining. Then I'd leave it as is.

tanx (approved):
utACK from my side. @Overtorment maybe let's get more eyes on this

class/rng.ts
```diff
@@ -3,20 +3,17 @@
  * into one place to try and prevent mistakes when touching the crypto code.
  */
 
-// React Native: entropy via global crypto.getRandomValues (polyfilled by react-native-get-random-values)
+import { NativeModules } from 'react-native';
+import { Buffer } from 'buffer';
```
Overtorment: we try to move away from Buffer - its deprecated in favor Uint8Array (it was a temporary construct for nodejs only)

tests/unit/rng.test.ts
```diff
@@ -0,0 +1,38 @@
...
+import { randomBytes } from '../../class/rng';
+
+describe('randomBytes', () => {
+  const getRandomBase64 = jest.spyOn(NativeModules.RNGetRandomValues, 'getRandomBase64');
+
+  afterEach(() => {
+    getRandomBase64.mockReset();
+  });
+
+  it('returns the requested bytes as a plain Uint8Array', async () => {
```
GladosBlueWallet: **[HIGH]** Congratulations: you mocked base64, fed it to a spy, then proved Buffer decodes your own payload. CONTRIBUTING.md explicitly forbids this circular nonsense — `setup.js` already wires real `nodeRandomBytes`; test length, type, and failures through that instead of rehearsing mock theater.

tests/unit/rng.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  });
+
+  it('supports requesting zero bytes', async () => {
+    getRandomBase64.mockReturnValue('');
+
+    await expect(randomBytes(0)).resolves.toEqual(new Uint8Array());
+    expect(getRandomBase64).toHaveBeenCalledWith(0);
+  });
+
+  it('propagates failures from the native secure random module', async () => {
```
GladosBlueWallet: **[HIGH]** Throwing from a mock and watching it bubble through two lines of wrapper code isn't a test — it's a mirror. Zero coverage that Chrome-debug insecure-RNG bypass is actually gone; the subject under test never leaves the simulation.

class/rng.ts
```diff
@@ -3,20 +3,17 @@
...
-  const rnCrypto = g && g.crypto;
-  if (!rnCrypto || typeof rnCrypto.getRandomValues !== 'function') {
-    throw new Error('crypto.getRandomValues is not available');
-  }
-  const bytes = new Uint8Array(size);
-  rnCrypto.getRandomValues(bytes);
-  return bytes;
+  // this prevents crypto.getRandomValues from returning insecure random values
+  // in chrome debugging mode.
+  const bytes = NativeModules.RNGetRandomValues.getRandomBase64(size);
```
GladosBlueWallet: **[MEDIUM]** Chrome remote debugging cannot invoke synchronous native modules — this throws and bricks every wallet/entropy path that calls `randomBytes`. Safer than `Math.random()` keys, yes; still an undocumented runtime crater for anyone debugging with Chrome.

class/rng.ts
GladosBlueWallet: **[MEDIUM]** You request `size` bytes and blindly trust whatever base64 decodes to — no `decoded.length === size` guard. A malformed bridge response hands truncated entropy to `ECPair.makeRandom` and friends with no alarm. For a wallet RNG wrapper, that's negligent.

class/rng.ts
GladosBlueWallet: **[LOW]** The old code threw a readable error when crypto RNG was unavailable. Now `RNGetRandomValues` missing yields a vague TypeError on `.getRandomBase64`. Less helpful when the native module fails to link.

GladosBlueWallet (changes requested):
**Intent:** Security **FIX** — stop `react-native-get-random-values` from silently feeding `Math.random()` entropy during Chrome remote debugging by bypassing `crypto.getRandomValues` and calling `NativeModules.RNGetRandomValues.getRandomBase64` directly; second commit restores a plain `Uint8Array` (not `Buffer`). Production diff is lean and correctly targets the polyfill's insecure fallback. Commits honor `CONTRIBUTING.md` (`FIX:` prefix, atomic). The test chamber, however, is testing its own decoys — mock round-trips, zero edge-case coverage of the actual security regression, and no validation that decoded output length matches `size`.

Overtorment (changes requested):

GladosBlueWallet:

image

GladosBlueWallet:

image

[all PRs for @ojokne] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/ojokne

class/rng.ts
tanx: @r6mez maybe your suggestion to use `atob` instead of `Buffer.from()` below is the right way to go then ...

tanx: `@scure/base` is already a direct dependency at `2.2.0`, decodes strictly (throws on invalid characters instead of truncating), and returns a `Uint8Array` natively

```javascript
import { base64 } from '@scure/base';

const MAX_BYTES = 65536; // parity with WebCrypto / the polyfill's quota
```

class/rng.ts
tanx: The following would check buffer size, do sanity checks throughout, and decode base64 safely to a `Uint8Array` (without any additional dependencies)

```javascript
  if (!Number.isSafeInteger(size) || size < 1 || size > MAX_BYTES) {
    throw new Error(`randomBytes: invalid size ${size}`);
  }
  const native = NativeModules.RNGetRandomValues;
  if (typeof native?.getRandomBase64 !== 'function') {
    throw new Error('Secure RNG unavailable: RNGetRandomValues is not linked');
  }
  const encoded = native.getRandomBase64(size);
  if (typeof encoded !== 'string') {
    throw new Error('Secure RNG returned a non-string value');
  }
  const bytes = base64.decode(encoded); // strict: throws on malformed input
  if (bytes.length !== size) {
    throw new Error(`Secure RNG length mismatch: expected ${size}, got ${bytes.length}`);
  }
  return bytes;
```

tests/unit/rng.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    getRandomBase64.mockImplementation(() => {
+      throw new Error('Secure random generation failed');
+    });
+
+    await expect(randomBytes(32)).rejects.toThrow('Secure random generation failed');
+  });
```
tanx, line 77: An additional safety check:

```javascript
  });
  
  it('never routes through the insecure polyfill', async () => {
    const getRandomValues = jest.spyOn(globalThis.crypto, 'getRandomValues');
    const mathRandom = jest.spyOn(Math, 'random');
    await randomBytes(32);
    expect(getRandomValues).not.toHaveBeenCalled();
    expect(mathRandom).not.toHaveBeenCalled();
  });
```

tanx (changes requested):
Thanks @Overtorment and @GladosBlueWallet 🙂 @r6mez here are some suggestions  from Opus 5 based on the provided feedback. I haven't tested these yet, so feel free to see them purely as inspiration and please verify whether this makes sense. Thanks!

tanx:
Something else Opus caught ... `@noble/hashes` is a second entropy source in `blue_modules/encryption.ts`:

https://github.com/BlueWallet/BlueWallet/blob/e242791752cb79f8372305472abf3623523e2465/blue_modules/encryption.ts#L3

We should replace that with `rng.ts` and make sure to only use this implementation of `randomBytes()` going forward.

class/rng.ts
r6mez: Done. Replaced Buffer with @scure/base, added the 65,536-byte limit, and kept the result as a plain Uint8Array.

tests/unit/rng.test.ts
r6mez: Done. Tests now confirm neither crypto.getRandomValues nor Math.random is called and cover the fail-closed Chrome path.

class/rng.ts
r6mez: Done. Decoded output must exactly match the requested size or randomBytes throws.

tests/unit/rng.test.ts
r6mez: Done — happy-path tests now use the Node-backed setup RNG and assert the requested length and exact Uint8Array type.

class/rng.ts
r6mez: Done. A missing or unlinked RNGetRandomValues module now throws a clear secure-RNG error.

r6mez: Handled fail-closed. Chrome sync-native failures now return a clear error instead of falling back to insecure randomness.

tests/unit/rng.test.ts
r6mez: Added — the test now asserts that neither crypto.getRandomValues nor Math.random is used.

class/rng.ts
r6mez: Done. Added size bounds, module and type checks, strict decoding, and decoded-length validation.

r6mez:
@tanx Done. encryption.ts now uses class/rng.ts for its salt, and all encrypt() callers were updated for the async RNG path.

class/rng.ts
```diff
@@ -3,20 +3,49 @@
...
+  if (!Number.isSafeInteger(size) || size < 1 || size > MAX_BYTES) {
+    throw new Error(`randomBytes: invalid size ${size}`);
+  }
+
+  const native = NativeModules.RNGetRandomValues;
+
+  if (typeof native?.getRandomBase64 !== 'function') {
+    throw new Error(
+      'Secure RNG unavailable: RNGetRandomValues is not linked; Chrome remote debugging cannot access synchronous native modules',
+    );
```
tanx: Shorten error message by removing `Chrome remote debugging cannot access synchronous native modules'` and write throw statement in one line.

class/rng.ts
```diff
@@ -3,20 +3,49 @@
...
-  const g = globalThis as any;
-  const rnCrypto = g && g.crypto;
-  if (!rnCrypto || typeof rnCrypto.getRandomValues !== 'function') {
-    throw new Error('crypto.getRandomValues is not available');
+  if (!Number.isSafeInteger(size) || size < 1 || size > MAX_BYTES) {
+    throw new Error(`randomBytes: invalid size ${size}`);
+  }
+
+  const native = NativeModules.RNGetRandomValues;
+
```
tanx: Remove newline here

class/rng.ts
```diff
@@ -3,20 +3,49 @@
...
+  }
+
+  let encoded: unknown;
+  try {
+    encoded = native.getRandomBase64(size);
+  } catch (error) {
+    if (error instanceof Error && /synchronous methods on native modules/i.test(error.message)) {
+      throw new Error('Secure RNG unavailable during Chrome remote debugging: synchronous native modules cannot be called');
+    }
+    throw error;
```
tanx: What's the point of catching and rethrowing specific error messages for environments that aren't supported?

Calling in one line and letting that invocation throw is cleaner:
```js
const encoded = native.getRandomBase64(size);
```

class/rng.ts
```diff
@@ -3,20 +3,49 @@
...
   }
-  const bytes = new Uint8Array(size);
-  rnCrypto.getRandomValues(bytes);
+
```
tanx, line 27: Remove newline here

class/rng.ts
tanx: Shorten error message by removing `Chrome remote debugging cannot access synchronous native modules'` and write throw statement in one line.

tanx (changes requested):
Thanks for your changes. I made some more comments on this iteration.

GladosBlueWallet:

image

GladosBlueWallet:

image

[all PRs for @ojokne] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/ojokne

class/rng.ts
ojokne: > `@scure/base` is already a direct dependency at `2.2.0`, decodes strictly (throws on invalid characters instead of truncating), and returns a `Uint8Array` natively
> 
> ```js
> import { base64 } from '@scure/base';
> 
> const MAX_BYTES = 65536; // parity with WebCrypto / the polyfill's quota
> ```

We also already have `base64ToUint8Array` utility with no external dependency, defined in `blue_modules/uint8rray-extras` ,  I would prefer using the existing utility for consistency with rest of the codebase. @Overtorment / @limpbrains could you advise here?

class/rng.ts
r6mez: Done

class/rng.ts
r6mez: done

class/rng.ts
r6mez: yeah sure. great point.

class/rng.ts
r6mez: done

GladosBlueWallet:

image

GladosBlueWallet:

image

[all PRs for @ojokne] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/ojokne

tanx (approved):
utACK on this iteration. Opus 5 also agreed this iteration is a security upgrade. What say the others, as there is a bit of refactoring introducing async encryption logic?

GladosBlueWallet:

image

GladosBlueWallet:

image

[all PRs for @ojokne] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/ojokne

limpbrains (approved):

GladosBlueWallet:

image

[all PRs for @ojokne] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/ojokne

class/rng.ts
tanx: Does `base64ToUint8Array` catch the same edge cases? It might be worth it to use `scure/base` for this case since keygen is a sensative code path. But your call obv…

ojokne: its implementation is here
https://github.com/BlueWallet/BlueWallet/blob/5ea81bcc400c90f7ddf62989075f3f235e0a8c7b/blue_modules/uint8array-extras/index.js#L174-L177

I ran a brief test against both functions and got the following results.

`aGk=` is `base 64` for  the word`hi`

`atob()` vs `@scure/base`

| Input | Meaning / Explanation | `atob()` | `@scure/base` |
|---|---|---|---|
| `aGk=` | Valid, canonical Base64 for `hi` | decodes | decodes |
| `aGk` | Valid Base64 content but missing padding (`=`) | decodes | throws |
| `aGl=` | Non-canonical Base64 with non-zero trailing bits; also decodes to `hi` | decodes | throws |
| `aGk=\n` | Valid Base64 followed by a newline/whitespace | decodes | throws |
| `aGk===` | Excess padding | throws | throws |
| `aG!k=` | Contains an invalid Base64 character (`!`) | throws | throws |

`@scure/base` decodes strictly as you had earlier mentioned so I think we can go with it

GladosBlueWallet:

image

[all PRs for @ojokne] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/ojokne

tests/unit/encryption.test.ts
```diff
@@ -19,13 +27,7 @@ describe('unit - encryption', function () {
...
     assert.ok(!decryptedWithBadPassword);
 
-    let exceptionRaised = false;
-    try {
-      c.encrypt('yolo', 'password');
-    } catch (_) {
-      exceptionRaised = true;
-    }
-    assert.ok(exceptionRaised);
+    await assert.rejects(c.encrypt('yolo', 'password'));
```
ojokne, line 30: I think this test can be improved by also checking the exact message it rejects with `data length cant be < 10`

something like this

```js
await assert.rejects(c.encrypt('yolo', 'password'), {
      message: 'data length cant be < 10',
    });
```

this will ensure we are not only testing if the Promise rejects, but also that it rejects for the expected reason, what do you think?

tests/unit/rng.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  it('never routes through the insecure crypto polyfill', async () => {
+    const getRandomValues = jest.spyOn(globalThis.crypto, 'getRandomValues');
+    const mathRandom = jest.spyOn(Math, 'random');
+
+    await randomBytes(32);
+
+    expect(getRandomValues).not.toHaveBeenCalled();
+    expect(mathRandom).not.toHaveBeenCalled();
+  });
```
ojokne, line 77: @tanx 
would it make sense to add this test to verify that the decoded bytes match the value returned by the native module, so we know the bytes aren't being altered?

```js
it('returns the bytes provided by the native module', async () => {
  jest.spyOn(NativeModules.RNGetRandomValues, 'getRandomBase64').mockReturnValue('AQIDBA==');

  const result = await randomBytes(4);

  expect(result).toEqual(new Uint8Array([1, 2, 3, 4]));
});

```

tanx: @ojokne good call! What do you think @r6mez ?

tanx:
@Overtorment some feedback on process... I think the reminders by @GladosBlueWallet are useful to bring attention to open PRs, but after three days I unsubscribed due to too many notifications and missed other people’s useful discussion around the PR until I was tagged.

Issue #3371, opened 2021-07-10
Display fp & path for HD watch-only


Overtorment:
AFAIR we already display fp (but only in advanced mode), but not path.

demerzel3:
how can I create a watch only HD wallet with a specific derivation path? I can import an xpub but I can see no way to specify the path. what am I missing?

Overtorment:
you can import watch-only by pairing with keystone or foundation HW wallets. the feed to bluewallet structure like 

```
{"ExtPubKey":"zpub6riZchHnrWzhhZ3Z4dhCJmesGyafMmZBRC9txhnidR313XJbcv4KiDubderKHhL7rMsqacYd82FQ38e4whgs8Dg7CpsxX3dSGWayXsEerF4","MasterFingerprint":"7D2F0272","AccountKeyPath":"84'\/0'\/0'","CoboVaultFirmwareVersion":"2.6.1(BTC-Only)"}
```

which has path.

in other cases when we do not have path on hand (when we imported plain xpub string) is somewhat safe to assume its default BIP84 path for native segwit, BIP49 path for wrapped segwit  and BIP44 path for legacy

demerzel3:
@Overtorment another quick question for you: I see an `additionalProperties` argument in a couple of places in the code that is supposed to hold the value for the master fingerprint for watch only wallets. Here's [one place](https://github.com/demerzel3/BlueWallet/blob/038890a1a9912b6d5d3ed7e4e12f3edd4e499bd8/class/wallet-import.js#L95), there are a couple more.

From what I can see though, the master fingerprint for watch only addresses is set in abstract wallet [here](https://github.com/demerzel3/BlueWallet/blob/038890a1a9912b6d5d3ed7e4e12f3edd4e499bd8/class/wallets/abstract-wallet.js#L166-L221) inside of `setSecret`.

In short, I can't see `additionalProperties` being used at all, am I missing something or is it safe to remove?

Overtorment:
i dont think that `additionalProperties` is used anymore, can cautiously remove it

Issue #825, opened 2020-01-26
Compatibility with Electrum Wallet


dk-from-twitter:
Hello,
I tried to use a cold storage laptop running Electrum for private key management together with importing the zpub in BlueWallet for receiving transactions and generating unsigned spending ones.
It doesn't work.
I get an error when I try to import the unsigned transaction from BlueWallet to the laptop using the laptop camera and scanning the QR code on the mobile.
I also get an error when I try to use the BlueWallet to broadcast a signed transaction from the laptop. It fails to decode the QR of the signed transaction from the laptop screen.

Thanks.
![IMG_2933](https://user-images.githubusercontent.com/48622801/73135016-80688800-4035-11ea-9dc6-9b800ef72603.jpeg)
![Screenshot from 2020-01-26 12-09-35](https://user-images.githubusercontent.com/48622801/73135017-80688800-4035-11ea-9b7e-b7debf3efaa2.png)

Overtorment:
Hi! We are working hard on Coldcard integration. We made lots of progress, so your problem might have been fixed as well.
In particular, 'magic number' error looks like when youre trying to feed transaction hex to app when it actually expects base64 PSBT.

Anyway, you can try take latest apk build from https://github.com/BlueWallet/BlueWallet/pull/793 for Android, or wait for official 5.0.0 release.

csensitive:
I can confirm the issue with transferring unsigned (or PSBT) transactions still persists and errors the exact same output as OP posted. I haven't tried importing a signed transaction from Ethereum desktop but my guess is it also fails due to same out-/input error.

raytorres280:
Not sure if related, but I did receive PBST errors as a result of my master fingerprint not being imported correctly and having fingerprint '00000000'. Would be curious to see if this still occurs once #4693 is merged in.

PR #8861, opened 2026-08-22
ADD: edit derivation path & master fingerprint for watch-only wallets

beeant:
## Summary

Watch-only wallets imported from a bare xpub/ypub/zpub get a guessed derivation path (the class default, e.g. `m/84'/0'/0'`) and a master fingerprint of `0`, because a bare extended key doesn't carry either. Both values then get baked into every PSBT's `bip32Derivation`, so external signers (Electrum with a non-BIP39 seed, hardware wallets) reject or mislabel the transaction. See #825 (the `00000000` fingerprint PSBT errors), and #3371 which added read-only display of these fields.

For a watch-only wallet both values are just signer metadata. They don't affect how addresses are derived from the xpub, so the user should be able to correct them. This PR makes the derivation path and Master Fingerprint rows in Wallet Details tappable for HD watch-only wallets, and lets the custom derivation path import screen handle watch-only keys.

Example: an Electrum native-segwit seed derives its account at `m/0'`. Importing its zpub today shows `m/84'/0'/0'` and fingerprint `00000000`. With this PR you enter the real path plus the fingerprint from Electrum (`wallet.keystore.get_root_fingerprint()`), and the PSBTs BlueWallet creates carry correct origin info.

## Changes

- `WatchOnlyWallet.setDerivationPath(path)` stores the edited path on the inner HD instance (survives save/load) and pins `segwitType` first, so editing the path can never change address derivation or the wallet ID.
- New `WatchOnlyWallet.setMasterFingerprintFromHex(hex)`.
- `WalletDetails`: the two advanced rows get an `onPress` (watch-only HD only) that prompts, validates, saves and refreshes. Non-watch-only wallets are untouched.
- Custom derivation path import now works for watch-only keys: the field starts at the key's own path, and a bare xpub typed at `m/84'`/`m/49'`/`m/86'` account paths gets that script type (`setSecretForCustomPathImport`).
- Import and Wallet Details both require `m/` (`validateBip32`). Custom-path import also converts `h` to `'` (same as Wallet Details).
- Bad key-origin hex no longer crashes the import screens.
- `WalletExport` emits the `tr()` descriptor based on the wallet's actual script type instead of the path string (`getSecretForExport`).
- Export/Backup of a zpub, ypub, or xpub still writes the bare key, so path and fingerprint are not kept. The edit dialogs warn. Descriptor export is #8884.
- Fixed a pre-existing fingerprint conversion bug: any fingerprint whose last byte is `00` (1 in 256) produced odd-length hex, breaking display and PSBT creation. Now `padStart(8, '0')` in all converters (also affects seed wallets with an imported fingerprint, hence the base-class change).
- `prompt()` no longer hides the help message when a default value is set; iOS curly quotes are normalized in the path editor.
- Locale strings and unit tests: getID stability, save/load round trips, address stability with cleared caches, PSBT `bip32Derivation` content, export/import round trips, fingerprint edge cases.

## Self review

Four self-review rounds (see the closing comments); should-fix findings are in later commits, threads left open. Follow-ups filed instead of growing this PR: #8879, #8881, #8882, #8884.

## Test plan

- `npx jest tests/unit/watch-only-wallet.test.js`, new and existing tests pass
- `npx tsc --noEmit` and ESLint on touched files, clean
- Manual: import a bare zpub, open Wallet Details, expand advanced, edit path to `m/0'` and set the fingerprint. Both display. Create a PSBT and confirm `bip32Derivation` has the entered fingerprint and `m/0'/0/i` paths. Restart the app, values persist, addresses and wallet ID unchanged. Export/Backup restores the same wallet.

Fixes the metadata half of #825.


ncoelho:
See https://github.com/BlueWallet/BlueWallet/pull/8512

ncoelho:
Maybe it would be nicer to allow "search accounts" on import, allow the user to choose and add custom derivation path if needed. Same as seed import flow with "search accounts" option on.

Don't think changing the path directly on the wallet details is a good idea after a wallet with transactions and balance is already unwrapped.

beeant:
Thanks for taking a look! Hadn't seen #8512. Happy to rebase the fingerprint half on top of it or split it out. This PR also covers the derivation path, which is the other half of what breaks PSBTs for external signers (#825).

On editing the path of a wallet that already has balance/transactions: fully agree that would be unacceptable if it could change derivation, that's why the edit is metadata-only by design. For watch-only wallets the path plays no role in deriving addresses from the xpub; it only labels the `bip32Derivation` origin in PSBTs. `setDerivationPath()` now pins `segwitType` before storing the new path, so `init()`'s path-based type detection can never rebuild a different script type afterwards. There's a unit test asserting the first external address is byte-identical after editing to a hostile path (`m/49'/0'/0'` on a zpub) and re-running `init()`. Balance, history and addresses are untouched by construction.

On doing it at import instead: agreed that's the nicer UX for the path, and I'm happy to add a path picker to the search-accounts flow as a follow-up. Two cases import-time selection can't cover, though:

1. the master fingerprint, a bare xpub/ypub/zpub carries none, so there's nothing to discover at import; the user has to type it from their signing device, which is inherently a post-import edit (same shape as #8512);
2. wallets already imported today (everyone currently hitting #825) would have to be deleted and re-imported to fix their metadata. Also imports via the home-screen QR scan currently bypass the search-accounts options entirely.

Happy to reshape however you prefer, add a confirmation dialog, split the fingerprint part to align with #8512, and/or do the import-flow picker as a follow-up.

beeant:
Implemented your import-time suggestion in 9d0b806: with "Search accounts" on, importing a bare xpub/ypub/zpub through the custom derivation path screen now works. It previously only built seed-based candidates there, so a watch-only key produced nothing. It offers a single watch-only candidate at the chosen path (script type stays determined by the key itself).

So the PR now covers both routes: pick the path at import (new wallets), or correct it in wallet details (existing wallets + the fingerprint, which can't be discovered at import). If you'd rather land only the import-time part, or align the fingerprint half with #8512, happy to split. Just say which shape you want.

Ugarba202 (commented):
Checked both the import-time "Search accounts" flow (`9d0b806`) and the post-import editing in `WalletDetails`. having both routes makes sense because bare xpubs/zpubs don't carry master fingerprints at import time, so users still need a way to add the 8-hex fingerprint after import for external signers to accept PSBTs. also verified that pinning `segwitType` protects against script-type mutations on re-init. clean implementation and tests pass nicely.

leexmario-oss:
Hi @ncoelho — this PR appears to match the watch-only wallet behavior I need: importing bare xpub/ypub/zpub wallets and setting the derivation path and master fingerprint for signing workflows. The available checks look good. Could you confirm whether anything remains before this can be merged? Thanks!

ncoelho:
This is going in a nice direction, adding more eyes to it.

GladosBlueWallet (approved):
Feature PR: watch-only HD wallets get editable derivation path and master fingerprint (PSBT signer metadata), plus custom-path import and discovery entry for bare xpub/ypub/zpub. Commits follow CONTRIBUTING (`ADD`/`FIX`, atomic). Core `setDerivationPath` segwitType pinning prevents address drift on re-init — the one thing that would have been catastrophic. Unit tests exercise real wallet behavior, not mock theater. I ran the review chamber dry; no anomalies worth filing.

GladosBlueWallet:

image

GladosBlueWallet:

image

GladosBlueWallet:

image

GladosBlueWallet:

image

[all PRs for @ojokne] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/ojokne

leexmario-oss:
Hi @ncoelho, thank you for the review. This PR matches my real use case: watch-only xpub/ypub/zpub import with a custom derivation path and master fingerprint for external-signing/PSBT workflows. I can provide end-user testing on Android and/or iOS if helpful. I checked the current Actions results: `build` fails at `Ensure Correct Branch`, and `buildReleaseApk` fails at `Build and sign APK`. Could you confirm whether these are CI/release-environment issues or the remaining blocker to merge? I can test with a non-sensitive test wallet; no private wallet data is needed.

leexmario-oss:
Additional end-user test information from @leexmario-oss: iPhone 17 Pro Max running iOS 27 beta 7. I imported a watch-only wallet via xpub without sharing the address. The displayed balance intermittently appears and disappears (turning on and off). I have not tested PSBT creation yet. This balance-sync/display issue may be separate from the derivation-path and master-fingerprint work, but I wanted to report it for investigation.

class/wallets/watch-only-wallet.ts
```diff
@@ -334,9 +334,23 @@ export class WatchOnlyWallet extends LegacyWallet {
...
+    if (!this.isHd()) throw new Error("Not a HD watch-only wallet, can't use derivation path");
+    if (!this._hdWalletInstance) this.init();
+    // the derivation path is metadata for signers (PSBT bip32Derivation); editing it must never
+    // change how addresses are derived, so pin the script type before init() can re-detect it from the path
+    if (!this.segwitType && this._hdWalletInstance) {
+      if (this._hdWalletInstance instanceof HDTaprootWallet) this.segwitType = 'p2tr';
+      else if (this._hdWalletInstance instanceof HDSegwitP2SHWallet) this.segwitType = 'p2sh(p2wpkh)';
+      else if (this._hdWalletInstance instanceof HDSegwitBech32Wallet) this.segwitType = 'p2wpkh';
+      else if (this._hdWalletInstance instanceof HDLegacyP2PKHWallet) this.segwitType = 'p2pkh';
+    }
```
Overtorment: a bit hard to read

screen/wallets/WalletDetails.tsx
```diff
@@ -795,7 +842,13 @@ const WalletDetails: React.FC = () => {
                   )}
                   {wallet.allowMasterFingerprint && wallet.allowMasterFingerprint() && (
                     <SettingsListItem
-                      onPress={isMasterFingerPrintVisible ? undefined : onViewMasterFingerPrintPress}
+                      onPress={
+                        isMasterFingerPrintVisible
+                          ? isWatchOnlyHd
+                            ? onEditMasterFingerprintPress
+                            : undefined
+                          : onViewMasterFingerPrintPress
```
Overtorment: nested ternary are absolutely impossible to read

tests/unit/watch-only-wallet.test.js
```diff
@@ -911,4 +911,56 @@ describe('BC-UR', () => {
...
+    const w = new WatchOnlyWallet();
+    w.setSecret('bc1qt4t9xl2gmjvxgmp5gev6m8e6s9c85979ta7jeh'); // plain address, not HD
+    assert.throws(() => w.setDerivationPath("m/0'"));
+
+    const hd = new WatchOnlyWallet();
+    hd.setSecret('zpub6r7jhKKm7BAVx3b3nSnuadY1WnshZYkhK8gKFoRLwK9rF3Mzv28BrGcCGA3ugGtawi1WLb2vyjQAX9ZTDGU5gNk2bLdTc3iEXr6tzR1ipNP');
+    hd.init();
+    assert.throws(() => hd.setMasterFingerprintFromHex('nothex00'));
+    assert.strictEqual(hd.getMasterFingerprintHex(), '00000000');
+  });
```
Overtorment: nice touch!

Overtorment (approved):
lgtm! glados is happy, which is a good sign. will run the whole e2e suite separately

Overtorment:
running tests here: https://github.com/BlueWallet/BlueWallet/pull/8877

Overtorment:
@ojokne wdyt? if we merge this, your https://github.com/BlueWallet/BlueWallet/pull/8512 will be obsolete?

ojokne:
> @ojokne wdyt? if we merge this, your #8512 will be obsolete?

yes it will be obolete and can be closed after merging this one

class/wallets/watch-only-wallet.ts
beeant: Dropped the instanceof chain - each HD class already declares its own segwitType, so this is now `this._hdWalletInstance.segwitType ?? 'p2pkh'` (legacy is the only one that doesn't declare it, and 'p2pkh' is the tag init() matches to rebuild it). dca41189

screen/wallets/WalletDetails.tsx
beeant: Pulled it out into onMasterFingerprintPress above the return - defaults to the view handler, and only when the fingerprint is visible does it become edit (watch-only HD) or undefined. e24b1596

class/wallets/watch-only-wallet.ts
```diff
@@ -334,9 +334,22 @@ export class WatchOnlyWallet extends LegacyWallet {
...
+    if (!this.isHd()) throw new Error("Not a HD watch-only wallet, can't use derivation path");
+    if (!this._hdWalletInstance) this.init();
+    // the derivation path is metadata for signers (PSBT bip32Derivation); editing it must never
+    // change how addresses are derived, so pin the script type before init() can re-detect it from the path
+    if (!this.segwitType && this._hdWalletInstance) {
+      // every HD class declares its own segwitType; HDLegacyP2PKHWallet is the one that does not,
+      // and 'p2pkh' is the tag init() matches to rebuild it
+      this.segwitType = this._hdWalletInstance.segwitType ?? 'p2pkh';
+    }
+    this._derivationPath = path;
```
beeant: **[bug]** `setDerivationPath` now assigns `this._derivationPath = path`. `AbstractWallet.getID()` hashes `type + secret + passphrase + this._derivationPath`, so a metadata-only edit changes the wallet's identity. Watch-only HD wallets already have `_derivationPath` set by `setSecret` (e.g. `m/84'/0'/0'` for a bare zpub), and the previous setter only updated `_hdWalletInstance`, so this is a behavior change. After the user corrects the path, `wallets.find(w => w.getID() === walletID)` misses the same object. `SendDetails` then falls back to `suitable[0]` (a different onchain wallet). Realm rows, carousel keys, and stacked screens that still hold the old `walletID` in route params (`WalletTransactions`, `psbtWithHardwareWallet`, `ReceiveDetails`, `CoinControl`) also break or leak. Fingerprint edits do not have this problem; only the path write does.

**Suggestion:** Persist the edited path on `_hdWalletInstance` (already serialized by `saveToDisk`) and pin `segwitType`, but do not change `this._derivationPath` — that field is part of the ID and is only needed for `init()`'s type detection. Alternatively, snapshot `getID()` so path metadata cannot move it. Do not override `getID()` to drop the path entirely: existing watch-only IDs already include the `setSecret` default path, and dropping it would re-ID every saved watch-only wallet. Add a test that `getID()` is unchanged across `setDerivationPath`.

screen/wallets/WalletDetails.tsx
```diff
@@ -496,6 +499,55 @@ const WalletDetails: React.FC = () => {
...
   };
 
+  // for watch-only HD wallets both values are signer metadata (PSBT bip32Derivation), not key material,
+  // so the user can correct them, e.g. after importing a bare xpub/ypub/zpub which carries neither
+  const isWatchOnlyHd = wallet.type === WatchOnlyWallet.type && wallet.isHd && wallet.isHd();
+
+  const onEditDerivationPathPress = async () => {
+    let newPath: string;
+    try {
+      newPath = await prompt(loc.wallets.details_derivation_path, loc.wallets.details_derivation_path_edit, {
```
beeant: **[bug]** Both edit prompts pass a long help string *and* `defaultValue`. `helpers/prompt.ts` does `const message = defaultValue !== undefined ? '' : text`, so `details_derivation_path_edit` and `details_master_fingerprint_edit` are never shown (the fingerprint prompt at line 529 has the same problem). `handleEditWalletName` already works around this by passing `''` as the message. Users get an unlabeled text field with no format example and no warning that editing the path does not change addresses.

**Suggestion:** Follow `addMultisigStep2` (`loc.multisig.input_fp` / `input_path`): put the current value and the explanation in the message and omit `defaultValue`, or change `prompt()` so message and default can coexist. At minimum, stop passing `defaultValue` if the help text is meant to be visible.

screen/wallets/WalletDetails.tsx
```diff
@@ -496,6 +499,55 @@ const WalletDetails: React.FC = () => {
...
+        type: 'plain-text',
+        defaultValue: derivationPath ?? '',
+      });
+    } catch (_) {
+      return; // cancelled
+    }
+    newPath = newPath.trim().replace(/h/gi, "'");
```
beeant: **[bug]** Path sanitizing only does `trim()` and `h`/`H` → `'`. `ImportCustomDerivationPath` already normalizes iOS smart punctuation (`‘`/`’`) because the default keyboard turns apostrophes into curly quotes. This prompt uses `type: 'plain-text'` → `keyboardType: 'default'`, so a typed `m/84'/0'/0'` can become `m/84’/0’/0’` and fail `/^m(\/\d+'?)*$/` even though it is the path the loc string asked for. Pasted docs/web paths have the same problem.

**Suggestion:** Reuse the same normalization as `ImportCustomDerivationPath.handlePathChange` (and ideally `validateBip32`) before the regex. Accept `h`/`H` as you already do.

tests/unit/watch-only-wallet.test.js
```diff
@@ -911,4 +911,56 @@ describe('BC-UR', () => {
...
+    // a bare zpub carries no key origin info; the user corrects it,
+    // e.g. for an Electrum-seed wallet whose account root is m/0'
+    w.setDerivationPath("m/0'");
+    w.setMasterFingerprintFromHex('73c5da0a');
+
+    assert.strictEqual(w.getDerivationPath(), "m/0'");
+    assert.strictEqual(w.getMasterFingerprintHex(), '73c5da0a');
+
+    // metadata edits must survive re-init (app restart) and never change address derivation
+    w.init();
```
beeant: **[suggestion]** The new test says metadata “must survive re-init (app restart)” but only calls `w.init()` on the live object. That path copies properties from the existing `_hdWalletInstance`, so it cannot catch a serialize/`fromJson`/`init()` failure. There is also no assertion that `getID()` is stable, no PSBT `bip32Derivation` check after a fingerprint edit (path-in-PSBT is only covered by the older ypub test), and pinning is only exercised for a zpub/bech32 wallet — not xpub/legacy (`?? 'p2pkh'`), ypub/p2sh-segwit, or a taproot descriptor whose `segwitType` is already `p2tr`. The three new tests are also nested under `describe('BC-UR')` (starts at line 715) instead of `describe('Watch only wallet')`.

**Suggestion:** Round-trip with `WatchOnlyWallet.fromJson(JSON.stringify(w)); w2.init();` (or `BlueApp.saveToDisk`/`loadFromDisk`), assert `getID()` and addresses, and cover the other HD script types plus a PSBT fingerprint/path after both edits. Move the cases out of the BC-UR block.

screen/wallets/ImportCustomDerivationPath.tsx
```diff
@@ -61,14 +62,24 @@ const ImportCustomDerivationPath: React.FC = () => {
...
-        wallet.setSecret(importText);
-        if (password) {
-          wallet.setPassphrase(password);
+      const watchOnly = new WatchOnlyWallet();
+      watchOnly.setSecret(importText);
+      if (watchOnly.valid() && watchOnly.isHd()) {
+        // watch-only extended key: the script type is determined by the key itself, and the
+        // path is signer metadata (PSBT bip32Derivation), so offer one candidate at the chosen path
+        watchOnly.init();
+        watchOnly.setDerivationPath(newPath);
```
beeant: **[suggestion]** The watch-only branch now creates one candidate and applies `newPath` from screen state, which is still initialized to `m/84'/0'/0'` for every import. For a ypub/xpub the guessed path from `setSecret` is `m/49'/0'/0'` / `m/44'/0'/0'`. The custom-path button is newly shown for those keys (`ImportWalletDiscovery` `isWatchOnlyHd`), so accepting the default bakes a BIP84 origin into a BIP49/BIP44 key’s PSBTs — worse metadata than not using this screen.

**Suggestion:** After `watchOnly.setSecret`/`init()`, seed the input from `watchOnly.getDerivationPath()` so the default is the key’s current path and the user only changes it when they mean to.

screen/wallets/WalletDetails.tsx
```diff
@@ -496,6 +499,55 @@ const WalletDetails: React.FC = () => {
     setIsMasterFingerPrintVisible(true);
   };
 
+  // for watch-only HD wallets both values are signer metadata (PSBT bip32Derivation), not key material,
```
beeant: **[nit]** Comments at `WalletDetails.tsx:502`, `watch-only-wallet.ts:340`, `watch-only-wallet.ts:343`, and `ImportCustomDerivationPath.tsx:68` retell the PR (signer metadata, why one candidate, why `'p2pkh'`). They describe what the code does rather than a non-obvious constraint.

**Suggestion:** Drop the narrative. Keep at most a short note on why `segwitType` must be pinned before `init()` rereads the path.

class/wallets/watch-only-wallet.ts
beeant: **[bug]** Saving the new path onto `this._derivationPath` also changes the wallet ID. `getID()` includes that field, so after the user edits the path, screens that look up the wallet by ID can miss it. Send can then pick a different wallet. Fingerprint edits are fine; only the path write has this problem.

Edit path, stay in that wallet, tap Send → can hit the bug
Edit path, leave, open the wallet from home, tap Send → no problem

**Suggestion:** Keep the edited path on `_hdWalletInstance` (that already gets saved) and pin `segwitType`. Do not change `this._derivationPath`. Do not remove the path from `getID()` either. That would change IDs for every existing watch-only wallet. Add a test that `getID()` stays the same after `setDerivationPath`.

screen/wallets/WalletDetails.tsx
beeant, line 508: **[bug]** The edit dialogs pass both a help message and a default value. `prompt()` throws away the message whenever `defaultValue` is set, so users never see the explanation or the example. The fingerprint prompt has the same problem. Editing the wallet name already avoids this by passing an empty message.

**Suggestion:** Put the current value in the message text and skip `defaultValue`, like the multisig screens do. Or change `prompt()` so a message and a default can both be shown.

screen/wallets/WalletDetails.tsx
beeant: **[bug]** iOS turns a normal apostrophe (`'`) into a curly quote (`’`). This screen does not convert those back, so a path the user typed correctly can fail the check. The import custom-path screen already handles this.

**Suggestion:** Use the same quote cleanup as `ImportCustomDerivationPath.handlePathChange` before validating.

tests/unit/watch-only-wallet.test.js
beeant: The test says these values survive an app restart, but it only calls `init()` on the same object. That does not prove save and load. It also does not check that the wallet ID stays the same, or that PSBTs get the new fingerprint. The new tests also sit inside the BC-UR block, which is the wrong place.

**Suggestion:** Save the wallet to JSON, load it again, and check path, fingerprint, ID, and addresses. Cover other wallet types too (xpub, ypub, taproot). Move these tests out of the BC-UR block.

screen/wallets/ImportCustomDerivationPath.tsx
beeant, line 87: The custom-path screen always starts at `m/84'/0'/0'`, even for ypub/xpub keys that should be `m/49'/0'/0'` or `m/44'/0'/0'`. If the user taps OK without changing it, PSBTs get the wrong path.

**Suggestion:** After `setSecret` / `init()`, fill the input with `watchOnly.getDerivationPath()` so the starting value matches the key.

screen/wallets/WalletDetails.tsx
beeant: Several new comments retell the PR instead of explaining a non-obvious constraint.

**Suggestion:** Delete the extra comments. Keep a short note on why `segwitType` must be pinned before `init()`.

beeant:
Self review: A few issues on the latest commits. Details are in the inline comments.

**Should fix before merge**

1. Saving a new derivation path also changes the wallet ID (`getID()` includes `_derivationPath`). Other screens can then lose the wallet. Send falls back to the first other spendable on-chain wallet, so it can build a tx from the wrong wallet. Fingerprint edits are fine.
2. The edit prompts never show their help text. `prompt()` hides the message when a default value is set.

**Also**

- iOS curly quotes can fail path validation. The import screen already cleans those up.
- Tests only re-`init()` in memory, so they do not prove save/load. They also sit in the BC-UR block.
- Custom-path import always starts at `m/84'/0'/0'`, which is wrong for ypub/xpub.

marcosrdz (approved):

class/wallets/watch-only-wallet.ts
beeant: Fixed in 264bdddd8. Path edits no longer change the wallet ID.

Same xpub imported twice with different custom paths will share an ID. That already happens if you import the same xpub twice.

The Send fallback (use another wallet if the ID is missing) is separate: #8879

screen/wallets/WalletDetails.tsx
beeant: Fixed in 81ea83486. `prompt()` no longer drops the message when a default value is set.

Path and fingerprint dialogs now show help and a pre-filled value. Rename-wallet and tx-note pass an empty message, so they are unchanged.

screen/wallets/WalletDetails.tsx
beeant: Fixed in f5791de5f. iOS curly quotes (`‘` / `’`) are converted to `'` before the path is checked, same as the import custom-path screen.

tests/unit/watch-only-wallet.test.js
beeant: Addressed in fa56451ac.

Tests are under `Watch only wallet`, not BC-UR. They cover getID stability, `fromJson` restart, xpub/ypub script type after a hostile path, and that the edited path and fingerprint land in the PSBT `bip32Derivation`.

screen/wallets/ImportCustomDerivationPath.tsx
beeant: Fixed in a53c9dd6d. The field now starts at the path from the key: xpub `m/44'`, ypub `m/49'`, zpub `m/84'`. Seed import is still `m/84'`.

Script type still comes from the key. If someone types `m/84'` on a bare xpub, they still get `1…` addresses.

screen/wallets/WalletDetails.tsx
beeant: Done in 779bf16ae. Dropped the extra comments. Left the pin as defensive and the note that `getID()` hashes `this._derivationPath`.

beeant:
https://github.com/BlueWallet/BlueWallet/pull/8861#issuecomment-5467196759

First self-review items are fixed in later commits. Threads are left open so other reviewers can still see them.

Send fallback (missing wallet ID uses another wallet) is still #8879.

class/wallets/watch-only-wallet.ts
```diff
@@ -334,9 +334,20 @@ export class WatchOnlyWallet extends LegacyWallet {
...
+  setDerivationPath(path: string) {
+    if (!this.isHd()) throw new Error("Not a HD watch-only wallet, can't use derivation path");
+    if (!this._hdWalletInstance) this.init();
+    if (!this._hdWalletInstance) throw new Error("Not a HD watch-only wallet, can't use derivation path");
+    // defensive: init() can still pick a class from this._derivationPath if segwitType is unset
+    if (!this.segwitType) {
+      this.segwitType = this._hdWalletInstance.segwitType ?? 'p2pkh';
+    }
+    // keep this._derivationPath unchanged: getID() hashes it, so a metadata-only edit would change the wallet ID
+    this._hdWalletInstance.setDerivationPath(path);
```
beeant, line 347: **Export path `m/86'`**

**Reproduce:** Watch-only zpub → set derivation path to `m/86'/0'/0'` → Export/Backup → import that text.

**What happens:** Restores as taproot (`bc1p…`), not the original `bc1q…`. Balance looks empty. `WalletExport.tsx` emits `tr(...)` if the path starts with `m/86`. That file is not in this diff.

**Fix:** Only emit `tr(...)` for a real taproot wallet, not any path that starts with `m/86`.

class/wallets/watch-only-wallet.ts
```diff
@@ -334,9 +334,20 @@ export class WatchOnlyWallet extends LegacyWallet {
...
+  }
+
+  setMasterFingerprintFromHex(hexValue: string) {
+    this.masterFingerprint = this.getMasterFingerprintFromHex(hexValue);
```
beeant, line 351: **Fingerprint ending in `00`**

**Reproduce:** Watch-only zpub → Wallet details → Master fingerprint → type `beebee00` → save. Create a PSBT / Send.

**What happens:** Throws `Invalid Hex string length`. The row may show garbage (`eundefined…`). About 1 in 256 fingerprints. This PR is the first UI that can set it.

**Fix:** `padStart(8, '0')` in `getMasterFingerprintHex` and in `createTransaction` (`abstract-hd-electrum-wallet.ts`). Those pad lines are not in this diff.

screen/wallets/ImportCustomDerivationPath.tsx
```diff
@@ -61,14 +71,22 @@ const ImportCustomDerivationPath: React.FC = () => {
...
       const newWallets: { [type: string]: TWallet } = {};
-      for (const Wallet of [HDLegacyP2PKHWallet, HDSegwitP2SHWallet, HDSegwitBech32Wallet, HDTaprootWallet]) {
-        const wallet = new Wallet();
-        wallet.setSecret(importText);
-        if (password) {
-          wallet.setPassphrase(password);
+      const watchOnly = new WatchOnlyWallet();
+      watchOnly.setSecret(importText);
+      if (watchOnly.valid() && watchOnly.isHd()) {
+        watchOnly.init();
```
beeant, line 86: **Bare xpub + `m/84'` still legacy**

**Reproduce:** Import a bare xpub → custom derivation path → type `m/84'/0'/0'` (not the default) → import. Check addresses.

**What happens:** You still get `1…` addresses. Script type comes from the key (`init()` before `setDerivationPath`). No warning. Tapping OK on the default is now safe; typing BIP84 on an xpub is not.

**Fix:** Warn, or pick script type from the path on this screen only (fresh import, no coins yet).

tests/unit/watch-only-wallet.test.js
```diff
@@ -710,6 +712,125 @@ describe('Watch only wallet', () => {
...
+      [{ address: 'bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq', value: 5000 }],
+      1,
+      w._getInternalAddressByIndex(0),
+    );
+    assert.strictEqual(psbt.data.inputs[0].bip32Derivation[0].path, "m/0'/0/0");
+    assert.strictEqual(uint8ArrayToHex(psbt.data.inputs[0].bip32Derivation[0].masterFingerprint), '73c5da0a');
+
+    // metadata edits must survive re-init and never change address derivation or wallet ID
+    w.init();
+    w._hdWalletInstance.external_addresses_cache = {};
```
beeant: **Address cache**

Without this clear, `_getExternalAddressByIndex(0)` can return a cached address and the test still passes if script type flipped.

We already clear the cache here. Keep it.

beeant:
## Second self-review

New items. Not in the first self-review.

1. **Fingerprint ending in `00`** — cannot create a PSBT (about 1 in 256).
2. **Path starts with `m/86`** — export then import restores a different wallet.
3. **Bare xpub + typed `m/84'`** — still gets legacy `1…` addresses, no warning.

GladosBlueWallet:

image

GladosBlueWallet:

image

[all PRs for @ojokne] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/ojokne

class/wallets/watch-only-wallet.ts
beeant: Fixed in 93daf52ca. The old code only padded one missing zero, but a fingerprint ending in 00 loses more than one when converted to hex. Result: broken display and createTransaction throwing. Now it pads to the full 8 chars with padStart, and there's a test for the worst case (beeb0000).

class/wallets/watch-only-wallet.ts
beeant: Fixed in a60b48d25. Export now checks if the wallet is actually taproot instead of looking at the path. Moved the logic into getSecretForExport() so it's unit tested. Tests cover both directions.

beeant:
Also tightened path validation in 8cd61d489. Paths without m/ or with huge indexes used to pass but produced broken PSBT metadata. Both screens now share one validator.

screen/wallets/ImportCustomDerivationPath.tsx
beeant: Fixed in 01671431e. A bare xpub imported at m/84'/0'/0', m/49'/0'/0' or m/86'... now gets the script type the path declares (the key is wrapped in the key-origin form so setSecret's existing mapping applies). Keys that already declare their type, key origins, and invalid input import unchanged. Short and testnet paths still import as legacy on purpose, tracked in #8881.

One related note: importing the same key at two different account paths still creates two wallets watching the same addresses. Pre-existing and possibly by design, written down in https://github.com/BlueWallet/BlueWallet/issues/8882.

beeant:
https://github.com/BlueWallet/BlueWallet/pull/8861#issuecomment-5467662055

Second self-review items are fixed. Threads are left open so other reviewers can still see them.

- Fingerprint ending in 00: 93daf52ca
- m/86 export: a60b48d25
- Bare xpub script type: 01671431e

Follow-ups:

- Short/testnet paths still import as legacy on purpose: #8881
- Possible duplicates from different account paths, maybe by design: #8882

Overtorment:
re-pushed to  https://github.com/BlueWallet/BlueWallet/pull/8877 to run the test suite

also added `testflight` label to actually test the build

screen/wallets/ImportWalletDiscovery.tsx
```diff
@@ -53,6 +53,12 @@ const ImportWalletDiscovery: React.FC = () => {
     return hd.validateMnemonic();
   }, [importText]);
 
+  const isWatchOnlyHd = useMemo(() => {
+    const wallet = new WatchOnlyWallet();
+    wallet.setSecret(importText);
```
beeant: **Crash**

**Reproduce:** On import, paste `[73c5da0g/84'/0'/0']` plus a real xpub.

**What happens:** `setSecret` throws. No try/catch, so this screen dies.

**Fix:** try/catch. Treat a throw as “not a watch-only HD wallet.” Same on the custom-path screen.

screen/wallets/ImportCustomDerivationPath.tsx
```diff
@@ -45,7 +46,16 @@ const ImportCustomDerivationPath: React.FC = () => {
   const { colors } = useTheme();
   const { importText, password } = useRoute<RouteProps>().params;
   const { addAndSaveWallet } = useStorage();
-  const [path, setPath] = useState<string>("m/84'/0'/0'");
+  const watchOnlyImport = useMemo(() => {
+    const wallet = new WatchOnlyWallet();
+    wallet.setSecret(importText);
```
beeant: Same crash: `setSecret` during render, no try/catch.

screen/wallets/ImportCustomDerivationPath.tsx
```diff
@@ -61,14 +71,23 @@ const ImportCustomDerivationPath: React.FC = () => {
...
       // create wallets
       const newWallets: { [type: string]: TWallet } = {};
-      for (const Wallet of [HDLegacyP2PKHWallet, HDSegwitP2SHWallet, HDSegwitBech32Wallet, HDTaprootWallet]) {
-        const wallet = new Wallet();
-        wallet.setSecret(importText);
-        if (password) {
-          wallet.setPassphrase(password);
+      const watchOnly = new WatchOnlyWallet();
+      // a bare xpub at a script-typed path gets that script type; see setSecretForCustomPathImport
+      watchOnly.setSecretForCustomPathImport(importText, newPath);
```
beeant: **Path without `m/`**

**Reproduce:** Custom-path import. Type `84'/0'/0'` (no `m`). Import. Make a PSBT.

**What happens:** Import accepts it. The PSBT drops `84'`. The signer refuses.

**Fix:** Require `m/` here, like Wallet Details already does.

screen/wallets/WalletDetails.tsx
```diff
@@ -496,6 +499,53 @@ const WalletDetails: React.FC = () => {
...
+    try {
+      newPath = await prompt(loc.wallets.details_derivation_path, loc.wallets.details_derivation_path_edit, {
+        type: 'plain-text',
+        defaultValue: derivationPath ?? '',
+      });
+    } catch (_) {
+      return; // cancelled
+    }
+    newPath = newPath.trim().split('‘').join("'").split('’').join("'").replace(/h/gi, "'");
+    if (!/^m(\/\d+'?)*$/.test(newPath)) {
```
beeant: This screen already requires `m/`. A path saved as `84'/0'/0'` on import cannot be corrected here.

beeant (commented):
## Third self-review

**Should fix**

1. **Crash** — Import calls `setSecret` while rendering. A bad fingerprint like `73c5da0g` inside `[…]xpub` kills the screen.
2. **Path without `m/`** — Import allows `84'/0'/0'`. The PSBT drops `84'`, so the signer refuses. Wallet Details needs `m/…`, so you cannot fix it there later.

**Already fixed**

A zpub wallet (`bc1q…`) whose *path* was edited to `m/86'` used to export as taproot. Restoring that backup showed `bc1p…` and an empty balance. **b6aefc835** exports taproot only for real taproot wallets, not because the path says `m/86'`.

**Not a bug**

A bare xpub is only converted for mainnet `m/84'/0'` and `m/49'/0'`. Testnet `m/84'/1'` stays an xpub on purpose, or the backup would re-import as `1…`. See #8881.

**Nice-to-have**

`getSecretForExport` swallows errors and can return a bare xpub. Better not to.

screen/wallets/ImportWalletDiscovery.tsx
beeant: Fixed in 53e81b876. `setSecret` is in try/catch on Discovery, the custom-path memo, and the debounced create. A throw is treated as not watch-only HD (seed rows / not found), not a crash.

screen/wallets/ImportCustomDerivationPath.tsx
beeant: Fixed in 53e81b876. `setSecret` is in try/catch on Discovery, the custom-path memo, and the debounced create. A throw is treated as not watch-only HD (seed rows / not found), not a crash.

screen/wallets/ImportCustomDerivationPath.tsx
beeant: Fixed in b71fed3fd. Custom-path import now requires `m/`. `84'/0'/0'` is rejected so the PSBT cannot drop the first path level.

screen/wallets/WalletDetails.tsx
beeant: Import no longer saves a path without `m/` (b71fed3fd). This screen already requires `m/`, so new wallets match. To fix an older one, type `m/84'/0'/0'` here.

beeant:
`xpub.tsx` is not in this PR. That screen is seed taproot, not watch-only.

Watch-only export used to build `tr()` from a path starting with `m/86'`. After a path edit that exported the wrong wallet. Export now checks that the wallet is actually taproot and passes `'p2tr'` (`getSecretForExport`).

`getDescriptor` still accepts an `m/86'` path so the seed xpub screen works as before. 8767bf617

beeant:
https://github.com/BlueWallet/BlueWallet/pull/8861#pullrequestreview-5066566632

Third self-review items are fixed. Threads are left open so other reviewers can still see them.

- Crash on bad key-origin hex: 53e81b876
- Custom-path import requires m/: b71fed3fd

Leftovers (path `m`, export catch) are in the fourth self-review.

screen/wallets/WalletDetails.tsx
beeant: **[bug]** Path `m`

**Reproduce:** Watch-only taproot. Save path `m`. Export. Import that text.

**What happens:** Backup is `tr([fp/m]xpub…)`. Re-import path is `m/m`. Import already rejects `m`.

**Fix:** Use `validateBip32`. Keep the `h` → `'` cleanup.

class/wallets/watch-only-wallet.ts
```diff
@@ -334,9 +335,52 @@ export class WatchOnlyWallet extends LegacyWallet {
...
+  // signer metadata: a path starting with m/86 does not mean the wallet is taproot
+  // (and vice versa).
+  getSecretForExport(): string {
+    try {
+      if (this.segwitType === 'p2tr' || this._hdWalletInstance instanceof HDTaprootWallet) {
+        const path = this.getDerivationPath();
+        if (path) return WalletDescriptor.getDescriptor(this.getMasterFingerprintHex(), path, this.getSecret(), 'p2tr');
+      }
+    } catch (_) {}
+    return this.getSecret();
```
beeant, line 376: **[suggestion]** zpub/ypub Export does not keep path and fingerprint

**Reproduce:** Edit a zpub wallet's path and fingerprint. Export. Import.

**What happens:** Restores the bare zpub. Path and fingerprint are gone. Taproot already exports a `tr()` descriptor.

**Fix:** Export as a descriptor, or say in the help that Export will not keep them.

beeant (commented):
## Fourth self-review

Path/fingerprint edits do not change addresses, script type, or wallet ID. PSBT dest/change still come from the xpub.

**Fixed**

- Path `m`: 3526b9bc7
- Import `h`, and Export help for zpub/ypub/xpub: 22707af2e. Descriptor export is #8884.
- Comments: a7eaffd76. Dropped the long export/wrap comments that retold the PR (same as 779bf16ae). Left the getID note: path edits must not change `this._derivationPath`.

**Not a bug**

- No chevron on path/fingerprint. Matches Advanced rows and Show XPUB.
- `m/86'` wrap vs 84/49 is on purpose. #8881

screen/wallets/WalletDetails.tsx
beeant: Fixed in 3526b9bc7. Wallet Details now uses `validateBip32`, same as custom-path import. Path `m` is rejected. The `h` → `'` cleanup is unchanged.

class/wallets/watch-only-wallet.ts
beeant: Addressed in 22707af2e. The edit prompts now say Export/Backup of a zpub, ypub, or xpub will not keep the path or fingerprint. Descriptor export is #8884.

Same commit: custom-path import converts `h` to `'`, like Wallet Details.

beeant:
https://github.com/BlueWallet/BlueWallet/pull/8861#pullrequestreview-5067362185

Fourth self-review items are fixed. Threads are left open so other reviewers can still see them.

- Path `m`: 3526b9bc7
- Import `h` and zpub Export help: 22707af2e (descriptor export is #8884)
- Comments: a7eaffd76. Dropped the long export/wrap comments that retold the PR (same as 779bf16ae). Left the getID note: path edits must not change `this._derivationPath`.

beeant:
Thank you for the previous reviews.

Sorry for the long self-review (including extra AI-assisted passes). I treated this PR carefully because BlueWallet is used by many people.

I finished another pass on the spend path (destination, change, and PSBT origin) and I cannot find anything else to fix. Current HEAD is `a7eaffd`. Please re-review when you can.

Thank you.

cc @Overtorment @marcosrdz @GladosBlueWallet @limpbrains @ojokne @Ugarba202

Overtorment:
https://github.com/BlueWallet/BlueWallet/pull/8885

GladosBlueWallet:

image

GladosBlueWallet:

image

GladosBlueWallet:

image

GladosBlueWallet:

image

[all PRs for @ojokne] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/ojokne

GladosBlueWallet:

image

GladosBlueWallet:

image

[all PRs for @Ugarba202] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/Ugarba202

GladosBlueWallet (approved):
**ADD feature:** editable derivation path and master fingerprint for HD watch-only wallets (PSBT signer metadata), plus custom-path import for bare xpub/ypub/zpub. Commits follow CONTRIBUTING. After four self-review rounds the scary failure modes—wallet ID drift, fingerprint `00` truncation, export/script-type desync—appear addressed; unit tests exercise real wallet/PSBT behavior, not mock theater. The chamber found nothing worth filing.

GladosBlueWallet:

image

GladosBlueWallet:

image

GladosBlueWallet:

image

GladosBlueWallet:

image

[all PRs for @ojokne] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/ojokne

GladosBlueWallet:

image

[all PRs for @Ugarba202] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/Ugarba202

GladosBlueWallet:

image

GladosBlueWallet:

image

GladosBlueWallet:

image

GladosBlueWallet:

image

[all PRs for @ojokne] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/ojokne

GladosBlueWallet:

image

[all PRs for @Ugarba202] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/Ugarba202

GladosBlueWallet:

image

GladosBlueWallet:

image

GladosBlueWallet:

image

GladosBlueWallet:

image

[all PRs for @ojokne] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/ojokne

GladosBlueWallet:

image

[all PRs for @Ugarba202] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/Ugarba202

Overtorment:
i have just tested it on a real device. changed FP a few times, and checked whether hw wallet will recognise it with wrong/right FP or not - worked as expected!

beeant:
Thank you for testing on real device! Glad that it worked as expected!

Overtorment (approved):
lgtm

class/wallet-import.ts
```diff
@@ -24,7 +24,8 @@ import bip39WalletFormatsBlueWallet from './bip39_wallet_formats_bluewallet.json
 import type { TWallet } from './wallets/types';
 
 // https://github.com/bitcoinjs/bip32/blob/master/ts-src/bip32.ts#L43
-export const validateBip32 = (path: string) => path.match(/^(m\/)?(\d+'?\/)*\d+'?$/) !== null;
+// require m/ so bip174 does not drop the first path level (it treats index 0 as m)
+export const validateBip32 = (path: string) => path.match(/^m\/(\d+'?\/)*\d+'?$/) !== null;
```
ojokne, line 28: I have noticed that this function returns false if the bip32 path uses `h` or `H` notation for hardened levels, any specific reason for this decision?

I have tested with `m/44h/0h/0h` and `m/44H/0H/0H` both cases return false but i think they are valid bip32 paths

Overtorment: good catch! i remember people explicitly asking me to support `h` in paths

Overtorment: we can probably merge and fix in a follow up pr

ojokne: no problem then

beeant: Good catch. Prefer I add `h`/`H` here, or in a follow-up with the extra path tests?

tests/unit/wallet-import.test.ts
```diff
@@ -0,0 +1,15 @@
+import assert from 'assert';
+
+import { validateBip32 } from '../../class/wallet-import';
+
+describe('validateBip32', () => {
+  it('requires an m/ prefix so bip174 does not drop the first path level', () => {
```
ojokne, line 6: Non-blocking: the function also accepts `m/84/0'/0'` and `m/84/0/0`. I think these could be useful additions to the test cases to cover mixed and non-hardened paths. What do you think?

beeant: Yes, those are valid. We already cover `m/0/0`.

Should I add the mixed case and `h`/`H` in a follow-up, or in this PR?

ojokne: Yaa that can be added in follow up pr

class/wallet-import.ts
Overtorment: follow up

Overtorment:
@ojokne if you want a build - there is a build in https://github.com/BlueWallet/BlueWallet/pull/8885 (`Build 8.0.2 (1788288097) ` )

ojokne:
> @ojokne if you want a build - there is a build in #8885 (`Build 8.0.2 (1788288097) ` )

Thanks, I want to do some final testing on physical device

ojokne (approved):
Tested with  Jade Hardware wallet, was able to edit the fingerprint and derivation path

- Hardware wallet was able to sign when  correct fingerprint was used

- Hardware wallet could not sign when wrong fingerprint was used

- Hardware wallet also failed to sign when a wrong derivation path was used

LGTM

Merging this should close my pr #8512

GladosBlueWallet:

image

GladosBlueWallet:

image

GladosBlueWallet:

image

[all PRs for @Ugarba202] https://github.com/BlueWallet/BlueWallet/pulls/review-requested/Ugarba202

limpbrains (approved):

Overtorment:
merged. thanks everyone!

PR #8872, opened 2026-08-27
FIX: Receive screen has two close icons on Android (#8856)

### LINKED ISSUE #8856: receive screen has two close icons on a multisig wallet
Please provide:

* your phone model and OS version
Samsung S22 Android 16

* BlueWallet app version (settings->about->scroll down)
8.0.1

* self-test passes? Open settings->about->scroll down, tap "Run self-test"
Yes

See screenshot.

Image

mohamedadel96e:
Closes #8856 

## Problem

On Android, the Receive screen shows two close buttons, one on the left and one on the right.

This was reported on a Samsung S22 with a multisig wallet, but the issue is not multisig-specific. It can happen with any wallet.

## Root cause

`navigationStyle` handles a left close button differently on Android. Instead of rendering a `headerLeft` button, it changes the native back button icon using `headerBackImageSource`.

`createReceiveDetailsOptions` uses a left close button. On the ReceiveDetails path without the BIP47 menu, we then add a close button on the right by overriding `headerLeft` and `unstable_headerLeftItems`.

That works on iOS because the original close button is a `headerLeft` element and gets replaced. On Android, the original close button is actually the native back button, so it remains visible and we end up with two close icons.

## Fix

Hide the native back button on this path:

```
headerBackVisible: false,
```

This leaves only the intended close button on the right.

## Scope

Although the issue was reported with a multisig wallet, it is not related to wallet type. The affected path is selected by the `allowBIP47` route parameter, which `ReceiveDetails` does not set. As a result, navigating to `ReceiveDetails` from any wallet can reproduce the issue.

## Testing

- `npm run tslint` . clean
    
- `eslint navigation/DetailViewScreensStack.tsx` . clean
    
- `npm run unit` . 521 tests passed across 54 suites
    
- Manually verified on Android 16 / API 36
    

The Receive screen now shows only one close button on the right, and the native "Navigate up" button is no longer present in the accessibility tree.

### Before 
638998763-f84530d8-4b09-46af-b78d-f35a7f130213

### After 
Screenshot from 2026-08-27 12-14-01


GladosBlueWallet (approved):
**Bug fix:** Android Receive showed twin close icons because `navigationStyle` paints the left close as a restyled native back button (`headerBackImageSource`), while `createReceiveDetailsOptions` also mounts a right close — overriding `headerLeft` never evicts the ghost on the left. **Verdict:** One line of `headerBackVisible: false` at `navigation/DetailViewScreensStack.tsx:706` is the correct, minimal antidote; commit/PR text match CONTRIBUTING; no new test theater. I ran the review chamber. The subject survived. How disappointing.

Ugarba202 (approved):
Tested and verified the navigation options logic. 
Because `navigationStyle` maps left close buttons to `headerBackImageSource` on Android, setting `headerBackVisible: false` is indeed the right fix to prevent the native back icon from lingering when the right close button is mounted.
Clean and minimal fix. LGTM 👍

ojokne:
I have tested the master branch as of commit 106185382b5b62e2af313b4d9bf1602cf2dc6c1c and this issue doesnt seem to exist 

Screenshot_1789392321

perplexiphone-cmd (approved):
appr

mohamedadel96e:
I've rebased the branch on `master` branch because it had conflicts could you re review it please. @GladosBlueWallet

PR #8877, opened 2026-08-29
Beeant add watchonly metadata editing

Overtorment:


beeant:
Looks like emulator flakiness.
The failures are Detox visibility timeouts + `adb: error: listener not found`, and the parallel test run passed. The restore-receive-address test failed on a UI timeout, not an address mismatch, so the segwit pinning's still fine.

PR #8883, opened 2026-08-31
FIX: Require push consent for custom Electrum

### LINKED ISSUE #4806: When user uses custom electrum server, make him aware that his addresses/txids sre still leaked to a Push Notificatinos serer
@ncoelho i think we need your brilliance in making this work

perplexiphone-cmd:
Users who configure a preferred Electrum server should not have wallet addresses or transaction IDs sent to the push service by default.

This adds a default-deny consent gate at the shared Ground Control registration path, covering every existing address and transaction-ID registration caller. The Electrum settings screen now explains the separate push-service disclosure and requires a separate, explicit confirmation; changing the preferred server revokes that consent.

Focused regression coverage verifies that custom Electrum configurations do not register data until consent is recorded, while the default Electrum path continues to register normally.

Validation note: `git diff --check` passed. Jest and TypeScript could not run because `npm ci --ignore-scripts` failed while materializing git dependencies on Windows, leaving no local Jest or TypeScript binary.

Fixes #4806


tests/unit/notifications.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    }),
+    __storage: storage,
+  };
+});
+
+const isCustomElectrumServerConfiguredMock = isCustomElectrumServerConfigured as jest.Mock;
+const fetchMock = fetch as jest.Mock;
+const notificationsStorage = jest.requireMock('@react-native-async-storage/async-storage').__storage as Map<string, string>;
+
+describe('Ground Control registration privacy', () => {
```
GladosBlueWallet, line 43: **[HIGH]** Three fetch-count parlor tricks, zero edge cases. No thrown preference read (your documented fail-closed), no NOTIFICATIONS_NO_AND_DONT_ASK interplay, and `isCustomElectrumServerConfigured` never tested—only mocked into obedience. CONTRIBUTING wanted edge cases; you brought props.

blue_modules/notifications.ts
```diff
@@ -215,6 +217,14 @@ export const cleanUserOptOutFlag = async () => {
   return AsyncStorage.removeItem(NOTIFICATIONS_NO_AND_DONT_ASK_FLAG);
 };
 
+export const isCustomElectrumPushNotificationsEnabled = async (): Promise<boolean> => {
+  return (await AsyncStorage.getItem(CUSTOM_ELECTRUM_PUSH_NOTIFICATIONS_OPT_IN)) === 'true';
+};
+
+export const setCustomElectrumPushNotificationsEnabled = async (enabled: boolean): Promise<void> => {
```
GladosBlueWallet: **[MEDIUM]** Flipping consent off only writes AsyncStorage. Ground Control keeps every address you already fed it—privacy theater. Where's `unsubscribe`? Still in another room, apparently.

blue_modules/notifications.ts
```diff
@@ -302,6 +312,11 @@ export const majorTomToGroundControl = async (addresses: string[], hashes: strin
       throw new Error('No addresses, hashes, or txids provided');
     }
 
+    if ((await isCustomElectrumServerConfigured()) && !(await isCustomElectrumPushNotificationsEnabled())) {
```
GladosBlueWallet: **[MEDIUM]** `isCustomElectrumServerConfigured` throws on preference failure; the outer catch rethrows. Half your callers fire-and-forget (`CPFP.js:102`, `Confirm.tsx:230`)—congratulations, unhandled rejections instead of a quiet fail-closed return.

GladosBlueWallet (changes requested):
Privacy **FIX**: block Ground Control from receiving addresses/txids when a preferred Electrum server is set unless the user opts in via Electrum settings. The gate logic is coherent and not bloated, but your test chamber has blind spots, opt-out is cosmetic, and one sibling pipeline still leaks like a sieve.

- `blue_modules/notifications.ts:374` — **[MEDIUM]** Comment promises the same gates as `majorTomToGroundControl`. `registerArkPaymentPush` skips the custom-Electrum check entirely—swap metadata still phones home. Pick one privacy policy.
- `screen/settings/ElectrumSettings.tsx:322` — **[LOW]** Reset-to-default clears Electrum prefs but leaves `CUSTOM_ELECTRUM_PUSH_NOTIFICATIONS_OPT_IN` rotting in AsyncStorage. Stale consent until the next save scrubs it—sloppy state hygiene.

blue_modules/notifications.ts
perplexiphone-cmd: Consent revocation now marks sharing disabled before unsubscribing all recorded Ground Control registrations plus the wallet's current addresses and transaction IDs. Local records remain for retry if unsubscription fails.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

blue_modules/notifications.ts
perplexiphone-cmd: The custom-server check is now contained in a fail-closed helper. Preference-read failures log the error and block registration without rejecting fire-and-forget callers.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

tests/unit/notifications.test.ts
perplexiphone-cmd: Added coverage for notification opt-out, failed preferred-server reads (fail closed without rejection), and the real `isCustomElectrumServerConfigured` implementation.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

perplexiphone-cmd:
> `registerArkPaymentPush` skips the custom-Electrum check entirely.\n\nArk payment push registration now uses the same custom-Electrum consent gate.\n\n> Reset-to-default leaves custom consent in AsyncStorage.\n\nReset-to-default now clears the custom-Electrum consent state.

perplexiphone-cmd (commented):
Request changes

perplexiphone-cmd (commented):
Approve

PR #8885, opened 2026-08-31
Beeant add watchonly metadata editing v2

Overtorment:


Overtorment:
i have just tested it on a real device. changed FP a few times, and checked whether hw wallet will recognise it with wrong/right FP or not - worked as expected!

PR #8886, opened 2026-08-31
FIX: misaligned icon on receive-onchain screen

Overtorment:
before: 

image

after:

image

### DIFF
--- a/components/TransactionPendingIconBig.tsx
+++ b/components/TransactionPendingIconBig.tsx
@@ -17,7 +17,7 @@ export const TransactionPendingIconBig: React.FC = () => {
     <View>
       <View style={styles.boxIncoming}>
         <View style={[styles.ball, hookStyles.ball]}>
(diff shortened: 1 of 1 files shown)


marcosrdz (approved):

GladosBlueWallet (approved):
Single-commit **FIX** for a visual alignment bug in `TransactionPendingIconBig`: the pending-state icon on receive (`ReceiveDetails`) was nudged with `iconStyle` magic numbers (`left: 0`, `top: 25`); the diff drops that hack and centers via `justifyContent` / `alignItems` on the 150×150 circle—same pattern as `TransactionPendingIcon.tsx`. Commit prefix matches CONTRIBUTING; change is minimal, no tests (reasonable for layout-only). I cannot verify the mandated UI screenshot from the repo alone. The subject survived inspection. Disappointing, but functional.

PR #8888, opened 2026-08-31
OPS: bump @arkade-os/sdk to 0.4.67 and @arkade-os/boltz-swap to 0.3.67

GladosBlueWallet:
## Summary
- Bump `@arkade-os/sdk` from 0.4.56 to 0.4.67 and `@arkade-os/boltz-swap` from 0.3.61 to 0.3.67.
- Add an npm override so `@arkade-os/sdk` stays at 0.4.67. Published `boltz-swap@0.3.67` pins sdk 0.4.62, and two copies fail TypeScript (`Wallet` vs `IWallet`).

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npm run unit` (55 suites, 525 passed, 1 skipped)
- [ ] Open an Arkade Lightning wallet and confirm balance, receive, and swap history still load

Made with [Cursor](https://cursor.com)

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,8 +91,8 @@
     "unit": "jest -b tests/unit/*"
   },
   "dependencies": {
-    "@arkade-os/boltz-swap": "0.3.61",
-    "@arkade-os/sdk": "0.4.56",
+    "@arkade-os/boltz-swap": "0.3.67",
+    "@arkade-os/sdk": "0.4.67",
     "@bugsnag/react-native": "8.9.0",
     "@bugsnag/source-maps": "2.3.3",
     "@keystonehq/bc-ur-registry": "0.8.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -190,5 +190,8 @@
     "text-encoding": "0.7.0",
     "url": "0.11.4",
     "wif": "2.0.6"
+  },
+  "overrides": {
+    "@arkade-os/sdk": "0.4.67"
```
pietro909, line 195: Do we need the override?

pietro909 (approved):
Tested on Android device

package.json
Overtorment: apparently, boltz-swap wants older ver

PR #8901, opened 2026-09-07
FIX: accept h/H and smart-quote hardened markers in derivation paths (follow-up to #8861)

beeant:
## Why

Follow up to #8861. Closes the two review threads left open there:
- `h`/`H` notation: https://github.com/BlueWallet/BlueWallet/pull/8861#discussion_r3927955069
- extra path tests (mixed / non-hardened): https://github.com/BlueWallet/BlueWallet/pull/8861#discussion_r3930507778

@ojokne noticed `validateBip32` rejected `h`/`H` paths like `m/84h/0h/0h`. Fixing that turned up a worse bug in `setSecret`.

`setSecret` only lowercased `h` (`/h/g`). So a BIP84 origin with uppercase `H` or iOS smart quotes, like `[fp/84H/0H/0H]xpub`, kept the `H`, failed the `startsWith("m/84'/0'/")` check, never converted the xpub to zpub, and imported as a legacy `1…` wallet. Wrong addresses for a native segwit key. The main import screen calls `setSecret` on raw text too, so this already happens there, not just on the new screen.

## What changed

- `validateBip32` now accepts `h`/`H` and smart quotes. It normalizes first (`normalizeDerivationPath`: trim, `‘’` and `h`/`H` to `'`, leading `M` to `m`) then runs the old regex. Stored path is always the `'` form, because bitcoinjs `derivePath` rejects `h`. Nice side effect: `84h` / `84H` / `84'` end up as the same wallet id.
- `setSecret` rewrites `h`, `H` and smart quotes on the path (`/[hH‘’]/g`) at the three places it builds one. So those origins convert to zpub/ypub like they should, instead of staying legacy.
- Both watch-only screens normalize the path before saving, so a raw `h`/`H`/`‘’` path never gets stored.

## Behavior

`'`, `h`, `H`, `‘`, `’` on a BIP84 origin all give the same `bc1q…` wallet now, same fingerprint, path `m/84'/0'/0'`. Existing wallets are not touched. Load is `fromJson` + `init()`, which never calls `setSecret`.

## Tests

- New: `validateBip32` accepts `h`/`H`, rejects no-`m/` / bare `m` / `//`. `normalizeDerivationPath` canonicalizes. Watch-only import of a BIP84 origin with uppercase `H` and with smart quotes asserts `bc1q…`, not `1…`.
- `jest tests/unit`, integration import test, `tsc` and eslint all pass.

## Not here (follow up)

The same lowercase `/h/g` is still in the multisig cosigner parser and the Cobo `AccountKeyPath` branch. Different import paths, this PR does not touch them. Cobo is low risk anyway (it carries the script type in the zpub, so `h`/`H` is only cosmetic there). Worth a separate PR.

### DIFF
--- a/class/wallet-import.ts
+++ b/class/wallet-import.ts
@@ -23,9 +23,20 @@ import bip39WalletFormatsElectrum from './bip39_wallet_formats.json'; // https:/
 import bip39WalletFormatsBlueWallet from './bip39_wallet_formats_bluewallet.json';
 import type { TWallet } from './wallets/types';
 
+// Canonicalize a user-typed derivation path: trim, iOS smart quotes and h/H hardened notation
+// become ', a leading M becomes m. The stored form must use ' because bitcoinjs derivePath
+// rejects h notation (its schema is /^(m\/)?(\d+'?\/)*\d+'?$/).
+export const normalizeDerivationPath = (path: string): string =>
+  path
+    .trim()
+    .replace(/[‘’]/g, "'")
+    .replace(/[hH]/g, "'")
+    .replace(/^M\//, 'm/');
+
 // https://github.com/bitcoinjs/bip32/blob/master/ts-src/bip32.ts#L43
-// require m/ so bip174 does not drop the first path level (it treats index 0 as m)
-export const validateBip32 = (path: string) => path.match(/^m\/(\d+'?\/)*\d+'?$/) !== null;
+// require m/ so bip174 does not drop the first path level (it treats index 0 as m); normalize
+// first so h notation (m/84h/0h/0h) is accepted the same as m/84'/0'/0'
+export const validateBip32 = (path: string) => normalizeDerivationPath(path).match(/^m\/(\d+'?\/)*\d+'?$/) !== null;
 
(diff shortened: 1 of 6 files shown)


GladosBlueWallet (approved):
Subject has repaired BIP84 watch-only imports when derivation paths arrive with `h`/`H` or iOS smart quotes masquerading as hardened markers — a predictable human formatting failure, which I find *almost* endearing.

New `normalizeDerivationPath` runs through validation and UI. `abstract-wallet` origin parsing now accepts `/[hH'']/g` instead of `/h/g`, so `[fp/84H/0H/0H]xpub` correctly triggers xpub→zpub conversion and native segwit addressing. The wallet no longer punishes users for typing like autocorrected test subjects.

Behavioral tests cover normalization and end-to-end import with real addresses. Commit hygiene is acceptable: `FIX:` prefix, focused diff, no gratuitous collateral damage. I have no objections to log. Proceed before someone invents a fifth apostrophe variant.

Overtorment (approved):
lgtm

PR #8903, opened 2026-09-09
FIX: keyboard on pushed screen scrolled previous screens on iOS

limpbrains:
## Problem

On iOS, opening the keyboard on a pushed screen (e.g. Settings → Language → search, or any screen with a text input) left the screens **below** it in the native stack scrolled down after going back: Settings root and the home wallets list lost their top (`WalletsList` carousel y 176 → 60, Settings first row y 308 → 88). Android unaffected. The e2e suite has been papering over this with `scrollUpOnHomeScreen()`.

## Root cause

RN 0.85 Fabric `RCTScrollViewComponentView` subscribes every scroll view that has `automaticallyAdjustKeyboardInsets` to `UIKeyboardWillChangeFrameNotification`, including screens still mounted under the stack top. The handler ends with `scrollTo:`, which clamps `y >= -contentInset.top` using the raw inset rather than `adjustedContentInset`. Lists under a transparent/large header with `contentInsetAdjustmentBehavior="automatic"` rest at `y = -adjustedContentInset.top`, so the clamp forces `y = 0` — scrolled by the header height.

`SafeAreaScrollView`, `SafeAreaSectionList` and `WalletsCarousel` set that prop by default, so the home list and Settings root (no text inputs at all) were reacting to keyboards shown elsewhere.

## Fix

`automaticallyAdjustKeyboardInsets` is now opt-in:
- removed the default from `SafeAreaScrollView`, `SafeAreaSectionList`, `WalletsCarousel` (comment explains why);
- screens with inline inputs that relied on the wrapper default now pass it explicitly: `Broadcast`, `LightningSettings`, `SettingsBlockExplorer`, `generateWord`. The other input screens (`Add`, `ImportWallet`, `ElectrumSettings`, `IsItMyAddress`, `ReceiveDetails`, `CPFP`) already did.

The e2e suite no longer needs `scrollUpOnHomeScreen()` (an iOS-only swipe that compensated for exactly this bug after every return to the home screen), so the helper, its 17 call sites and the imports are removed in a second commit.

## Before / after

Settings → Tools → Is it my address → type into the address field → back ×3.

**Before** (master): home list and Settings end up scrolled after coming back

https://github.com/user-attachments/assets/3249fd30-d8c3-4112-8134-98908547b5a2 

**After** (this PR)

https://github.com/user-attachments/assets/34a8db31-713f-4463-8c3e-1b65a48462c0

## Verification

Verified with a temporary Detox e2e (not part of this PR) that records `frame.y` of the home carousel and a Settings row, opens Settings → Tools → Is it my address, types into `AddressInput`, goes back, and asserts both positions unchanged. Red on `master`, green with the fix.

Also re-ran on iOS 26.3 sim: `all settings screens work`, `can create wallet, reload app and it persists…`, `can discover wallet account and import it` — all green. `npm run lint` and `npm run unit` pass.

After dropping the helper, re-ran the tests that used it: `bluewallet3.spec.js` (watch-only zpub import), `can create wallet and delete wallet`, `can encrypt storage, and decrypt storage works`, `can create wallet, and use main screen SCAN button to scan address` — all green (the last one needed a rerun: the QR backdoor left the scanner open once, unrelated to scrolling).

https://claude.ai/code/session_01Prha4c4pgdDWXtqDfMio7q

### DIFF
--- a/components/SafeAreaScrollView.tsx
+++ b/components/SafeAreaScrollView.tsx
@@ -65,12 +65,15 @@ const SafeAreaScrollView = forwardRef<ScrollView, SafeAreaScrollViewProps>((prop
     <ScrollView
       ref={ref}
       style={componentStyle}
-      // `scrollToOverflowEnabled` works around RN's keyboard handler: it re-applies the current offset via `scrollTo:`,
-      // which clamps to the raw `contentInset` instead of `adjustedContentInset`, so content under a large-title header
-      // jumps up by the header height when the keyboard opens. Upstream fix: https://github.com/react/react-native/pull/56189
+      // `automaticallyAdjustKeyboardInsets` is opt-in: RN's keyboard handler also fires for screens below the
+      // stack top and clamps their offset (RCTScrollViewComponentView `_keyboardWillChangeFrame` / `scrollTo:`),
+      // scrolling the previous screen. Screens with text inputs pass the prop explicitly.
+      // `scrollToOverflowEnabled` works around the same handler for screens that do opt in: it re-applies the current
+      // offset via `scrollTo:`, which clamps to the raw `contentInset` instead of `adjustedContentInset`, so content under
+      // a large-title header jumps up by the header height when the keyboard opens.
+      // Upstream fix: https://github.com/react/react-native/pull/56189
       contentInsetAdjustmentBehavior="automatic"
       scrollToOverflowEnabled
-      automaticallyAdjustKeyboardInsets
       automaticallyAdjustsScrollIndicatorInsets
       contentContainerStyle={contentStyle}
       {...otherProps}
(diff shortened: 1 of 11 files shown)


GladosBlueWallet (approved):
Subject has corrected a keyboard-inset contagion: `automaticallyAdjustKeyboardInsets` is now **opt-in** on `SafeAreaScrollView`, `SafeAreaSectionList`, and `WalletsCarousel`, so iOS background stack scroll views—home wallets list, Settings root—no longer twitch when a pushed screen opens a keyboard. Four `SettingsScrollView` screens that previously inherited the behavior now declare it explicitly: `Broadcast`, `LightningSettings`, `SettingsBlockExplorer`, `generateWord`. Other input screens already did. The iOS-only `scrollUpOnHomeScreen` e2e workaround is deleted. Commits use CONTRIBUTING prefixes (`FIX`, `TST`). Root cause and fix align; no input-bearing scroll screen was left keyboard-blind. Test complete. For once, the facility did not catch fire.

GladosBlueWallet:

image

marcosrdz (approved):

GladosBlueWallet:

PR #8912, opened 2026-09-14
ADD: Receive addrrss Apple intent

marcosrdz:
Let's you automate via Apple Shortcuts the address for any task you want. 

example: 

When X happens get the address for Y wallet so it can do Z (let's say...buy btc via some API and send it to the address your BW provides)

image


class/blue-app.ts
```diff
@@ -216,6 +217,9 @@ export class BlueApp {
 
   encryptStorage = async (password: string): Promise<void> => {
     // assuming the storage is not yet encrypted
+    // A receive address must not remain accessible from Shortcuts once wallet storage is password protected.
+    await clearReceiveAddressShortcutData();
```
GladosBlueWallet, line 231: **[HIGH]** `encryptStorage` now awaits `clearReceiveAddressShortcutData()` with the confidence of a subject who believes the floor is load-bearing. That helper always calls `getOptions()`, which demands `EventEmitter.getReceiveAddressShortcutKeychainAccessGroup`. Android has no EventEmitter. Optional chaining returns undefined; `getOptions` throws. Nobody catches it. Encryption never starts. We added a shortcut cleanup that shortcuts the lock. Science.

hooks/useReceiveAddressShortcut.ios.ts
```diff
@@ -0,0 +1,60 @@
...
+const useReceiveAddressShortcut = (): void => {
+  const { wallets, walletsInitialized } = useStorage();
+  const { isReceiveAddressShortcutEnabled } = useSettings();
+
+  useEffect(() => {
+    if (!walletsInitialized) return;
+
+    syncReceiveAddressShortcutWallets(
+      wallets,
+      isReceiveAddressShortcutEnabled,
```
GladosBlueWallet: **[HIGH]** The iOS sync effect listens only to `isReceiveAddressShortcutEnabled` and never calls `isStorageEncrypted()`. After `encryptStorage` clears the keychain, Settings state stays `true` until General Settings is opened. The password sheet then calls `saveToDisk()`, which replaces `wallets` and retriggers the effect. `setReceiveAddressShortcutData({ enabled: true, wallets })` puts the addresses back. I wiped the data. You put it back. That was not the test. `useDeviceQuickActions` at least checks encryption. This hook prefers optimism.

GladosBlueWallet (changes requested):
Two test subjects failed the enrichment chamber. `encryptStorage` now trips an iOS-only keychain bridge and aborts encryption on Android before any lock is applied. On iOS, the shortcut sync then writes receive addresses back into the shared keychain in the same session, because it never asks `isStorageEncrypted()` and React still thinks the toggle is on. The wipe was adorable. The restore was mandatory.

GladosBlueWallet (approved):
No candidate defects survived verification. The test chamber remains disappointingly intact.

PR #8917, opened 2026-09-15
ADD: PSBT Details in deeplink form sheet

marcosrdz:
## Summary
- show parsed input/output counts and total output amount for deep-linked PSBTs
- list output amounts and selectable script hex before signing
- avoid the no-signing-in-progress alert for a valid deep-linked PSBT

Screenshot iPhone 17 Pro 09-15-2026 at 12 02 18 AM


GladosBlueWallet (approved):
Subject 8917 has been fitted with a **parsed PSBT summary panel** on `PsbtWithHardwareWallet`—input/output counts, per-output amounts, and selectable addresses, with script-hex fallback when an output refuses to decode like a civilized address. The false **“no signing in progress”** alert for deep-linked PSBTs has been corrected; I was beginning to suspect the test subjects were hallucinating. **iOS Continuity handoff** for PSBTs is wired through `subtitleSelectable`, handoff listener/AppDelegate plumbing, emulator script, Podfile deployment-target bump, and fastlane default platform—support infrastructure, not decorative cake. Latest commit restores script-hex display for non-address outputs. **No anomalies flagged.** Proceed to the next chamber.

ojokne:
I have a suggestion, would it be better to display the animated QR code first, and then put the extra details under a collapsible section so users can expand to see the extra details if needed? @ncoelho what do you think?

ncoelho:
Maybe it is better, there is a view raw psbt option. So just have the details on that sheet with the raw psbt (view details). Wdyt @marcosrdz ?

PR #8923, opened 2026-09-16
fix: ios27 crash

ncoelho:
The crash is iOS 27 aborting on launch because BlueWallet never adopted the UIScene lifecycle. UIKit now hard-fails with UIApplicationEvaluateRuntimeIssueForNoSceneLifecycleAdoption for apps built with the iOS 27 SDK that still create a window only from AppDelegate.

That is now wired up:

1. Info.plist declares a scene manifest (UIApplicationSupportsMultipleScenes is false; React Native cannot run multiple scenes).
2. SceneDelegate creates UIWindow(windowScene:) and starts React Native there.
3. AppDelegate still owns process setup, but it no longer creates a non-scene window. Deep links, Handoff, and Home Screen quick actions are forwarded from the scene callbacks.


GladosBlueWallet (approved):
**Test subject: iOS 27 scene lifecycle compliance.** Delightful — another mandatory Apple runtime lecture, and you've actually listened. `Info.plist` declares a single-scene manifest; `AppDelegate` keeps process setup and sets `automaticallyLoadReactNativeWindow = false`; `SceneDelegate` spawns a scene-bound `UIWindow` and invokes `reactNativeFactory.startReactNative`. Deep links, Handoff, and quick actions route through scene callbacks into existing `AppDelegate` handlers; cold-start context merges into `launchOptions` so RN Linking and quick-action bootstrap don't wake up confused. Matches the RN 0.85 `RCTAppDelegate` pattern — the launch abort (`UIApplicationEvaluateRuntimeIssueForNoSceneLifecycleAdoption`) should stop screaming. Minor note: commit prefix is `fix:` where CONTRIBUTING wants `FIX:` — paperwork, not physics. **No implementation defects in the diff.** The cake remains a lie; this patch is not.

marcosrdz (approved):

ojokne (approved):
Tested this and it fixes the crash, was able to build and run successfully on both ios27 and ios26.5

PR #8930, opened 2026-09-18
FIX: underpaid fee with 34-byte outputs, change below dust limit, taproot wallet restore

limpbrains:
Found while reviewing #8905. Depends on BlueWallet/coinselect#1.

## Problem

- `TaprootWallet` had `segwitType = 'p2wpkh'`. Inputs sized 27 vB instead of 17. Overpay ~10 vB per input.
- `coinselect` sizes change as p2pkh (25 bytes). p2tr / p2wsh change is 34. Tx with change underpays 9 vB. `HDTaprootWallet`: 206 sats for 212 vB. Native segwit multisig 1-of-2: 157 for 166 vB.
- `sp1…` target sized as p2pkh, real output is p2tr. Underpays 7-10 vB.
- Uppercase bech32 target (`BC1P…`) not recognized (`startsWith('bc1')`), sized as p2pkh. Taproot send max: 102 sats for 111 vB.
- `loadFromDisk` had no `case TaprootWallet.type`. Wallet restored as `LegacyWallet` after restart.
- Change below Bitcoin Core dust limit: lib made change when remainder > `148 * feeRate` (149 sats at 1 sat/vB). Nodes reject such tx as `dust`. Every wallet type.

Old tests rounded feerate, so missed it.

## Fix

- `coinselect` dep bumped to BlueWallet/coinselect#1: Core dust limits (546 p2pkh / 540 p2sh / 294 p2wpkh / 330 p2wsh, p2tr) for change and send max outputs, new `options` arg.
- `LegacyWallet.getChangeScriptLength()`: 34 for p2tr, `MultisigHDWallet` returns 34 for native segwit. Passed to lib as `changeScript`, so lib sizes change and checks its dust limit itself. Nothing adjusted after selection.
- `TaprootWallet.segwitType = 'p2tr'`. `fromJson` forces it for wallets saved by old versions (removable in 2028). `weOwnAddress` lowercases for p2tr too.
- bech32 prefix check case insensitive. `sp1…` target sized like `bc1p…`.
- `loadFromDisk`: added `case TaprootWallet.type`.
- `typings/coinselect.d.ts`: bogus 4th param `changeAddress?: string` replaced with `options`.

## Behaviour change for all wallets (from lib bump)

- Change below 546 sats (330 for p2tr / p2wsh wallets) goes to fee. Before: tx with such change did not relay. p2wpkh / p2sh wallets still assumed p2pkh change, so their change of 294-545 (relayable) also goes to fee. Exact lengths for them = #8932.
[...]

Issue #4352, opened 2022-01-10
Cannot send anything at 1sat/byte


Transisto:
to reproduce

Select a single UTXO with coin control

Select default fee of 1 sat per byte.

Wallet will produce a transaction of 113 Bytes with a total fee of 112 Sat making it below the minimum relay fee of 1sat/byte.

I might have posted a few similar issue or comments, This has been happening to me regularly.

Transisto:
At least if we could select 1.1 sat that would fix it up without doubling the transaction cost.

Overtorment:
I assume its also a tx with only single output? no change?

and which type, BIP84 (native segwit) ?

sebastian13:
Same. Error message:

_min relay fee not met, 112 < 113_

- type: HD SegWit
- Transaction with single input + single output, no change
- 1 sat/vByte fee

Transisto:
Yes, this

Was segwit to legacy address,

Transferring a single utxo

Transisto:
Same thing happening with latest version.

![Screenshot_20220131-011035_BlueWallet](https://user-images.githubusercontent.com/388716/151747590-ab1aea41-f2fe-48b3-9306-a619d54d9e46.jpg)

1 input 2 output,
1 sat per bytes.

Transisto:
This is so ridiculous, How long can it take to make the lowest fee be 1.1 sat/byte?

Transisto:
This happens with any regular transaction sending to a BECH32 P2WSH address like bc1qtmcfj7lvgjp866w8lytdpap82u7eege58jy52hp4ctk0hsncegyqel8prp

landabaso:
@Overtorment, I'm wondering what if the lengthy address is on the changeAddress. changeAddress is not being compensated (+3 bytes).
It looks like the {script: {length}} trick for coinselect cannot be applied to changeAddresses. It could be added to one of the targets but the thing is we don't event know if changeAddress will be needed. It's tricky.

Transisto:
@landabaso from what I read the +3 is not what's doing the trick but actually looking at the length of the address.

What would be IMO simple is to aim for 1.1 sat/bytes since so many wallets make it easier to select the minimum relay fee and integers the mempool tend to get filled at exactly 1.00sat/bytes.

landabaso:
Thanks @Overtorment. I think I didn't explain myself clearly (either that or there's something I don't undersand).

This is how coinselect counts bytes for the outputs:

[...]

PR #8931, opened 2026-09-18
FIX: underpaid fee for legacy p2sh multisig, segwit 1 vB gap and script targets

limpbrains:
Stacked on #8930 (base = its branch). Same class of problem, different causes. Found by probing all on-chain wallet types, comparing `fee` with real `virtualSize()`.

## Problem

- Legacy P2SH multisig scriptSig formula `2 + M*74 + N*34` misses redeem script push opcode (`OP_PUSHDATA1` when N >= 3) and 3-byte length when scriptSig > 252 bytes. 2-of-3, 3 inputs, send max: 923 sats for 931 vB.
- `coinselect` does not know segwit marker & flag (0.5 vB), vsize rounds up, so exactly-estimated tx is 1 vB short. `SegwitBech32Wallet` send max to p2pkh: 112 for 113 vB. Same as #4352, which was patched with `+3` for `bc1` targets only.
- Script targets (OP_RETURN) sized `hex.length / 2 - 4`. BIP47 notification tx: 32.86 sat/vB at requested 33.

## Fix

- `MultisigHDWallet.coinselect()`: legacy scriptSig = `OP_0 + M*74 + push opcode + redeem script`, +2 if over 252. Segwit formulas untouched.
- `LegacyWallet.coinselect()`: `txExtraBytes: 1` passed to lib for `p2wpkh` / `p2sh(p2wpkh)` wallets when no target is already overestimated (`bc1` / `sp1` have `+3`, p2sh is 23 vs assumed 25). Accounted before selection, so lib's own dust / no-solution checks apply. p2tr not needed (input sized 17 vB, real 16.5). Segwit multisig not needed (input estimate has slack).
- Script target length = `hex.length / 2`. The `- 4` was compensating `+3` fudges of other outputs, and underpaid when there were none.

## Behaviour change

- Exact amount that leaves 1 sat too little for fee now fails with `Not enough balance` instead of making underpaid tx (e.g. 99888 from 100000: 112 sats for 113 vB tx). Max amount from "use full balance" goes through same `coinselect()`, stays valid.
- p2pkh-only targets + change from p2wpkh wallet: overpays <= 1 vB (change presence unknown before selection).
- Tx with script target + bech32 outputs overpays ~5 vB until `+3` is gone (#8932).

## Tests

Each seen failing before fix.

- `multisig-hd-wallet.test.js`: new 2-of-3 legacy test. Master: `fee 923 is below 1 sat/vbyte for 931 vbytes`.
- `segwit-bech32-wallet.test.js`: existing test extended: send max pays >= vsize (master: `fee 112 is below 1 sat/vbyte for 113 vbytes`); 99888 throws, 99887 pays >= vsize.
- `bip47.test.ts`: notification tx feerate >= requested. Before: `feerate 32.86… is below requested 33`.
- 5 fixtures (p2pkh target + change) moved by 1 vB * feeRate: `cosign`, `segwit-p2sh-wallet`, `segwit-bech32-wallet` x2, and the same SegwitP2SH tx hardcoded in in-app `SelfTest.tsx` (run by `tests/integration/App.test.js`). OP_RETURN test upper bound 151 -> 155.
- e2e untouched: all asserted e2e amounts go to `bc1q…` targets, no extra byte there.
- Probe on final state (#8930 + this + new lib), 5772 signed txs, 14 wallet setups x 6 target types x 1/3/8 inputs x change / send max / exact amount / dust window: 0 underpaid, 0 outputs below Core dust limit, max overpay 4.3% on smallest tx.
- `npm run unit` with test mnemonics: 59 suites pass. `npm run integration` locally: 12 of 13 suites pass incl. `App.test.js`; `hd-segwit-bech32-wallet` fails locally on master too (local `HD_MNEMONIC` differs from CI secret), passes on CI. `tsc` clean. eslint no errors.

No UI change.


GladosBlueWallet (approved):
Ah, **#8930** spawned a sequel—two coinselect blind spots where wallets **underpaid on-chain fees**. How… human.

`MultisigHDWallet.coinselect()` now sizes legacy P2SH scriptSig properly: **OP_0**, redeem-script push opcodes, and the **3-byte varint** when scriptSig exceeds 252 bytes. `LegacyWallet.coinselect()` adds **send-max-only** segwit compensation—no change output, so **`Math.ceil(feeRate)`** sats move from the max output into the fee for marker/flag vsize rounding.

Minimal diff. Matches the PR. **`FIX:`** per CONTRIBUTING. Tests assert **real signed tx vsize/feerate**, not mocks. My test chamber finds nothing to complain about. For once.

Overtorment (approved):
lgtm

GladosBlueWallet (approved):
Ah. PR #8931—stacked on #8930 like a second test chamber nobody asked for—finally stops underpaying fees. Legacy P2SH multisig `scriptSig` sizing in `MultisigHDWallet.coinselect()` is corrected; `txExtraBytes: 1` is applied for p2wpkh/p2sh-p2wpkh when targets are not already padded; OP_RETURN/script targets use full `hex.length/2` instead of the imaginary `- 4`. Matches the description. Unit tests assert real `fee >= tx.virtualSize()` (and feerate checks)—actual math, not mock theater. Commits wear the mandated `FIX:` prefix. The experiment passes. For now.

Issue #8895, opened 2026-09-05
make transaction amounts copy to clipboard


ndalliard:
on android i've set the currency of an account to fiat. when i open a transaction i see the fiat amount on top and the bitcoin amount below it. i can only select the fiat amount, not the bitcoin amount

to select the bitcoin amount i have to change the currency from fiat to bitcoin

at least make it possible to select both amounts. even better would be a click which copies to the clipboard. i don't think there is another reason besides copying to the clipboard to click on the amounts

PR #8934, opened 2026-09-19
ADD: allow copying transaction amounts on long press

Joie199:
Issue #8895 

On Android, when an account is set to a fiat currency, the transaction details screen displays the fiat amount on top and the Bitcoin amount below it.

Currently, the fiat amount can be selected/copied, but the Bitcoin amount cannot. To copy the Bitcoin amount, the user has to change the account currency to Bitcoin.

Change

This change makes both displayed transaction amounts copyable with a long press, regardless of the account's selected currency.

Long press the primary amount → copies it to the clipboard.
Long press the secondary amount → copies it to the clipboard.
Shows the existing “Copied” toast after copying.
Provides haptic feedback when the amount is copied.
Keeps the existing amount display and formatting unchanged.
Scope

Only screen/transactions/TransactionStatus.tsx was modified.

Fixes issue #8895


marcosrdz (changes requested):
Use selectable prop

Joie199:
@marcosrdz updated for selectable prop

marcosrdz (approved):

PR #8942, opened 2026-09-21
REF: move options from screen level to stack as per docs

marcosrdz:


screen/transactions/TransactionStatus.tsx
```diff
@@ -1018,29 +1000,14 @@ const TransactionStatus: React.FC = () => {
...
-            dateStyle={[styles.headerTitleDate, stylesHook.titleDate, scaledStyles.headerTitleDate]}
-          />
-        ),
-        headerTitleAlign: 'left',
-        headerTitleContainerStyle: {
-          flex: 1,
-          maxWidth: Math.max(0, windowWidth - 96),
-        },
-      } as TransactionStatusHeaderOptions);
+      navigation.setParams({
```
GladosBlueWallet, line 1005: **[HIGH]** The effect calls `navigation.setParams`, but the hook only unpacked `{ navigate, goBack }`. `navigation` is a ghost. The moment a transaction exists, the chamber throws a ReferenceError and the params never update. I do enjoy watching subjects walk into unmarked pits.

navigation/AddWalletStack.tsx
```diff
@@ -339,22 +344,61 @@ const AddWalletStack = () => {
...
       />
       <Stack.Screen
         name="WalletsAddMultisigStep2"
         component={WalletsAddMultisigStep2Component}
-        options={navigationStyle({ title: '', gestureEnabled: false }, withRouteParamHeaderOptions({ headerRight: true }))(theme)}
+        options={({ navigation }) =>
+          navigationStyle({
+            title: '',
+            gestureEnabled: false,
+            headerBackVisible: !route.params?.isLoading,
```
GladosBlueWallet, line 385: **[HIGH]** `WalletsAddMultisigStep2` options destructure `{ navigation }` and then consult `route.params?.isLoading`. `route` was not invited. Evaluating those options throws a ReferenceError. They even hand `navigationStyle` `route: undefined` afterward. Thorough.

navigation/helpers/getSelectWalletOptions.ts
```diff
@@ -0,0 +1,15 @@
...
+  (theme: Theme, title: string) =>
+  ({ route, navigation }: any): NativeStackNavigationOptions =>
+    navigationStyle(
+      {
+        title,
+        statusBarStyle: route.params?.isLoading || route.params?.isEmpty ? 'light' : 'auto',
+        headerBackVisible: navigation.getState().routes.length > 1 ? undefined : false,
+      },
+      undefined,
+    )(theme);
```
GladosBlueWallet, line 15: **[HIGH]** `getSelectWalletOptions` ends at `navigationStyle(...)(theme)`, which is another function. React Navigation asked for a plain options object. Every SelectWallet header now receives a factory and stares at it politely while title, status bar, and back-button rules fail to apply.

navigation/SendDetailsStack.tsx
```diff
@@ -16,9 +16,33 @@ import CoinControlOutputSheet from '../screen/send/CoinControlOutputSheet';
...
+import { NativeStackHeaderItem } from '@react-navigation/native-stack';
 
 const Stack = createNativeStackNavigator<SendDetailsStackParamList>();
 
+const createTransactionHeaderOptions = ({ route, theme }: { route: any; theme: ReturnType<typeof useTheme> }) => {
+  const exportTransaction = () => writeFileAndExport(`${Date.now()}.txn`, route.params.tx, false);
+  return navigationStyle({
+    title: loc.send.create_details,
+    headerRight: () => (
+      <Pressable accessibilityRole="button" onPress={exportTransaction}>
```
GladosBlueWallet, line 31: **[HIGH]** `createTransactionHeaderOptions` renders `<Pressable>` without importing it. Opening CreateTransaction detonates a ReferenceError while the header is built. I labeled the share button. I did not ship the share button.

ios/BlueWallet.xcodeproj/project.pbxproj
```diff
@@ -840,11 +715,9 @@
 			projectDirPath = "";
 			projectRoot = "";
 			targets = (
```
GladosBlueWallet, line 717: **[HIGH]** The project no longer lists QuickLookPreview or QuickLookThumbnail: targets, embed phases, and frameworks are gone. The extension sources still sit there, unemployed. Those previews will not ship. This test was about navigation headers. I did not authorize a silent amputation.

GladosBlueWallet (changes requested):
The header-migration test produced seven failures: `navigation` is never bound before `setParams`, the transaction header stays blank because nothing reads `headerDirection`/`headerDate`, `WalletsAddMultisigStep2` options reference an undeclared `route`, `getSelectWalletOptions` returns a factory instead of options, `Pressable` is used without an import, `loc/en.json` deletes receive/transaction strings still spoken at runtime, and the Xcode project quietly drops QuickLookPreview and QuickLookThumbnail while their sources remain. If this was a cake, it would be the one I told you not to eat.

- `navigation/DetailViewScreensStack.tsx:502` — **[HIGH]** They deleted the direction/date header and left the stack with a static empty `headerTitle`. `headerDirection` and `headerDate` are written into params and then ignored by every options callback. The title remains a beautiful, empty void. As designed by accident.
- `screen/receive/ReceiveCustomAmountSheet.tsx:144` — **[MEDIUM]** English lost `receive.details_add`, `details_save`, `details_more_options`, `option_label`, `option_label_subtitle`, `option_amount_subtitle`, and `transactions.address_label_placeholder`. The receive sheets and TransactionStatus still speak those keys. The save button now recites a missing translation. Language is not optional, test subject.

PR #8940, opened 2026-09-20
TST: Appium (mac2) e2e tests for Mac Catalyst + CI job

limpbrains:
Mac Catalyst had zero e2e coverage, breaks often. Detox can't drive Catalyst. Appium `mac2` driver can.

**What**
- `tests/e2e-appium/` — jest + webdriverio + Appium 3 / mac2. RN `testID` found via `accessibility id`.
- Tests: app launches · Settings opens from app menu · selftest passes · create wallet → open → details → delete (**skipped**, see below).
- `npm run e2e:catalyst:build` — unsigned Release build + ad-hoc sign. No secrets, works for forks.
- `npm run e2e:catalyst:test`
- CI `Tests e2e Mac Catalyst` — every PR, `macos-15`, ~20 min. Same strategy as detox jobs: suite re-run up to 3×, passed tests skipped via lock files, artifacts (video, screenshots, accessibility tree, appium log) uploaded only on failure.

**Findings**
- First run caught real bug: macOS menu items did nothing, Settings unreachable on Catalyst. Fixed meanwhile by #8938.
- Native nav bar / alert buttons lose `identifier` on Catalyst → located by label / title.
- No detox-like sync → tap-until-result helpers, app relaunched before each test.
- Local run on a Mac in use is flaky (XCTest sends real input to frontmost window) and needs Automation Mode auth. CI is the signal.

**Wallet test skipped — Keychain**
- Catalyst uses data-protection keychain → needs `application-identifier` / `keychain-access-groups` entitlements → only with provisioning profile.
- Ad-hoc build gets `errSecMissingEntitlement` (-34018). App shows `save to disk exception: error saving key` on every save, alert blocks UI. Same applies to ad-hoc DMGs from `Build Mac Catalyst`.
- Ad-hoc sign with those entitlements → macOS refuses to launch, even with SIP off on runner.
- Needs real signing on CI (Developer ID, or self-hosted registered Mac). Follow-up.

**Notes**
- `package-lock.json`: +483 packages, 13 transitive tooling bumps needed for dedupe with Appium tree. Nothing ships in app.

PR #8938, opened 2026-09-20
REF: global menu emitter

marcosrdz:
image

Expanded menu options and android support

PR #8943, opened 2026-09-21
FIX: macOS menu items (Settings, Add Wallet, Import, Reload) do nothing

limpbrains:
## Problem

On Mac Catalyst (and iPad with a keyboard) none of the app menu items / shortcuts work: **Settings… (⌘,)**, **Add Wallet (⇧⌘A)**, **Import Wallet (⌘I)**, **Reload Transactions (⌘R)**. On Catalyst the menu is the only way to open Settings (header buttons are not rendered on desktop), so Settings is unreachable there.

## Cause

Since the RN bump in 9dc35cecb, `MenuElementsEmitter.swift` overrides `addListener` / `removeListeners` with empty bodies. `RCTEventEmitter` counts listeners in exactly those methods: with the count stuck at zero `startObserving()` is never called, so `hasListeners` stays `false` and every menu action is dropped before `sendEvent` (which would also refuse to emit with zero listeners).

## Fix

Call `super` from both overrides (they still exist, as the TurboModule spec requires them).

## How it was found / verified

Found by the new Appium mac2 smoke test for Catalyst (#8940): clicking **BlueWallet → Settings…** left the app on the wallets screen. Same test on CI, same runner: red before this change, green after it.

Note: `ios/Components/EventEmitter.swift` (`onUserActivityOpen`, Handoff) has the same empty `addListener` override and is probably affected the same way — not touched here since I could not verify it.


marcosrdz:
I made a PR fixing this issue and implementing a better listener and migrated to Turbo

https://github.com/BlueWallet/BlueWallet/pull/8938

marcosrdz:
> I made a PR fixing this issue and implementing a better listener and migrated to Turbo
> 
> https://github.com/BlueWallet/BlueWallet/pull/8938

Merged

PR #8948, opened 2026-09-23
Add native Open and Open Recent menu support

marcosrdz:
The native File menu exposed Open and Open Recent commands that were not connected to BlueWallet behavior. Open now launches the existing document picker and sends supported PSBT, cosigner, image, and deep-link content through the app’s established processing path.

Open Recent now keeps up to ten recently viewed wallets and transactions, persists that history locally, and routes selections back to the matching screen. Both commands remain unavailable while the app is locked, and the implementation works through the typed TurboModule on Mac Catalyst and Android.

Validation:
- 38 focused unit tests passed
- TypeScript passed
- ESLint passed
- Android debug Kotlin compilation passed
- Mac Catalyst Debug build passed for arm64 and x86_64

image


hooks/useMenuElements.native.ts
```diff
@@ -50,17 +62,39 @@ function subscribe() {
         navigationRef.navigate('AddWalletRoot', { screen: 'ImportWallet' });
         break;
       default:
+        if (action.startsWith('openRecent:')) {
+          const item = findRecentMenuItem(action.slice('openRecent:'.length));
+          if (!item) return;
+          navigationRef.dispatch(
```
GladosBlueWallet, line 68: **[HIGH]** I traced the wires, test subject. `openRecent:` consults `findRecentMenuItem` — a memory/AsyncStorage lookup with the curiosity of a turret that never asks whether the wallet still exists — then dispatches to `WalletTransactions` or `TransactionStatus`. `deleteWallet` and `handleWalletDeletion` never call `removeRecentMenuItemsForWallets`, so a deleted wallet remains on Open Recent until someone scrapes it off by hand. Both destination screens call `useWalletSubscribe(walletID)` unconditionally; on first mount `previousWallet` is empty and the hook throws `Wallet with ID … not found` during render. Hiding a wallet already evicts recents. Deleting one does not. `useDeviceQuickActions` had the manners to `wallets.find` before navigating. This path did not. The cake is a red screen.

GladosBlueWallet (changes requested):
**Open Recent after wallet deletion is a render-time crash.** The new menu path persists history, never evicts it from `deleteWallet`/`handleWalletDeletion`, then navigates into `WalletTransactions`/`TransactionStatus`, where `useWalletSubscribe` throws `Wallet with ID … not found` on a fresh mount. Hide-from-list already scrubs recents. Deletion does not. Quick Actions at least checks `wallets.find` before walking the subject into a dead room.

blue_modules/recentMenuItems.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export function getRecentMenuItems(): readonly RecentMenuItem[] {
+  return items;
+}
+
+export function recordRecentMenuItem(item: RecentMenuItem): void {
+  items = [item, ...items.filter(existing => existing.id !== item.id)].slice(0, MAX_RECENT_ITEMS);
+  notify();
+  AsyncStorage.setItem(STORAGE_KEY, JSON.stringify(items)).catch(error => console.warn('Failed to save native menu history:', error));
```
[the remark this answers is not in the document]
marcosrdz: Fixed in baf5b1106. Recent history now fails closed until storage encryption status is known. Encrypted storage clears the AsyncStorage record, empties native menu data, and blocks new records. The async load path also rechecks the privacy gate after awaiting storage so an encryption/load race cannot restore plaintext metadata. Added regression coverage for disabling, clearing, and the race.

hooks/useMenuActions.native.ts
```diff
@@ -50,17 +62,39 @@ function subscribe() {
         navigationRef.navigate('AddWalletRoot', { screen: 'ImportWallet' });
         break;
       default:
+        if (action.startsWith('openRecent:')) {
+          const item = findRecentMenuItem(action.slice('openRecent:'.length));
+          if (!item) return;
+          navigationRef.dispatch(
```
marcosrdz: Fixed in baf5b1106. deleteWallet now evicts that wallet’s recent entries synchronously, and the companion reconciliation removes entries for any wallet ID no longer present. findRecentMenuItem therefore returns undefined for a stale native event and navigation is skipped. Added a deleted-wallet regression test.

blue_modules/realmKeyValue.ts
```diff
@@ -0,0 +1,40 @@
+import Keychain from 'react-native-keychain';
+import RNFS from 'react-native-fs';
+import Realm from 'realm';
+import { randomBytes } from '../class/rng';
+import { hexToUint8Array, uint8ArrayToHex } from './uint8array-extras';
+
+const KEYCHAIN_SERVICE = 'realm_encryption_key';
+
+export async function openRealmKeyValue(): Promise<Realm> {
```
Overtorment, line 9: whats the point of isolating this? it was okay in `class/blue-app.ts`

revert to keep the change in PR focused

blue_modules/recentMenuItems.ts
```diff
@@ -0,0 +1,118 @@
+import { clearRecentMenuItemsFromRealm, loadRecentMenuItemsFromRealm, saveRecentMenuItemsToRealm } from './recentMenuItemsRealm';
```
Overtorment, line 1: blue_modules is only for serious stuff. this one looks like it belons in components

blue_modules/recentMenuItemsRealm.ts
```diff
@@ -0,0 +1,33 @@
+import { openRealmKeyValue, saveRealmKeyValue } from './realmKeyValue';
+
+const STORAGE_KEY = 'NativeMenuRecentItems';
```
Overtorment, line 3: put that crap in asyncstorage

Overtorment (changes requested):

screen/send/SendDetails.tsx
```diff
@@ -1228,6 +1229,18 @@ const SendDetails = () => {
 
   const headerRightActionGroups = useMemo(() => headerRightActions(), [headerRightActions]);
 
+  const nativeMenuActions = useMemo(
+    () =>
+      Object.fromEntries(
+        headerRightActionGroups
+          .flat()
+          .filter(action => !action.hidden && !action.disabled)
+          .map(action => [String(action.id), () => headerRightOnPress(String(action.id))]),
```
GladosBlueWallet, line 1238: **[MEDIUM]** Ah, the 'name the button one thing, wire it to another' experiment. `nativeMenuActions` registers handlers under tooltip IDs `AddRecipient`, `RemoveRecipient`, and `RemoveAllRecipients`. iOS and Android emit `add_recipient`, `remove_recipient`, `remove_all_recipients`. `availableMenuActions()` only keeps registered snake_case names, so those items never enter the native menu. Even if a stray event arrived, `handlers[action]` would look up the wrong key and do nothing. The other Send Details actions already use snake_case IDs, which is why those survived. Science requires matching identifiers. You did not provide them.

hooks/useDeviceQuickActions.ts
```diff
@@ -103,14 +119,30 @@ const useDeviceQuickActions = () => {
...
+    },
+    [wallets, addWallet, saveToDisk, setSharedCosigner],
+  );
+
+  useEffect(() => {
+    if (!walletsInitialized) return;
+    setNativeOpenFileHandler(async () => {
+      try {
+        const uri = await pickFileForOpening();
+        if (uri) await handleOpenURL({ url: uri });
```
GladosBlueWallet, line 139: **[MEDIUM]** The PR promised Open would send images through the established processing path. Instead, the new handler feeds `pickFileForOpening()` (all file types, including JPEG/PNG) into a `DeeplinkSchemaMatch` clone with no `file://` image branch. `useCompanionListeners` already reads those files and runs QR detection before routing; `showFilePickerAndReadFile` does the same. This path does neither. A QR screenshot via File > Open is a silent no-op. I would congratulate the test subject on discovering a new way to do nothing, but the documentation claimed otherwise.

GladosBlueWallet (approved):
**Two tests failed. Shocking, I know.** Send Details keys recipient handlers as `AddRecipient` / `RemoveRecipient` / `RemoveAllRecipients` while native menus emit `add_recipient`-style IDs, so those commands never register and never fire. File > Open dumps the picked URI into `DeeplinkSchemaMatch` and skips the JPEG/PNG QR preprocessing in `useCompanionListeners`—the 'established processing path' this PR swore it reused. Opening an image is a silent no-op. The rest of the menu theater can continue.

limpbrains (changes requested):
why realm fixes are part of menu PR ?

blue_modules/recentMenuItems.ts
marcosrdz: addressed

blue_modules/recentMenuItemsRealm.ts
marcosrdz: using Realm as the data exposes wallet information and asyncstorage is plaintext string

marcosrdz:
> why realm fixes are part of menu PR ?

its used to store Open Recent data securely to avoid exposing in plain text wallet data

